PR #224, opened 2022-10-14
test: adding unit tests

karenc-bq:
### Summary

Adding unit tests and configuring Jacoco

### DIFF
--- a/wrapper/build.gradle.kts
+++ b/wrapper/build.gradle.kts
@@ -124,14 +124,41 @@ tasks.spotbugsTest {
     }
 }
 
-tasks.jacocoTestCoverageVerification {
+tasks.withType<JacocoCoverageVerification> {
     violationRules {
         rule {
             limit {
-                // Coverage verification will pass if it is greater than or equal to 1%.
-                minimum = "0.01".toBigDecimal()
+                minimum = BigDecimal(0.57)
             }
         }
+    }
+
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.map {
+            fileTree(it).apply {
+                exclude(
+                        "software/amazon/jdbc/wrapper/*",
+                        "software/amazon/jdbc/util/*",
+                        "software/amazon/jdbc/profile/*",
+                        "software/amazon/jdbc/plugin/DataCacheConnectionPlugin*"
+                )
+            }
+        }))
+    }
+}
+
+tasks.withType<JacocoReport> {
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.map {
+            fileTree(it).apply {
+                exclude(
+                        "software/amazon/jdbc/wrapper/*",
+                        "software/amazon/jdbc/util/*",
+                        "software/amazon/jdbc/profile/*",
+                        "software/amazon/jdbc/plugin/DataCacheConnectionPlugin*"
+                )
+            }
+        }))
     }
 }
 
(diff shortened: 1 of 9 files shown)


wrapper/src/test/java/software/amazon/jdbc/plugin/DataCacheConnectionPluginTest.java
```diff
@@ -0,0 +1,123 @@
...
+    // Execute the query again with a different result set.
+    rs = plugin.execute(
+        ResultSet.class,
+        SQLException.class,
+        mockStatement, methodName,
+        () -> mockResult2,
+        new String[]{"foo"}
+    );
+
+    compareResults(mockResult1, rs);
```
aaron-congo, line 113: should we be verifying somehow that jdbcMethodFunc was only actually called once because the result set was returned from the cache the second time?

wrapper/src/test/java/software/amazon/jdbc/util/ConnectionUrlBuilderTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.MockitoAnnotations;
+import software.amazon.jdbc.HostSpec;
+
+class ConnectionUrlBuilderTest {
+
+  private AutoCloseable closeable;
```
aaron-congo, line 35: looks like there are no mocks so this can be removed

wrapper/src/test/java/software/amazon/jdbc/plugin/DataCacheConnectionPluginTest.java
```diff
@@ -0,0 +1,131 @@
...
+    rs = plugin.execute(
+        ResultSet.class,
+        SQLException.class,
+        mockStatement, methodName,
+        mockCallable,
+        new String[]{"foo"}
+    );
+
+    compareResults(mockResult1, rs);
+    verify(mockCallable).call();
```
aaron-congo, line 121: Should we be verifying this was only called once instead of potentially multiple times (eg `verify(mockCallable, time(1)).call()`)? As is I think the test would still pass if the results were not fetched from the cache the second time, which shouldn't be the case (unless I'm misunderstanding)

wrapper/build.gradle.kts
```diff
@@ -124,15 +124,42 @@ tasks.spotbugsTest {
...
     }
+
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.map {
+            fileTree(it).apply {
+                exclude(
+                        "software/amazon/jdbc/wrapper/*",
+                        "software/amazon/jdbc/util/*",
+                        "software/amazon/jdbc/profile/*",
+                        "software/amazon/jdbc/plugin/DataCacheConnectionPlugin*"
```
sergiyvamz, line 143: Why not to include all plugin package?

karenc-bq: This code excludes certain classes from the test coverage. We have tests for some of the more important plugins, like EFM and Failover.

wrapper/src/test/java/software/amazon/jdbc/plugin/DataCacheConnectionPluginTest.java
karenc-bq: When you don't specify the number of invocations in `verify` it is default to 1, i.e. `verify(mockCallable).call();` and `verify(mockCallable, time(1)).call()` are exactly the same.
 
https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#exact_verification

PR #230, opened 2022-10-18
feat: track the current connection's prepared transaction state

karenc-bq:
### Summary

Keep track of the connection's transaction state and skip querying topology if it is in the `PREPARED` state.

### Additional Reviewers

@sergiyvamz @congoamz @joyc-bq @alecc-bq

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
@@ -48,4 +48,9 @@ public class SubscribedMethodHelper {
       "CallableStatement.executeUpdate",
       "CallableStatement.executeLargeUpdate"
   );
+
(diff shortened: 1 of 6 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
```diff
@@ -48,4 +48,9 @@ public class SubscribedMethodHelper {
       "CallableStatement.executeUpdate",
       "CallableStatement.executeLargeUpdate"
   );
+
+  public static final List<String> XA_PREPARE_TRANSACTION = Arrays.asList(
+      "MysqlXAConnection.prepare",
```
sergiyvamz, line 53: Where are we using this method names? In what wrapper class?

sergiyvamz: I'm afraid a proper implementation of such feature should include a complete support of XAResource/XAConnection interfaces

PR #233, opened 2022-10-19
refactor: make topology queries more consistent between different databases

karenc-bq:
### Summary

Modify the queries used by MySQL and PostgreSQL to query topologies so they are more consistent

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
@@ -82,7 +82,8 @@ public class AuroraHostListProvider implements DynamicHostListProvider {
   static final String MYSQL_RETRIEVE_TOPOLOGY_SQL =
       "SELECT SERVER_ID, SESSION_ID"
           + "FROM information_schema.replica_host_status "
-          + "WHERE time_to_sec(timediff(now(), LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID'" // 5 min
+          // filter out nodes that haven't been updated in the last 5 minutes
+          + "WHERE time_to_sec(timediff(now(), LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID'"
           + "ORDER BY LAST_UPDATE_TIMESTAMP";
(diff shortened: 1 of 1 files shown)


sergiyvamz (approved):

davecramer:
I don't think removing information is the way to go here. 

I'd prefer to see us adding `replica_lag_in_ms` to the postgres query. 

Additionally we should be prioritizing the most recent replica by using the following
`order by highest_lsn_rcvd desc NULLS FIRST` in PostgreSQL
I think there is an equivalent predicate in MySql

This will be especially important in the Read/Write Splitting plugin where we would want to make sure we are providing the client with the most recent data.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -80,9 +80,10 @@ public class AuroraHostListProvider implements DynamicHostListProvider {
           + "ORDER BY LAST_UPDATE_TIMESTAMP";
 
   static final String MYSQL_RETRIEVE_TOPOLOGY_SQL =
-      "SELECT SERVER_ID, SESSION_ID, LAST_UPDATE_TIMESTAMP, REPLICA_LAG_IN_MILLISECONDS "
+      "SELECT SERVER_ID, SESSION_ID"
           + "FROM information_schema.replica_host_status "
-          + "WHERE time_to_sec(timediff(now(), LAST_UPDATE_TIMESTAMP)) <= 300 " // 5 min
+          // filter out nodes that haven't been updated in the last 5 minutes
+          + "WHERE time_to_sec(timediff(now(), LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID'"
```
davecramer, line 86: We should leave REPLICA_LAG_IN_MILLISECONDS in and add it to the postgres version. We should also be ordering by replica_lag_in_milliseconds so that the most recent replica is first

karenc-bq: We are not using `REPLICA_LAG_IN_MILLISECONDS` in the code right now, so it should be safe to keep it out until we need it. During failover we randomize the readers so this ordering isn't really used.

davecramer: we should not be randomizing readers. We should be using the reader that is the most consistent with the writer.

Issue #236, opened 2022-10-21
Ability to use IAM over Jump Boxes


frugecn:
### Describe the feature

Be able to use the iam plugin when going through a jump box.  Currently, it errors saying that localhost does not match the rds endpoint structure.

### Use Case

I'm trying to use JDBC to test applications locally connecting to Aurora Database through JDBC where it is required to use a jump box.  With the jump box configuration, you have to use the localhost address when connecting to the database.

### Proposed Solution

Add a field to the wrapper that would be the endpoint of the database instance (failover??) that when localhost is used the endpoint field could be used in its place.

### Other Information

Unsupported AWS hostname localhost. Amazon domain name in format *.AWS-Region.rds.amazonaws.com or *.rds.AWS-Region.amazonaws.com.cn is expected.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

1.0.0

### JDK version used

OpenJDK 18.0.02

### Operating System and version

Windows 11

karenc-bq:
Hi @frugecn, thank you for reaching out.

When you are using a jump box, you can specify the AWS domain name through the connection parameter provided by the Failover Connection Plugin: `clusterInstanceHostPattern`.

Here is a sample connection string: `jdbc:aws-wrapper:postgresql://localhost:5432/db?wrapperPlugins=iam,failover&clusterInstanceHostPattern=?.hash-number.us-east-2.rds.amazonaws.com`

You can learn more about `clusterInstanceHostPattern` [here](https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/using-plugins/UsingTheFailoverPlugin.md#host-pattern).

frugecn:
Tried this connection string:

`jdbc:aws-wrapper:postgresql://localhost:5432/mypgdb?wrapperPlugins=iam,failover&clusterInstanceHostPattern=?.cluster-cpjzfiwcidkb.us-east-1.rds.amazonaws.com`

but get this result:

Unsupported AWS hostname localhost. Amazon domain name in format *.AWS-Region.rds.amazonaws.com or *.rds.AWS-Region.amazonaws.com.cn is expected.

I'm not quite understanding the documentation.  It refers to customHost in two spots, but in the case of localhost, how would it know to use the correct endpoint during the IAM authentication?  

Based off of this part of the article:

[...]

PR #240, opened 2022-10-25
feat: add IAM configuration parameter iamHost

karenc-bq:
### Summary

feat: add IAM configuration parameter to use a different hostname to generate IAM token.

### Description

This change allows users connecting via SSH Tunneling to override the hostname used to generate IAM tokens.
Without this change, if users are connecting with `localhost:5432` as the hostname, the IAM authentication plugin will attempt to generate a token with `localhost`, which generates an incorrect token.

#236 

### Additional Reviewers

@sergiyvamz


docs/using-the-jdbc-driver/using-plugins/UsingTheIamAuthenticationPlugin.md
```diff
@@ -24,14 +24,13 @@ IAM database authentication use is limited to certain database engines. For more
...
+| `iamDefaultPort` | String  |    No    | This property will override the default port that is used to generate the IAM token. The default port is determined based on the underlying driver protocol. For now, there is support for `jdbc:postgresql:` and `jdbc:mysql:`. Target drivers with different protocols will require users to provide a default port. | `1234`                                              |
+| `iamHost`        | String  |    No    | This property will override the default hostname that is used to generate the IAM token. The default hostname is derived from the connection string. This parameter is useful when using SSH tunneling is in use and the connection string uses `localhost` instead of an RDS endpoint to establish connections.       | `database.cluster-hash.us-east-1.rds.amazonaws.com` |
```
sergiyvamz, line 31: I'm not sure I understand this sentence. Can we re-word it?

> This parameter is useful when using SSH tunneling is in use and the connection string uses localhost instead of an RDS endpoint to establish connections.

sergiyvamz (approved):

sergiyvamz (approved):

PR #241, opened 2022-10-25
add new parameter failoverStrictReader

sergiyvamz:
### Summary

Implement a new parameter to control failover logic. https://github.com/awslabs/aws-advanced-jdbc-wrapper/issues/223

### Additional Reviewers

@karenc-bq 
@congoamz 
@hsuamz


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -137,15 +138,19 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
           "enableClusterAwareFailover", "true",
           "Enable/disable cluster-aware failover logic");
 
+  public static final AwsWrapperProperty FAILOVER_STRICT_READER =
+      new AwsWrapperProperty(
+          "failoverStrictReader", "false",
+          "Enable/disable failover to a writer node for read-only connection.");
```
aaron-congo, line 144: I think your other description above is a bit clearer:
"Set to true to only allow failover to reader nodes during the reader failover process. If enabled, reader failover to a writer node will only be allowed for single-node clusters. This logic mimics the logic of the Aurora read-only cluster endpoint."

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -212,7 +239,8 @@ public List<HostSpec> getHostsByPriority(List<HostSpec> hosts) {
     Collections.shuffle(downHostList);
 
     List<HostSpec> hostsByPriority = new ArrayList<>(activeReaders);
-    if (writerHost != null) {
+    int numOfReaders = activeReaders.size() + downHostList.size();
```
karenc-bq, line 242:
```suggestion
    final int numOfReaders = activeReaders.size() + downHostList.size();
```

aaron-congo:
Should we add a test for this scenario or do you think it isn't needed?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -137,15 +138,19 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
           "enableClusterAwareFailover", "true",
           "Enable/disable cluster-aware failover logic");
 
+  public static final AwsWrapperProperty FAILOVER_STRICT_READER =
+      new AwsWrapperProperty(
+          "failoverStrictReader", "false",
```
karenc-bq, line 143: Since the default value is false, can we rename this variable to `enableFailoverStrictReader`?

karenc-bq: And the description will be `Enable reader-only connection during failover`.

aaron-congo: nit: "Enable to enforce reader-only connection during reader failover" might be a bit clearer

karenc-bq (approved):

docs/using-the-jdbc-driver/using-plugins/UsingTheFailoverPlugin.md
```diff
@@ -21,14 +21,15 @@ The failover plugin will NOT be loaded unless you explicitly include it by addin
...
+| `failoverStrictReader`                 | Boolean |                                        No                                        | Set to `true` to disable connecting to a primary node during reader failover process. A connection to a primary node is allowed only for single-node clusters. Such logic mimics the logic of Aurora read-only cluster endpoint.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | `false`                                                                                                                                                                                              |
```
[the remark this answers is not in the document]
sergiyvamz: Great! Thank you for suggestion!

PR #250, opened 2022-10-31
doc: update Hikari failover examples

karenc-bq:
### Summary

Add a failover-specific example.

### DIFF
--- a/examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
+++ b/examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
@@ -17,14 +17,17 @@
 package software.amazon;
 
 import com.zaxxer.hikari.HikariDataSource;
-import software.amazon.jdbc.ds.AwsWrapperDataSource;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Properties;
+import software.amazon.jdbc.ds.AwsWrapperDataSource;
 
-public class HikariExample {
+/**
+ * An example to use the AWS Wrapper with Hikari with failover enabled.
+ */
+public class HikariFailoverExample {
 
   private static final String USER = "username";
   private static final String PASSWORD = "password";
@@ -47,6 +50,11 @@ public class HikariExample {
       ds.addDataSourceProperty("portPropertyName", "portNumber");
       ds.addDataSourceProperty("serverPropertyName", "serverName");
 
+      // The failover plugin throws failover-related exceptions that need to be handled explicitly by HikariCP,
+      // otherwise connections will be closed immediately after failover. Set `ExceptionOverrideClassName` to provide
+      // a custom exception class.
+      ds.setExceptionOverrideClassName("software.amazon.jdbc.util.HikariCPSQLException");
+
       // Specify the driver-specific data source for AwsWrapperDataSource:
       ds.addDataSourceProperty("targetDataSourceClassName", "org.postgresql.ds.PGSimpleDataSource");
 
@@ -55,6 +63,9 @@ public class HikariExample {
       targetDataSourceProps.setProperty("serverName", ENDPOINT);
       targetDataSourceProps.setProperty("databaseName", DATABASE_NAME);
       targetDataSourceProps.setProperty("portNumber", "5432");
+
+      // Enable the failover and host monitoring connection plugins.
+      targetDataSourceProps.setProperty("wrapperPlugins", "failover,efm");
 
       ds.addDataSourceProperty("targetDataSourceProperties", targetDataSourceProps);
 


examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+    targetDataSourceProps.setProperty("serverName", "db-identifier.cluster-XYZ.us-east-2.rds.amazonaws.com");
+    targetDataSourceProps.setProperty("databaseName", "postgres");
+    targetDataSourceProps.setProperty("portNumber", "5432");
+
+    // Enable the Failover and Host Monitoring plugins.
+    targetDataSourceProps.setProperty("wrapperPlugins", "failover,efm");
+
+    ds.addDataSourceProperty("targetDataSourceProperties", targetDataSourceProps);
+
+    // Try and make a connection:
```
joyc-bq, line 70: maybe "Attempt a connection"

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+    // Specify the driver-specific data source for AwsWrapperDataSource:
+    ds.addDataSourceProperty("targetDataSourceClassName", "org.postgresql.ds.PGSimpleDataSource");
+
+    // Configuring PGSimpleDataSource:
+    Properties targetDataSourceProps = new Properties();
+    targetDataSourceProps.setProperty("serverName", "db-identifier.cluster-XYZ.us-east-2.rds.amazonaws.com");
+    targetDataSourceProps.setProperty("databaseName", "postgres");
+    targetDataSourceProps.setProperty("portNumber", "5432");
+
+    // Enable the Failover and Host Monitoring plugins.
```
joyc-bq, line 65: should failover and host monitoring be lowercase?

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+import com.zaxxer.hikari.HikariDataSource;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+import software.amazon.jdbc.ds.AwsWrapperDataSource;
+
+/*
+An example to use the AWS Wrapper with Hikari with failover enabled.
```
joyc-bq, line 28: should line 28 also have a * before the comment

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+    // a custom exception class.
+    ds.setExceptionOverrideClassName("software.amazon.jdbc.util.HikariCPSQLException");
+
+    // Specify the driver-specific data source for AwsWrapperDataSource:
+    ds.addDataSourceProperty("targetDataSourceClassName", "org.postgresql.ds.PGSimpleDataSource");
+
+    // Configuring PGSimpleDataSource:
+    Properties targetDataSourceProps = new Properties();
+    targetDataSourceProps.setProperty("serverName", "db-identifier.cluster-XYZ.us-east-2.rds.amazonaws.com");
+    targetDataSourceProps.setProperty("databaseName", "postgres");
```
joyc-bq, line 62: should the database name be something other than postgres? I believe in other examples I was advised to change it to something else

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+    // otherwise connections will be closed immediately after failover. Set `ExceptionOverrideClassName` to provide
+    // a custom exception class.
+    ds.setExceptionOverrideClassName("software.amazon.jdbc.util.HikariCPSQLException");
+
+    // Specify the driver-specific data source for AwsWrapperDataSource:
+    ds.addDataSourceProperty("targetDataSourceClassName", "org.postgresql.ds.PGSimpleDataSource");
+
+    // Configuring PGSimpleDataSource:
+    Properties targetDataSourceProps = new Properties();
+    targetDataSourceProps.setProperty("serverName", "db-identifier.cluster-XYZ.us-east-2.rds.amazonaws.com");
```
joyc-bq, line 61: In other examples these are grouped above where your USER and PASSWORD env vars are so that the user can see what they need to configure   
```
  // User configures connection properties here
public static final String POSTGRESQL_CONNECTION_STRING =
      "jdbc:aws-wrapper:postgresql://test-db.cluster-XYZ.us-east-2.rds.amazonaws.com:5432";
```
could also set property like: getDataSourceProps.setProperty("databaseName", DATABASE_NAME")  with DATABASE_NAME set above on line 33 as well

examples/HikariExample/src/main/java/software/amazon/HikariExample.java
```diff
@@ -59,9 +59,9 @@ public static void main(String[] args) throws SQLException {
     // Try and make a connection:
```
joyc-bq, line 59: Attempt a connection

examples/HikariExample/src/main/java/software/amazon/HikariExample.java
```diff
@@ -59,9 +59,9 @@ public static void main(String[] args) throws SQLException {
...
     try (final Connection conn = ds.getConnection();
         final Statement statement = conn.createStatement();
-        final ResultSet rs = statement.executeQuery("SELECT * FROM employees")) {
+        final ResultSet rs = statement.executeQuery("SELECT * from aurora_db_instance_identifier()")) {
```
sergiyvamz, line 62: nit: Can we replace * with some field name?

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
```diff
@@ -0,0 +1,79 @@
...
+
+    // Enable the Failover and Host Monitoring plugins.
+    targetDataSourceProps.setProperty("wrapperPlugins", "failover,efm");
+
+    ds.addDataSourceProperty("targetDataSourceProperties", targetDataSourceProps);
+
+    // Try and make a connection:
+    try (final Connection conn = ds.getConnection();
+        final Statement statement = conn.createStatement();
+        final ResultSet rs = statement.executeQuery("SELECT * from aurora_db_instance_identifier()")) {
```
sergiyvamz, line 73: Similar note.

sergiyvamz (approved):

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
karenc-bq: we could, postgres is a default database that gets created when users create their server so the code would still run if they don't customize it.

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
karenc-bq: I was using them as proper nouns here, I suppose either way is fine.

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
karenc-bq: postgres is generally the default database, what was the reasoning to change it?

examples/HikariExample/src/main/java/software/amazon/HikariExample.java
karenc-bq: This function only has 1 field, I am okay with *

examples/HikariExample/src/main/java/software/amazon/HikariFailoverExample.java
joyc-bq: It's probably fine then

joyc-bq (approved):

PR #257, opened 2022-11-02
Implement the read-write splitting plugin

aaron-congo:


sergiyvamz (approved):

davecramer:
Since this is splitting on transaction boundaries, how are we handling read after writes ?

examples/AWSDriverExample/src/main/java/software/amazon/ReadWriteSplittingPostgresExample.java
```diff
@@ -0,0 +1,137 @@
...
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+import software.amazon.jdbc.plugin.failover.FailoverFailedSQLException;
+import software.amazon.jdbc.plugin.failover.FailoverSuccessSQLException;
+import software.amazon.jdbc.plugin.failover.TransactionStateUnknownSQLException;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
+
+
```
joyc-bq, line 33: nit: extra space here

congoamz:
@davecramer can you clarify your question? The plugin only switches at transaction boundaries if setReadOnly(true) has been called and reader load-balancing is enabled. Because of this, only reads can be performed in this state. The plugin does not switch between a writer/reader on transaction boundaries; instead, setReadOnly is used to switch

joyc-bq (approved):

davecramer:
@congoamz thanks for the clarification. That makes sense. 
One thing to keep in mind is that opening connections on PostgreSQL is not free. I would expect that most apps will have a pool which minimizes the overhead, but if they don't then splitting on transaction boundaries might be an anti-pattern.

congoamz:
@davecramer good to know, thanks. Switching to a new reader on transaction boundaries is disabled by default. We are working on creating some performance tests which will give us some insight into how this feature effects performance

PR #259, opened 2022-11-08
Simplify code using try with resources for the Reentrant lock

davecramer:
Small optimization to make the code simpler

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
@@ -41,10 +41,7 @@ import software.amazon.jdbc.plugin.efm.HostMonitoringConnectionPluginFactory;
 import software.amazon.jdbc.plugin.failover.FailoverConnectionPluginFactory;
 import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsPluginFactory;
 import software.amazon.jdbc.profile.DriverConfigurationProfiles;
-import software.amazon.jdbc.util.Messages;
-import software.amazon.jdbc.util.SqlState;
-import software.amazon.jdbc.util.StringUtils;
-import software.amazon.jdbc.util.WrapperUtils;
+import software.amazon.jdbc.util.*;
 import software.amazon.jdbc.wrapper.ConnectionWrapper;
 
 /**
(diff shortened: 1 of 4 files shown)


wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -105,12 +102,15 @@ public ConnectionPluginManager(ConnectionProvider connectionProvider, Connection
     this.connectionWrapper = connectionWrapper;
   }
 
-  public void lock() {
-    lock.lock();
+  public ResourceLock obtain() {
```
sergiyvamz, line 105: I'd suggest to reconsider a method name. It's not clear what "obtain" on plugin manager object does. Does it obtain a plugin manager?

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -105,12 +102,15 @@ public ConnectionPluginManager(ConnectionProvider connectionProvider, Connection
...
+    return lock.obtain();
   }
 
-  public void unlock() {
-    lock.unlock();
+  /*
+  For testing only
+   */
+  public void closeResourceLock() {
```
sergiyvamz, line 112: "releaseLock"?

davecramer: I'm not married to any name. releaseLock is fine, and then we can use acquireLock instead of obtain.

wrapper/src/main/java/software/amazon/jdbc/util/ResourceLock.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Unlock on exit of try-with-resources block.
+   */
+  @Override
+  public void close() {
+    this.unlock();
+  }
+}
```
karenc-bq, line 54:
```suggestion
}

```

wrapper/src/test/java/software/amazon/jdbc/util/ResourceLockTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final int preLock;
+        final int postLock;
+
+        CounterPair(int preLock, int postLock) {
+            this.preLock = preLock;
+            this.postLock = postLock;
+        }
+    }
+
+}
```
karenc-bq, line 101:
```suggestion
}

```

davecramer:
close in favour of #265

PR #274, opened 2022-11-22
Use default connection check intervals/timeouts in the EFM plugin

aaron-congo:
- previously, the interval/timeout values could be set to 0 or Long.MAX_VALUE. This PR changes these values to specified defaults when a user-supplied setting is not available
- also changed various EFM variables to thread-safe variants. These variables are shared between the main thread and the monitor thread


sergiyvamz (approved):

davecramer:
So I'm looking at the title of this and trying to determine how it could be set to MAX_VALUE ?

congoamz:
> So I'm looking at the title of this and trying to determine how it could be set to MAX_VALUE ?

There were a couple spots where this.connectionCheckInvervalMillis was being set to the return value of findShortestIntervalMillis. The original implementation of this function returned MAX_VALUE if there were no contexts in the MonitorImpl class. The last change in the MonitorImpl file in this PR fixes this

sergiyvamz (approved):

karenc-bq (approved):

PR #283, opened 2022-12-09
fix: check whether the stale DNS plugin needs to do a force topology refresh

karenc-bq:
### Summary

fix: check whether the stale DNS plugin needs to do a force topology refresh

### Additional Reviewers

@sergiyvamz

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
@@ -46,18 +46,25 @@ public class AuroraStaleDnsHelper {
 
   private HostSpec writerHostSpec = null;
   private InetAddress writerHostAddress = null;
+
   private static final int RETRIES = 3;
+  private static final String POSTGRESQL_READONLY_QUERY = "SELECT pg_is_in_recovery() AS is_reader";
+  private static final String MYSQL_READONLY_QUERY = "SELECT @@innodb_read_only AS is_reader";
+  private static final String IS_READER_COLUMN = "is_reader";
 
   public AuroraStaleDnsHelper(PluginService pluginService) {
     this.pluginService = pluginService;
   }
 
   public Connection getVerifiedConnection(
+      final String driverProtocol,
       HostSpec hostSpec,
       Properties props,
-      JdbcCallable<Connection, SQLException> connectFunc,
-      final String query)
-      throws SQLException {
+      JdbcCallable<Connection, SQLException> connectFunc) throws SQLException {
+
+    final String query = driverProtocol.contains("postgresql")
+        ? POSTGRESQL_READONLY_QUERY
+        : MYSQL_READONLY_QUERY;
 
     if (!this.rdsUtils.isWriterClusterDns(hostSpec.getHost())) {
       return connectFunc.call();
@@ -174,7 +181,7 @@ public class AuroraStaleDnsHelper {
       try (final Statement statement = connection.createStatement();
           ResultSet rs = statement.executeQuery(query)) {
         if (rs.next()) {
-          return rs.getBoolean(AuroraStaleDnsPlugin.IS_READER_COLUMN);
+          return rs.getBoolean(IS_READER_COLUMN);
         }
       } catch (SQLSyntaxErrorException e) {
         throw e;
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
@@ -181,7 +181,7 @@ public class AuroraStaleDnsHelper {
       } catch (SQLException e) {
 
         // Return false if the SQLException is not a network error. Otherwise, retry.
-        if (!SqlState.isConnectionError(e)) {
+        if (!pluginService.isNetworkException(e)) {
           return false;
         }
       }
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
@@ -21,6 +21,7 @@ import java.net.UnknownHostException;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLSyntaxErrorException;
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -74,7 +77,12 @@ public Connection getVerifiedConnection(
       return conn;
     }
 
-    this.pluginService.refreshHostList(conn);
+    if (isReadOnly(conn, query)) {
+      this.pluginService.forceRefreshHostList(conn);
```
sergiyvamz, line 81: Let's add descriptive comment why force topology refresh is needed here.

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -153,4 +161,16 @@ private HostSpec getWriter() {
...
   }
+
+  private boolean isReadOnly(Connection connection, final String query) {
+    try (final Statement statement = connection.createStatement();
+        ResultSet rs = statement.executeQuery(query)) {
+      while (rs.next()) {
+        return rs.getBoolean(AuroraStaleDnsPlugin.IS_READER_COLUMN);
+      }
+    } catch (SQLException e) {
+      return false;
```
sergiyvamz, line 172: I'd separate sql syntax exception from others. Syntax exception means that the connection is not to an Aurora cluster and it should re-thrown. Communication-related exceptions probably require a retry. I wouldn't simply return false in such cases.

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsPlugin.java
```diff
@@ -55,6 +55,16 @@ public class AuroraStaleDnsPlugin extends AbstractConnectionPlugin {
       Arrays.asList("connect", "initHostProvider", "notifyNodeListChanged", "*")));
   private final PluginService pluginService;
   private final AuroraStaleDnsHelper helper;
+  private static final String POSTGRESQL_READONLY_QUERY = "SELECT rt.server_id, CASE\n"
```
sergiyvamz, line 58: Will it work for Aurora Mysql?

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
davecramer: This should probably be an if statement not a while statement

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -153,4 +167,25 @@ private HostSpec getWriter() {
...
+          ResultSet rs = statement.executeQuery(query)) {
+        if (rs.next()) {
+          return rs.getBoolean(AuroraStaleDnsPlugin.IS_READER_COLUMN);
+        }
+      } catch (SQLSyntaxErrorException e) {
+        throw e;
+      } catch (SQLException e) {
+
+        // Return false if the SQLException is not a network error. Otherwise, retry.
+        if (!SqlState.isConnectionError(e)) {
```
sergiyvamz, line 184: Should we use our custom exception handler?

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsPlugin.java
sergiyvamz, line 58: Would you consider to move these queries to AuroraStaleDnsHelper? Then getVerifiedConnection() can accept driver protocol and decide what query to run.

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -48,16 +48,29 @@ public class AuroraStaleDnsHelper {
...
   private static final int RETRIES = 3;
 
+  private static final String POSTGRESQL_READONLY_QUERY = "SELECT rt.server_id, CASE\n"
+      + "           WHEN 'MASTER_SESSION_ID' = rt.session_id THEN 'false'\n"
+      + "           ELSE 'true'\n"
+      + "           END AS is_reader\n"
+      + "FROM aurora_replica_status() rt,\n"
+      + "     aurora_db_instance_identifier() di\n"
+      + "WHERE rt.server_id = di;";
+
```
davecramer, line 58: `select pg_is_in_recovery () as is_reader;` is much simpler

karenc-bq: Thank you

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsPlugin.java
```diff
@@ -56,6 +56,8 @@ public class AuroraStaleDnsPlugin extends AbstractConnectionPlugin {
   private final PluginService pluginService;
   private final AuroraStaleDnsHelper helper;
 
+  static final String IS_READER_COLUMN = "is_reader";
```
sergiyvamz, line 59: Do we still need it here in the  plugin? Should we relocate it to the helper class?

sergiyvamz (approved):

shunyy:
@karenc-bq Hi!

I have run some tests on my end with snapshots and it seems to be working great!
Do you have any plans to release these changes?

karenc-bq:
Hi @shunyy, thank you for testing out the changes.

Are you currently evaluating the driver for production use or do you have a timeline where you'd want an official release? We are aiming to do a aiming to do a release by end of January as there are other fixes that we want to include. https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/298. 

For convenience, we do offer snapshot builds here that would contains the fix from this PR: https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#using-a-snapshot-of-the-driver

shunyy:
@karenc-bq 

Thanks for the reply!

I've been testing the snapshot version and wanted to use it in a production environment as soon as possible , but with the schedule you gave me, it's perfectly fine!

Looking forward to it!

PR #293, opened 2022-12-31
Readwrite splitting load performance tests

joyc-bq:
This set of tests measures the CPU performance of the read-write splitting plugin and reader load balancing on heavy tests. 10 connections which are all initially connected to the writer are set up in parallel and execute a set of heavy queries with the prepared test data. This measures the performance of this activity for 3 cases: a) the read-write splitting plugin is not enabled, b) the read-write splitting plugin is enabled but reader-load balancing is not enabled and c) the read-write splitting plugin is enabled and reader-load balancing is enabled.

 

If the read-write splitting plugin is not enabled (a), we can expect to see most of the activity on the writer cluster and very little on the others. If the read-write splitting plugin is enabled without reader load-balancing (b), we can expect to see the most activity on the writer and a single reader, and very little activity on the other readers. If the read-write splitting plugin is enabled with reader load-balancing (c), we can expect to see the activity spread across the instances. This will show users that read-write splitting and reader load balancing spreads activity across the instances rather than heavily active on one node. 

The benchmark results will be added to the README in a later ticket


benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,184 @@
...
+      "order by " +
+      "l_returnflag, " +
+      "l_linestatus;";
+
+  public static void setUp() throws SQLException {
+    if (!org.postgresql.Driver.isRegistered()) {
+      org.postgresql.Driver.register();
+    }
+
+    if (!Driver.isRegistered()) {
```
davecramer, line 71: why is this necessary ?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,184 @@
...
+
+        Thread.sleep(5000);
+        startLatch.countDown(); // notify that this thread is ready for work
+        startLatch.await(5,
+            TimeUnit.MINUTES); // wait for another threads to be ready to start the test
+
+        // Execute long query
+        final Statement statement = conn.createStatement();
+        try (final ResultSet result = statement.executeQuery(QUERY_1)) {
+          fail("Sleep query finished, should not be possible with network downed.");
```
davecramer, line 166: network down ?

davecramer: it should be "with the network down"

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,138 @@
+package software.amazon.jdbc.benchmarks;
```
karenc-bq, line 1: Missing license

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
karenc-bq: We generate the META-INF file during compilation, so we use this to register the driver in the classpath in tests.

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,197 @@
...
+    for (int i = 0; i < NUM_THREADS; i++) {
+      connectionThreadsList.add(getThread_PGReadWriteSplitting(initNoPluginPropsWithTimeouts()));
+    }
+
+    for (Thread connectionThread : connectionThreadsList) {
+      connectionThread.start();
+    }
+
+    for (Thread connectionThread : connectionThreadsList){
+      connectionThread.interrupt();
```
congoamz, line 125: I'm not super familiar with jmh, but why ware we interrupting the threads right after starting them? Won't this stop the thread from completing its task?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,197 @@
...
+
+        Connection conn = connectToInstance(POSTGRESQL_CONNECTION_STRING, props);
+
+        Thread.sleep(5000);
+
+        // execute long query
+        final Statement statement = conn.createStatement();
+        statement.executeQuery(QUERY_1);
+
+        try (final ResultSet result = statement.executeQuery(QUERY_1)) {
```
congoamz, line 175: Is there a reason we are executing the query on both line 173 and line 175? Should we get rid of line 173?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,197 @@
...
+          fail("Sleep query finished, should not be possible with the network down.");
```
congoamz, line 176: I'm a bit confused about this line. Where are we bringing the network down, and why? Why do we not expect the query to complete? Why does it say the "sleep query" finished - it seems like we are not executing a sleep statement but instead the query defined by QUERY_1?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
joyc-bq: it stops the thread after it's task is complete

congoamz: I thought that calling interrupt stops the thread wherever its at in its task (complete or not). How do we ensure the task is complete before interrupt is called?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
davecramer: So you are saying that the META-INF service file isn't available at this point ?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,197 @@
...
+  // User configures connection properties here
+  public static final String POSTGRESQL_CONNECTION_STRING =
+      "jdbc:aws-wrapper:postgresql://test-db.cluster-XYZ.us-east-2.rds.amazonaws.com:5432/readWriteSplittingExample";
+  private static final String USERNAME = "username";
+  private static final String PASSWORD = "password";
+
+  private static final int NUM_THREADS = 10;
+  private static final int EXECUTE_QUERY_TIMES = 5;
+  protected static final String QUERY_1 = "select " +
+      "l_returnflag, " +
```
davecramer, line 56: It would probably be better to query pg_sleep() as that is a defined delay.

davecramer:
What exactly is this testing ?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
karenc-bq: > So you are saying that the META-INF service file isn't available at this point ?

No

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+import org.openjdk.jmh.annotations.Warmup;
+import org.postgresql.PGProperty;
+import software.amazon.jdbc.Driver;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
+
+@State(Scope.Benchmark)
+@Fork(1)
+@Warmup(iterations = 1)
+@Timeout(time = 10, timeUnit = TimeUnit.HOURS)
```
congoamz, line 49: nit: should we lower this timeout or no?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+  protected static Connection connectToInstance(final String url, final Properties props)
+      throws SQLException {
+    return DriverManager.getConnection(url, props);
+  }
+
+  protected static Properties initNoPluginPropsWithTimeouts() {
+    final Properties props = new Properties();
+    props.setProperty(PGProperty.USER.getName(), USERNAME);
+    props.setProperty(PGProperty.PASSWORD.getName(), PASSWORD);
+    props.setProperty(PGProperty.TCP_KEEP_ALIVE.getName(), Boolean.FALSE.toString());
```
congoamz, line 86: nit: not super important but this could be removed since the setting is false by default

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
joyc-bq: it could definitely be lower, I made it very high because I was getting an error with a lower timeout. It could probably be 60 minutes

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    // start all connection threads
+    for (Thread connectionThread : connectionThreadsList) {
+      connectionThread.start();
+    }
+
+    // stop all connection threads after finishing
+    for (Thread connectionThread : connectionThreadsList) {
+      try { connectionThread.join();
```
sergiyvamz, line 139: We can do it at line 135

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+    if (!org.postgresql.Driver.isRegistered()) {
+      org.postgresql.Driver.register();
+    }
+
+    if (!Driver.isRegistered()) {
+      Driver.register();
+    }
+  }
+
+  protected static Connection connectToInstance(final String url, final Properties props)
```
sergiyvamz, line 77: Why it's static?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+      throws SQLException {
+    return DriverManager.getConnection(url, props);
+  }
+
+  protected static Properties initNoPluginPropsWithTimeouts() {
```
sergiyvamz, line 82: Why static?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+    final Properties props = new Properties();
+    props.setProperty(PGProperty.USER.getName(), USERNAME);
+    props.setProperty(PGProperty.PASSWORD.getName(), PASSWORD);
+    props.setProperty(PGProperty.TCP_KEEP_ALIVE.getName(), Boolean.FALSE.toString());
+    return props;
+  }
+
+  protected static Properties initReadWritePluginProps() {
```
sergiyvamz, line 90: Why static?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingLoadBenchmarks.java
```diff
@@ -0,0 +1,171 @@
...
+    final Properties props = initNoPluginPropsWithTimeouts();
+    props.setProperty(PropertyDefinition.PLUGINS.name, "auroraHostList,readWriteSplitting");
+    return props;
+  }
+
+  protected static Properties initReadWritePluginLoadBalancingProps() {
```
sergiyvamz, line 96: Why static?

congoamz (approved):

PR #299, opened 2023-01-10
chore: update the project dependency in example subprojects

karenc-bq:
### Summary

The Hikari example project was using a specific released version of the driver instead of the latest changes in the project.


davecramer:
Looks good to me

sergiyvamz (approved):

PR #316, opened 2023-01-19
Add read-write benchmark and performance test results to readme

joyc-bq:
Added benchmark and aurora performance test results to readme and renamed another readme to specify the performance test it measured


docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
```diff
@@ -6,17 +6,3 @@
 | No Plugin Enabled                                             | 1,105,514.4 | ms/op |
 | Read-Write Splitting Plugin Enabled                           | 1,106,074.0 | ms/op |
 | Read-Write Splitting Plugin and Reader Load Balancing Enabled | 1,107,379.9 | ms/op |
-
```
davecramer: Why are we removing these ?

joyc-bq: We are investigating the numbers. The read-write plugin implementation is not finalized yet so we will add these performance results later

docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
```diff
@@ -0,0 +1,8 @@
+# Read-Write Splitting Plugin Performance Results
+
+## Benchmarks
+| Benchmark                                                     | Score       | Units |
+|---------------------------------------------------------------|-------------|-------|
+| No Plugin Enabled                                             | 1,105,514.4 | ms/op |
```
karenc-bq, line 6: 1 million ms? Should we use a different time unit?

karenc-bq, line 6: Are there any graphs like the ones we have for efm and failover performance tests?

davecramer: 1 million ms ?? that is 1000 seconds ... this surely can't be correct

docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
```diff
@@ -0,0 +1,8 @@
+# Read-Write Splitting Plugin Performance Results
+
+## Benchmarks
+| Benchmark                                                     | Score   | Units  |
+|---------------------------------------------------------------|---------|--------|
+| No Plugin Enabled                                             | 110.551 | sec/op |
```
sergiyvamz, line 6: I'd like to see an overhead numbers instead of raw performance numbers provided by benchmark tests. Something that look like this (numbers are not real):

No Plugin Enabled: 0 (baseline)
Read-Write Splitting Plugin Enabled: +30 ms
Read-Write Splitting Plugin and Reader Load Balancing Enabled: +50 ms

docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
joyc-bq: @davecramer the 1000 second numbers were the time it took to run the entire test 10 times. We should show the results for a single iteration so I divided that number by 10 to get results of ~110 seconds. Sergiy suggested below to change the results to show the overhead as the time difference between running the test with the plugin(s) and the baseline of no plugins enabled. Those numbers give a better understanding of the overhead caused by using the plugins

docs/development-guide/DevelopmentGuide.md
```diff
@@ -118,11 +118,20 @@ For more details on failure detection settings, see [here](../using-the-jdbc-dri
...
 ![](../images/jdbc_wrapper_postgresql_efm_30000_5000_3.png)
 ![](../images/jdbc_wrapper_postgresql_efm_6000_1000_1.png)
-See [here](PerformanceResults.md#enhanced-failure-monitoring-performance-with-different-failure-detection-configuration) for a more detailed performance breakdown.
+See [here](EFMAndFailoverPluginPerformanceResults.md#enhanced-failure-monitoring-performance-with-different-failure-detection-configuration) for a more detailed performance breakdown.
+
+#### Read-Write Splitting and Reader Load Balancing Performance Tests
+
+The diagrams in this section show the AWS JDBC Driver's performance when using the read-write splitting plugin, with or without reader load balancing. This test sets up a large number of connections in parallel; the initial connection is to the writer but will be changed to various reader instances if the plugin is enabled. The test executes a long query many times to simulate heavy queries.
+
+The average overhead time is measured as the average time difference between running the test when no plugins are enabled and running the test with the plugins enabled. The baseline number indicates there is no overhead time when no plugins are enabled.
```
congoamz, line 131: Reworded for a bit more clarity: `The average overhead time is measured as the average time difference between running the test with the read-write plugin and running the test without any plugins. The baseline overhead time is 0 because there are no plugins in this scenario and thus there is no plugin overhead.`

docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
```diff
@@ -0,0 +1,8 @@
+# Read-Write Splitting Plugin Performance Results
+
+### Read-Write Splitting Plugin Postgres Performance Results
+| Benchmark                                                     | Overhead Time | Units |
+|---------------------------------------------------------------|---------------|-------|
+| No Plugin Enabled                                             | 0 (baseline)  | ms/op |
+| Read-Write Splitting Plugin Enabled                           | +56           | ms/op |
```
sergiyvamz, line 7: Can you add percent as well.

docs/development-guide/ReadWriteSplittingPluginPerformanceResults.md
joyc-bq: done

aaron-congo (approved):

congoamz (approved):

karenc-bq (approved):

PR #321, opened 2023-01-25
fix: close aws secrets manager client

karenc-bq:
### Summary

Close SecretsManagerClient to prevent leaking PoolingHttpClientConnectionManager.

### Description

AWSSecretsManagerPlugin creates a new SecretsManagerClient for each connection, which creates a new PoolingHttpClientConnectionManager. The driver never closes these SecretsManagerClient so as the user application runs on the number of PoolingHttpClientConnectionManager instances just keep climbing.

### Additional Reviewers

@joyc-bq 
@aaronchung-bitquill 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import software.amazon.awssdk.regions.Region;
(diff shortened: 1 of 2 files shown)


sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -137,16 +140,6 @@ public Connection connect(
       final JdbcCallable<Connection, SQLException> connectFunc)
       throws SQLException {
 
-    if (this.secretsManagerClient == null || this.getSecretValueRequest == null) {
-      this.secretsManagerClient = SecretsManagerClient.builder()
-          .credentialsProvider(AwsCredentialsManager.getProvider(hostSpec, props))
```
congoamz, line 142: This line was added to support the custom credentials request that was recently merged in, I think we'll have to add it to the default secretsManagerClientFunc. It ensures that customers have a way to set their own credential provider

joyc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
karenc-bq: Good catch

PR #323, opened 2023-01-26
fix - ReconnectToWriterHandler success flag by fixing isCurrentHostWr…

aaronchung-bitquill:
…iter()

Fix isCurrentHostWrite() which the ReconnectToWriterHandler success flag uses. Added more checks on host aliases fields to make it more robust

### Summary

The ClusterAwareWriterFailoverHandler has a sub class ReconnectToWriterHandler. It returns a success flag. In some scenarios is was returning back success=false. This was because the success flag is based on the isCurrentHostWrite() method which was incorrectly returning false. It does this by comparing the aliases of the originalWriterHost to the newWriterHost. This fix changes the to use the HostSpec.allAlias fields to make it more robust in case different HostSpec objects pointing to the same host have some differing aliases fields. 

### Additional Reviewers

@sergiyv-improving @karenc-bq @aaron-congo @joyc-bq 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
@@ -293,14 +293,12 @@ public class ClusterAwareWriterFailoverHandler implements WriterFailoverHandler 
     }
 
     private boolean isCurrentHostWriter(final List<HostSpec> latestTopology) {
-      final Set<String> currentAliases = this.originalWriterHost.getAliases();
       final HostSpec latestWriter = getWriter(latestTopology);
-      if (currentAliases == null) {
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
```diff
@@ -293,14 +293,27 @@ public WriterFailoverResult call() {
     }
 
     private boolean isCurrentHostWriter(final List<HostSpec> latestTopology) {
-      final Set<String> currentAliases = this.originalWriterHost.getAliases();
       final HostSpec latestWriter = getWriter(latestTopology);
```
aaronchung-bitquill, line 296: Going to slim this down a bit to be simpler and less redundant

karenc-bq (approved):

PR #324, opened 2023-01-27
Add configuration for reader load balancing in autocommit-on mode

aaron-congo:
- User can configure reader load balancing to pick a new reader while autocommit is on after every X statements matching a given regex 
- By default, reader load balancing is disabled while autocommit is on
- If the user sets the autocommit-on statement limit to a positive value, the regex will match every SQL statement by default. This can be configured to another regex pattern

### DIFF
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
@@ -2,19 +2,6 @@
 
 The read-write splitting plugin adds functionality to switch between writer/reader instances via calls to the `Connection#setReadOnly` method. Upon calling `setReadOnly(true)`, the plugin will establish a connection to a random reader instance and direct subsequent queries to this instance. Future calls to `setReadOnly` will switch between the established writer and reader connections according to the boolean argument you supply to the `setReadOnly` method.
 
-### Session State Limitations with the Read-Write Splitting Plugin
-
-There are many session state attributes that can change during a session, and many ways to change them. Consequently, the read-write splitting plugin has limited support for transferring session state between connections. The following attributes will be automatically transferred when switching connections:
-
-- autocommit value
-- transaction isolation level
-
-All other session state attributes will be lost when switching connections. There are two scenarios when the plugin may switch a connection:
-1. You have loaded the plugin but have kept reader load balancing disabled. In this case, the connection will switch between the writer/reader when calling `setReadOnly`.
-2. You have loaded the plugin and have enabled reader load balancing. In this case, the connection will still switch between the writer/reader when calling `setReadOnly`. It will also switch at transaction boundaries. See the section on [reader load balancing](#reader-load-balancing) for more information on what is consided a transaction boundary.
-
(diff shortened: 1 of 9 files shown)


docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
```diff
@@ -23,48 +10,73 @@ final Properties properties = new Properties();
...
+
+#### Multi-statement SQL limitations with reader load balancing
+
+When reader load balancing is enabled and autocommit is off, the read-write splitting plugin will analyze SQL statements executed to determine if `COMMIT` or `ROLLBACK` are executed. In addition to `commit()` and `rollback()`, these statements indicate a transaction boundary. This analysis does not support SQL strings containing multiple statements. If your SQL strings use `COMMIT` or `ROLLBACK` and they contain multiple statements, we recommend that you do not enable reader load balancing as the plugin will not detect the transaction boundary if these keywords are not the first statement.
```
congoamz, line 56: I'm wondering if we should remove support for detecting transaction boundaries when `COMMIT` or `ROLLBACK` are executed as SQL. The JBDC spec says [here](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html) that `Applications should not invoke SQL commands directly to change the connection's configuration when there is a JDBC method available.` It seems silly and confusing to have to write this limitation paragraph for something that isn't even supported by the JDBC spec. Let me know what you think.

wrapper/src/main/java/software/amazon/jdbc/util/SqlMethodAnalyzer.java
```diff
@@ -84,16 +84,32 @@ public boolean doesCloseTransaction(final Connection conn, final String methodNa
...
     final String statement = getFirstSqlStatement(String.valueOf(args[0]));
-    return isStatementClosingTransaction(statement);
+    return isTransactionBoundary(statement);
   }
 
-  public boolean isExecuteDml(final String methodName, final Object[] args) {
+  public boolean isTransactionBoundary(String statement) {
+    return statement.startsWith("COMMIT")
+        || statement.startsWith("ROLLBACK")
+        || statement.startsWith("END");
```
congoamz, line 93: Can we remove "END"? I can't remember why this was included originally, as far as I can tell it isn't a command used to close transactions

davecramer: Because it is in PostgreSQL

davecramer: https://www.postgresql.org/docs/current/sql-end.html

docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
sergiyvamz: I'd keep it. Unfortunately developers don't read JDBC spec. Second line of defence makes sense to me.

congoamz: I see your point. However, to play devil's advocate:
- the mysql driver does not detect executing `COMMIT` or `ROLLBACK` so the expectation that it should work has not been set
- does it make sense to add code to account for scenarios that users shouldn't be following?
- if users complain we can point to the JDBC spec to show they are not following it
- removing this code will simplify the code and the documentation

sergiyvamz (approved):

congoamz:
We have decided not to merge this because the JDBC API does not provide a standard way to change the underlying connection used by a Statement object. Consequently, old Statement objects do not use the correct connection

PR #329, opened 2023-02-02
Integration test improvements

aaron-congo:
- always specify a failover target
- after failover, wait until cluster endpoint points to new IP and the new instance is verified as the writer
- include test names in report
- cleanup old test results before running
- make AWS_SESSION_TOKEN optional
- skip deauthorizing IP if ran against a pre-existing cluster


karenc-bq (approved):

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -621,8 +630,17 @@ protected void makeSureInstancesUp(List<String> instances, boolean finalCheck)
 
   public void failoverClusterAndWaitUntilWriterChanged(String clusterWriterId)
       throws InterruptedException {
-    failoverClusterAndWaitUntilWriterChanged(
-        TestEnvironment.getCurrent().getInfo().getAuroraClusterName(), clusterWriterId);
+    List<TestInstanceInfo> instances = TestEnvironment.getCurrent().getInfo().getDatabaseInfo().getInstances();
+    String anyReaderId = "";
+    for (TestInstanceInfo instance : instances) {
+      if (!clusterWriterId.equals(instance.getInstanceName())) {
+        anyReaderId = instance.getInstanceName();
```
sergiyvamz, line 637: anyReaderId = instance.getInstanceName();
break;

?

sergiyvamz (approved):

joyc-bq (approved):
lgtm

karenc-bq (approved):

.github/workflows/run-integration-tests.yml
```diff
@@ -12,8 +12,16 @@ concurrency:
 
 jobs:
   all-integration-tests:
-    name: 'Run integration tests on all environments'
+    name: 'Run integration tests on all Aurora environments'
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
```
sergiyvamz, line 19: I'd prefer to let integration suite to deal with all different environments. I see no point to have matrix here.

congoamz: This is a change that @karenc-bq suggested. One advantage is that the tests are ~20 minutes faster because mysql/pg are tested in parallel. Not sure if she had other advantages in mind

wrapper/src/test/java/integration/host/AuroraMysqlContainerTest.java
```diff
@@ -163,7 +163,7 @@ static void setUp() throws SQLException, InterruptedException, UnknownHostExcept
       container.start();
     }
     mysqlProxyPort =
-        containerHelper.createInstanceProxies(
+        containerHelper.getProxyPort(
```
sergiyvamz, line 166: I'm not sure the name gets better. getProxyPort() gives an idea what it returns but it doesn't highlight that a container instance is created behind the scene. 

createInstanceProxiesAndReturnPort() ?

congoamz: Ah, I didn't realize this method also creates proxies, I thought it was just determining the proxy port. I'll change the name

wrapper/src/test/java/integration/refactored/container/TestDriverProvider.java
```diff
@@ -92,6 +99,51 @@ public void beforeEach(ExtensionContext context) throws Exception {
                   // Enable all proxies
                   ProxyHelper.enableAllConnectivity();
                 }
+
+                if (TestEnvironment.getCurrent().getInfo().getRequest()
```
sergiyvamz, line 103: This logic doesn't need for most of the test scenarios. Would you consider to relocate it to MakeSureFirstInstanceWriterExtension? It might also need to add this extension to some other test classes.

sergiyvamz: Another option is to create another extension with this logic.

.github/workflows/run-integration-tests.yml
karenc-bq: Mostly to speed up the tests. MySQL tests are slower now so i think we actually save half an hour

sergiyvamz (approved):

PR #337, opened 2023-02-08
Password propagation

### LINKED ISSUE #336: Password change not picked up
### Describe the bug

When are using the wrapper in conjunction with Hikari and you change the username and password the Wrapper seems to be caching credentials and not changing them.
In our application we have rotating credentials and this causes a failure.
This is a regression in 1.0.1 from 1.0.0

### Expected Behavior

changing the username and password in hikari should update the AWS wrapper credentials

none

### Current Behavior

Hikari connections keep using old credentails

### Reproduction Steps

I have attached a gradle project with a unit test which fails with version 1.0.1 and passes with version 1.0.0
To run the test simply run 
`./gradlew test`
[bug.tar.gz](https://github.com/awslabs/aws-advanced-jdbc-wrapper/files/10673979/bug.tar.gz)

### Possible Solution

_No response_

_No response_

### The AWS Advanced JDBC Driver version used

1.0.1

### JDK version used

17.0.4

MacOS Ventura and Linux

karenc-bq:
### Summary

Ensure connection properties from the JDBC URL don't override existing credentials when reusing the AWSWrapperDataSource object.

Fixes #336.

### Description

When establishing new connections with the same AWSWrapperDataSource object but different credentials, the credentials get overwritten by the cached JDBC URL value. This is because the wrapper parses the cached JDBC URL for connection properties, and directly stores those properties onto the pre-existing properties, overriding all duplicate keys.

### Additional Reviewers

@joyc-bq
@brunos-bq
@aaronchung-bitquill 


wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -102,11 +103,13 @@ public Connection getConnection(String username, String password) throws SQLExce
...
           && !StringUtils.isNullOrEmpty(props.getProperty(this.urlPropertyName)))) {
+        final Properties parsedProperties = new Properties();
         if (!StringUtils.isNullOrEmpty(this.jdbcUrl)) {
-          parsePropertiesFromUrl(this.jdbcUrl, props);
+          parsePropertiesFromUrl(this.jdbcUrl, parsedProperties);
         } else {
-          parsePropertiesFromUrl(props.getProperty(this.urlPropertyName), props);
+          parsePropertiesFromUrl(props.getProperty(this.urlPropertyName), parsedProperties);
         }
+        props.forEach(parsedProperties::putIfAbsent);
```
congoamz, line 112: I'm confused, it looks like parsedProperties is never used. We store the properties from the URL in it, then we store the properties from props in it (if absent), but we don't do anything with it after. Is this line supposed to be flipped  (`parsedProperties.forEach(props::putIfAbsent)`)?

wrapper/src/test/java/software/amazon/jdbc/ds/AwsWrapperDataSourceTest.java
```diff
@@ -0,0 +1,122 @@
...
+    final List<String> urls = urlArgumentCaptor.getAllValues();
+    final List<Properties> properties = propertiesArgumentCaptor.getAllValues();
+    assertEquals(2, urls.size());
+    assertEquals(2, properties.size());
+    assertEquals(expectedUrl, urls.get(0));
+    assertEquals(expectedUrl, urls.get(1));
+    assertEquals(expectedProperties1, properties.get(0));
+    assertEquals(expectedProperties2, properties.get(1));
+  }
+}
```
congoamz, line 122: Missing newline

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
karenc-bq: Good catch, thanks

congoamz (approved):

PR #342, opened 2023-02-10
chore: move url scheme logic into DatabaseDialect WIP

davecramer:
Currently we have a lot of hard coded tests for postgres and mysql 
We also don't support mariadb everywhere.
This PR is a first pass at creating Dialects to encapsulate that logic into Dialect classes.
There is more work to be done as there will be differences between Aurora and other databases


wrapper/src/main/java/software/amazon/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,27 @@
+package software.amazon.jdbc.dialect;
```
karenc-bq, line 1: Please add license to new files.

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -46,12 +46,15 @@
 import software.amazon.jdbc.PluginServiceImpl;
 import software.amazon.jdbc.PropertyDefinition;
 import software.amazon.jdbc.cleanup.CanReleaseResources;
+import software.amazon.jdbc.dialect.*;
```
karenc-bq, line 49: Please do not use star imports

wrapper/src/main/java/software/amazon/jdbc/dialect/DatabaseDialect.java
davecramer: fixed

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
davecramer: fixed

wrapper/src/main/java/software/amazon/jdbc/dialect/DefaultDatabaseDialect.java
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.dialect;
+
+import software.amazon.jdbc.exceptions.ExceptionHandler;
+import software.amazon.jdbc.exceptions.GenericExceptionHandler;
+
+public class DefaultDatabaseDialect implements DatabaseDialect{
```
karenc-bq, line 22:
```suggestion
public class DefaultDatabaseDialect implements DatabaseDialect {
```

wrapper/src/test/java/software/amazon/jdbc/PluginServiceImplTests.java
```diff
@@ -47,12 +47,15 @@
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
+import software.amazon.jdbc.dialect.DatabaseDialect;
+import software.amazon.jdbc.dialect.TestDatabaseDialect;
+
```
karenc-bq, line 52:
```suggestion
```

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -139,15 +146,31 @@ protected void init(
...
-  protected String getProtocol(final String url) {
+  protected DatabaseDialect getDialect(final String url, Properties info) {
+    String dialectClassName = DATABASE_DIALECT.getString(info);
+    if (dialectClassName != null ){
+      try {
+        return (DatabaseDialect) Class.forName(dialectClassName).getDeclaredConstructor().newInstance();
+      }catch (Exception ex) {
+
+      }
+    }
```
karenc-bq, line 157:
```suggestion
    if (dialectClassName != null) {
      try {
        return (DatabaseDialect) Class.forName(dialectClassName).getDeclaredConstructor().newInstance();
      } catch (Exception ex) {

      }
    }
```

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -139,15 +146,31 @@ protected void init(
...
   }
 
-  protected String getProtocol(final String url) {
+  protected DatabaseDialect getDialect(final String url, Properties info) {
+    String dialectClassName = DATABASE_DIALECT.getString(info);
+    if (dialectClassName != null ){
+      try {
+        return (DatabaseDialect) Class.forName(dialectClassName).getDeclaredConstructor().newInstance();
+      }catch (Exception ex) {
+
```
karenc-bq, line 155: Are we intentionally ignoring the exception? We may want to throw an exception here.

wrapper/src/test/java/software/amazon/jdbc/dialect/TestDatabaseDialect.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.jdbc.dialect;
+
+public class TestDatabaseDialect extends DefaultDatabaseDialect implements DatabaseDialect {
+
+  private final String urlScheme;
+  public TestDatabaseDialect(String urlScheme){
```
karenc-bq, line 21:
```suggestion
  public TestDatabaseDialect(String urlScheme) {
```

wrapper/src/test/java/software/amazon/jdbc/dialect/TestDatabaseDialect.java
```diff
@@ -0,0 +1,28 @@
...
+    this.urlScheme = urlScheme;
+  }
+  @Override
```
karenc-bq, line 24:
```suggestion

  @Override
```

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/HostMonitoringConnectionPluginTest.java
```diff
@@ -49,17 +49,13 @@
...
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
-import software.amazon.jdbc.HostAvailability;
-import software.amazon.jdbc.HostSpec;
-import software.amazon.jdbc.JdbcCallable;
-import software.amazon.jdbc.NodeChangeOptions;
-import software.amazon.jdbc.OldConnectionSuggestedAction;
-import software.amazon.jdbc.PluginService;
+import software.amazon.jdbc.*;
```
karenc-bq, line 56: Please don't use star imports

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/JMHDatabaseDialect.java
```diff
@@ -0,0 +1,29 @@
...
+import software.amazon.jdbc.dialect.DefaultDatabaseDialect;
+
+public class JMHDatabaseDialect extends DefaultDatabaseDialect {
+  final String protocol;
+  public JMHDatabaseDialect(String protocol) {
+    this.protocol = protocol;
+  }
+  public String getURLScheme() {
+    return protocol;
+  }
```
karenc-bq, line 27:
```suggestion
  final String protocol;
  
  public JMHDatabaseDialect(String protocol) {
    this.protocol = protocol;
  }
  
  public String getURLScheme() {
    return protocol;
  }
```

wrapper/src/main/java/software/amazon/jdbc/dialect/MySQLDialect.java
```diff
@@ -0,0 +1,81 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.jdbc.dialect;
+
+import software.amazon.jdbc.exceptions.ExceptionHandler;
+import software.amazon.jdbc.exceptions.MySQLExceptionHandler;
+
+public class MySQLDialect implements DatabaseDialect{
```
karenc-bq, line 21:
```suggestion
public class MySQLDialect implements DatabaseDialect {
```

wrapper/src/main/java/software/amazon/jdbc/dialect/PostgreSQLDialect.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String PG_RETRIEVE_HOST_PORT_SQL =
+      "SELECT CONCAT(inet_server_addr(), ':', inet_server_port())";
+
+  private static final String POSTGRESQL_READONLY_QUERY = "SELECT pg_is_in_recovery() AS is_reader";
+  private static final String READ_ONLY_COLUMN_NAME = "is_reader";
+  private static final String PG_GET_INSTANCE_NAME_SQL = "SELECT aurora_db_instance_identifier()";
+  private static final String PG_INSTANCE_NAME_COL = "aurora_db_instance_identifier";
+  private static final int PG_PORT = 5432;
+  private static final String URL_SCHEME="jdbc:postgresql://";
+  @Override
```
karenc-bq, line 37:
```suggestion

  @Override
```

wrapper/src/main/java/software/amazon/jdbc/dialect/PostgreSQLDialect.java
```diff
@@ -0,0 +1,80 @@
...
+  public String getInstanceNameQuery() {
+    return PG_GET_INSTANCE_NAME_SQL;
+  }
+  public String getInstanceNameColumn() {
+    return PG_INSTANCE_NAME_COL;
+  }
+  @Override
```
karenc-bq, line 44: Can you please separate the methods? Same for other places

wrapper/src/main/java/software/amazon/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,53 @@
...
+import software.amazon.jdbc.AwsWrapperProperty;
+import software.amazon.jdbc.exceptions.ExceptionHandler;
+
+public interface DatabaseDialect {
+  public static final AwsWrapperProperty DATABASE_DIALECT =
+      new AwsWrapperProperty(
+          "databaseDialect",
+          "false",
+          "Set to true to automatically load-balance read-only transactions when setReadOnly is "
+              + "set to true");
```
karenc-bq, line 28: This description needs to be updated

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/HostMonitoringConnectionPlugin.java
```diff
@@ -289,17 +285,15 @@ private void generateHostAliases(
...
-          Messages.get(
-              "HostMonitoringConnectionPlugin.unsupportedDriverProtocol",
-              new Object[] {driverProtocol}));
+  private String getHostPortSql(final @NonNull DatabaseDialect dialect) {
+    if (dialect.isSupported()) {
+      return dialect.getHostPortQuery();
     }
+    throw new UnsupportedOperationException(
+        Messages.get(
+            "Dialect.unsupportedDriverProtocol",
```
karenc-bq, line 294: DatabaseDialect.unsupportedDriverProtocol

wrapper/src/main/resources/messages.properties
```diff
@@ -214,6 +214,7 @@ AuroraStaleDnsHelper.reset=Reset stored writer host.
 OpenedConnectionTracker.unableToPopulateOpenedConnectionQueue="[Internal Error] The driver is unable to track this opened connection because the instance endpoint is unknown."
 OpenedConnectionTracker.invalidatingConnections="Invalidating opened connections to host: ''{0}''"
 
+DatabaseDialect.unsupportedDriverProtocol=Driver protocol ''{0}'' is not supported.
```
karenc-bq, line 217: Can you please remove "ReadWriteSplittingPlugin.unsupportedDriverProtocol=Driver protocol ''{0}'' is not supported." from this file as well? Seems like we are no longer using it.

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -139,15 +146,34 @@ protected void init(
     }
   }
 
-  protected String getProtocol(final String url) {
+  protected DatabaseDialect getDialect(final String url, Properties info) {
+    String dialectClassName = DATABASE_DIALECT.getString(info);
```
sergiyvamz, line 150: One of the topic that has been discussed a long time ago was to avoid providing class names in a connection string. Configuration Profiles is a feature that allows load custom plugins for enterprise users and to avoid class names in a connection string. I'd suggest to consider using similar approach for dialect classes.

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -139,15 +146,34 @@ protected void init(
...
+    }
     final int index = url.indexOf("//");
     if (index < 0) {
       throw new IllegalArgumentException(
           Messages.get(
               "ConnectionWrapper.protocolNotFound",
               new Object[] {url}));
     }
-    return url.substring(0, index + 2);
+    switch (url.substring(0, index + 2)) {
```
sergiyvamz, line 166: I'd suggest to create a Dialect Manger class and hide this logic there.

davecramer: Ya, that was pretty ugly :)

wrapper/src/main/java/software/amazon/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public static DatabaseDialect getInstance() {
+    return new DefaultDatabaseDialect();
+  }
+
+  public String getInstanceNameQuery();
+
+  public String getInstanceNameColumn();
+
+  public boolean isSupported();
```
sergiyvamz, line 37: Need a better name that reflects what is supported. I can assume it's about supporting instance name query.

wrapper/src/main/java/software/amazon/jdbc/dialect/MySQLDialect.java
```diff
@@ -0,0 +1,88 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.dialect;
+
+import software.amazon.jdbc.exceptions.ExceptionHandler;
+import software.amazon.jdbc.exceptions.MySQLExceptionHandler;
+
+public class MySQLDialect implements DatabaseDialect {
```
sergiyvamz, line 22: This dialects makes sense for Aurora MySQL clusters. RDS MySQL and vanilla MySQL databases don't support queries mentioned here. Does it mean that there should be two separate dialects, one for Aurora, another one for MySQL?

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
sergiyvamz: It seems that protocol is not enough to determine what dialect class to create. As I noted above. Aurora dialect and vanilla MySQL dialects are different but they share the same protocol.

wrapper/src/main/java/software/amazon/jdbc/dialect/DatabaseDialect.java
davecramer: It's pretty much there just to cause failures for test cases.

wrapper/src/main/java/software/amazon/jdbc/dialect/MySQLDialect.java
davecramer: It did expose the MariaDB gap.

wrapper/src/main/java/software/amazon/jdbc/DriverConnectionProvider.java
```diff
@@ -40,28 +42,29 @@ public DriverConnectionProvider(final java.sql.Driver driver) {
...
-    final StringBuilder urlBuilder = new StringBuilder();
-    urlBuilder.append(protocol).append(hostSpec.getUrl()).append(databaseName);
+    final StringBuilder urlBuilder = new StringBuilder(databaseDialect.getURLScheme());
+    urlBuilder.append(hostSpec.getUrl()).append(databaseName);
 
     // In the case where we are connecting to MySQL using MariaDB driver,
     // we need to append "?permitMysqlScheme" to the connection URL
-    if (protocol.startsWith("jdbc:mysql:")
+    // TODO: this is a bit ugly
+    if (databaseDialect instanceof MariaDBDialect
```
karenc-bq, line 67: This should be `MySQLDialect`. It is for the case where we are connecting to a MySQL Engine with a MariaDB driver, and we are using the `"jdbc:mysql:"` protocol.

The `permitMysqlScheme` parameter is not required if we are using the `jdbc:mariadb` protocol directly

```suggestion
    if (databaseDialect instanceof MySQLDialect
```

davecramer: If we use jdbc:mysql we will load the mysql driver, no?

karenc-bq: This is for the scenario where only the mariadb driver is on the classpath. Similar to our AWS MySQL JDBC driver, the MariaDB driver supports the jdbc:mysql protocol too.

davecramer: I wasn't aware of that. However, I don't think we can make that presumption. That is I don't think the driver can assume that there is only one driver on the classpath.

karenc-bq: While I agree with you, this if clause isn't assuming there is only one driver on the classpath, it is to address the specific case where mariadb driver is used for the mysql protocol.

The other cases do not need this extra parameter `permitMysqlScheme`.

| Protocol     | Driver  | Requires permitMysqlScheme |
|--------------|---------|----------------------------|
| jdbc:mysql   | mysql   | false                      |
| jdbc:mysql   | mariadb | true                       |
| jdbc:mariadb | mariadb | false                      |

hsuamz:
Closing this PR, this has been included in: https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/372.

Issue #343, opened 2023-02-10
This method should be synchronized, or otherwise protect `initialized`


davecramer:
https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/d11a866e4616e24e90703ffff32b065f5e083b97/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java#L161

davecramer:
closed with #347

PR #347, opened 2023-02-15
Lock initialization of AuroraHostListProvider

aaron-congo:
Addresses #343 


davecramer (approved):

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -207,6 +215,7 @@ protected void init() throws SQLException {
     }
 
     this.isInitialized = true;
+    lock.unlock();
```
davecramer, line 218: It might make sense to create a utility lock that implements `Closable` and then wrap this in `try` block, either way it should be in a try block with the unlock in a `finally`

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -162,51 +164,63 @@ protected void init() throws SQLException {
...
-      } else {
-        final String clusterRdsHostUrl =
-            this.rdsHelper.getRdsClusterHostUrl(this.initialHostSpec.getUrl());
-        if (!StringUtils.isNullOrEmpty(clusterRdsHostUrl)) {
-          this.clusterId = this.clusterInstanceTemplate.isPortSpecified()
-              ? String.format("%s:%s", clusterRdsHostUrl, this.clusterInstanceTemplate.getPort())
-              : clusterRdsHostUrl;
-          this.isPrimaryClusterId = true;
+    try {
+      lock.lock();
```
sergiyvamz, line 168: Lock should be out of tr-finally block. We don't need to unlock if lock fails.

sergiyvamz (approved):

PR #354, opened 2023-02-23
Remove transaction splitting from the read-write splitting plugin

aaron-congo:
Removing this feature due to a JDBC limitation that makes it impractical. When Statements/ResultSets are created, they are internally bound to the database connection established at that moment. There is no standard JDBC functionality to switch out the connection used by these objects. Consequently, Statements/ResultSets continue to use their original connection even if the plugin internally switches on transactions.


wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
```diff
@@ -368,21 +359,6 @@ public void test_setReadOnlyTrue_allReadersDown() throws SQLException {
     }
   }
 
-  @TestTemplate
```
congoamz, line 371: I propose that we remove this test. Most of the functionality it covers in the plugin is already covered by test_setReadOnlyTrue_allReadersDown. The part that is not covered in other tests is highly driver/database dependent (whether an exception is thrown or not depends on if  the underlying driver tries to communicate with the database when setReadOnly(true) is called). Let me know if you disagree

sergiyvamz:
ReadWriteSplittingPlugin.getRandomReaderHost() line 517: There's a room for optimization. Shuffling entire collection is a solution but it's quite expensive. I'd recommend to get a random index and get an item by it.

sergiyvamz (approved):

sergiyvamz (approved):

PR #358, opened 2023-02-27
Allow selection of specific test classes/methods using test filters

aaron-congo:

### DIFF
--- a/wrapper/src/test/build.gradle.kts
+++ b/wrapper/src/test/build.gradle.kts
@@ -81,5 +81,10 @@ tasks.withType<Test> {
 }
 
 tasks.register<Test>("in-container") {
+    filter.excludeTestsMatching("software.*") // exclude unit tests
+    filter.excludeTestsMatching("integration.container.*") // exclude old integration tests
+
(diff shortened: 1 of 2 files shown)


joyc-bq (approved):

wrapper/src/test/build.gradle.kts
```diff
@@ -81,5 +81,10 @@ tasks.withType<Test> {
 }
 
 tasks.register<Test>("in-container") {
+    filter.excludeTestsMatching("software.*") // exclude unit tests
```
karenc-bq, line 84: this task runs the unit tests as well?

congoamz: With the exact code in this PR, this line isn't needed. However, if I use a filter like `*ReadWriteSplittingTests` I get a compilation error in one of the unit tests. This was unexpected but I think the * in the filter somehow triggers compilation of the unit tests. For reference, the error is `java.lang.ClassNotFoundException: software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest`

Issue #386, opened 2023-04-01
MySQL Aurora failover ends up on reader instead of writer


arosso:
### Describe the bug

We are seeing our connections to RDS MySQL Aurora 'failover' incorrectly to the reader node rather than the writer node when using the following configuration:

* JDBC Wrapper 1.0.2
* MySQL Connector 8.0.32
* Hikari 5.0.1
* A Custom Domain CNAME to the cluster writer (TTL = 1s)
* A jdbc url like this: jdbc:aws-wrapper:mysql://[custom-domain]:3306/db?clusterInstanceHostPattern=?.[clusterid].us-east-1.rds.amazonaws.com&wrapperLoggerLevel=ALL

Our test is similar to the one described in: https://github.com/awslabs/aws-mysql-jdbc/issues/377

We simply have a loop periodically getting the 1 and only connection from the connection pool, doing some transactions, releasing the connection. We then initiate a failover from the writer to a reader. We see this issue when the connection is idle in the pool and subsequently checked out. We then get this sort of error:

java.sql.SQLException: The MySQL server is running with the --read-only option so it cannot execute this statement

Attached is the anonymized debug log during the cut over.

When using the cluster writer endpoint directly the issue does not seem to happen.

### Expected Behavior

The failover should end up on the writer node.

### What plugins are used? What other connection properties were set?

Default plugins are used. No additional properties other than in the JDBC url.

### Current Behavior

The connection ends up on the reader instead of the writer.

### Reproduction Steps

See description. Attached is the debug log.
[wrapper.txt](https://github.com/awslabs/aws-advanced-jdbc-wrapper/files/11128169/wrapper.txt)

### The AWS Advanced JDBC Driver version used

1.0.2

### JDK version used

Corretto 11.0.17.8.1

### Operating System and version

Amazon Linux 2 (4.14.296-222.539.amzn2.x86_64)

karenc-bq:
Hi @arosso, we have merged PR #390 addressing this issue. Could you please checkout our [snapshot build](https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#using-a-snapshot-of-the-driver) and let us know if the issue persists? Thank you.

crystall-bitquill:
Hi @arosso,

[...]

PR #390, opened 2023-04-05
fix: fetch instance endpoint to the host alias  for custom domains

karenc-bq:
### Summary

Fetch the instance endpoint and add it to the host aliases for connections established using custom domains.

### Description

AuroraConnectionTrackerPlugin only looks up the instance endpoint for the current connection if the initial hostname is a cluster endpoint. 

If the connection is established via a custom domain or via the IP address, the AuroraConnectionTracker will not be able to track the opened connection since the tracker uses instance endpoints to track connections.
```
[2023-04-04 20:07:01.240150500] [FINEST ] software.amazon.jdbc.plugin.OpenedConnectionTracker populateOpenedConnectionQueue
 [Internal Error] The driver is unable to track this opened connection because the instance endpoint is unknown. 
[2023-04-04 20:07:01.259151100] [FINEST ] software.amazon.jdbc.plugin.OpenedConnectionTracker logOpenedConnections
 Opened Connections Tracked: 
[

] 
```

This behaviour leads to side effects where the user application may attempt to use an outdated connection in their workflows such as the issue raised in #386 

This change ensures the wrapper looks up the instance endpoints for other types of endpoints.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
@@ -59,7 +59,7 @@ public class AuroraConnectionTrackerPlugin extends AbstractConnectionPlugin {
   private final RdsUtils rdsHelper;
   private String retrieveInstanceQuery;
   private String instanceNameCol;
-  private String clusterInstanceTemplate;
+  private String clusterInstanceTemplate = "";
   private final OpenedConnectionTracker tracker;
 
   AuroraConnectionTrackerPlugin(PluginService pluginService, Properties props) {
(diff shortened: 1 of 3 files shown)


sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -99,7 +99,7 @@ public Connection connect(String driverProtocol, HostSpec hostSpec, Properties p
         : hostSpec;
 
     if (conn != null) {
-      if (rdsHelper.isRdsClusterDns(currentHostSpec.getHost())) {
+      if (!rdsHelper.isRdsInstance(currentHostSpec.getHost())) {
```
sergiyvamz, line 102: && currentHostSpec.getAliases().isEmpty()
?

PR #395, opened 2023-04-11
test: add logs to flakey test

karenc-bq:
### Summary

Add logs to flakey test so we can acquire more information when the test fails.


sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -402,6 +402,7 @@ public ReaderFailoverResult call() {
             () -> Messages.get(
                 "ClusterAwareReaderFailoverHandler.successfulReaderConnection",
                 new Object[] {this.newHost.getUrl()}));
+        LOGGER.fine("New reader connection object: " + conn);
```
brunos-bq, line 405: What's the goal/interest behind this log line?
Is the object supposed to have changed or remained? are we looking at any property on that object?

karenc-bq: To look at the hashcode of this object and to compare it with the connection we get after failover in AuroraFailoverTest.

PR #398, opened 2023-04-21
Fix: Mariadb hanging failover tests

joyc-bq:
Mariadb requires urlPropertyName set to the url, and permitMysqlScheme set to 1 in order to complete failover tests

readme updated to reflect this


docs/using-the-jdbc-driver/DataSource.md
```diff
@@ -18,15 +18,15 @@ while another might use the method `setUsername` for the same task. See the tabl
...
+| Database name               | `setDatabasePropertyName`      | The name of the database property.                                                                | `String` | No                                                                                                    | `databaseName`                         |
+| Port name                   | `setPortPropertyName`          | The name of the port property.                                                                    | `String` | No                                                                                                    | `port`                                 |
+| URL name                    | `setUrlPropertyName`           | The name of the URL property.                                                                     | `String` | Only required for MariaDB                                                                             | `url`                                  |
```
joyc-bq, line 26: this is the content change :  Only required for MariaDB    
the rest is formatting change so just ignore

crystall-bitquill (approved):

docs/using-the-jdbc-driver/DataSource.md
sergiyvamz: I'd keep "No" as a general option and add a note that some drivers, like MariaDb, require some parameters to be included in url so we recommend to provide this parameter.

sergiyvamz (approved):

PR #359, opened 2023-03-06
Implement internal connection pools for the read-write plugin

congoamz:
Add functionality to use internal connection pools to establish connections. The motivation for using pools comes from the read-write plugin's need to regularly establish new connections. Using pools will increase performance and decrease demands on the database by allowing connections to be re-used.

### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #402, opened 2023-04-21
docs: Remove reader load balancing from readme and benchmark results

joyc-bq:
Reader load balancing benchmark results removed from readme


docs/development-guide/DevelopmentGuide.md
```diff
@@ -125,9 +125,9 @@ See [here](PluginPipelinePerformanceResults.md#failover-performance-with-differe
...
 
-#### Read-Write Splitting and Reader Load Balancing Performance Tests
+#### Read-Write Splitting Performance Tests
 
-The diagram in this section shows the AWS JDBC Driver's performance when using the read-write splitting plugin, with or without reader load balancing. This test sets up a large number of connections in parallel; the initial connection is to the writer but will be changed to various reader instances if reader load balancing is enabled. The test executes a long query many times to simulate heavy queries.
+The diagram in this section shows the AWS JDBC Driver's performance when using the read-write splitting plugin. This test sets up a large number of connections in parallel; the initial connection is to the writer. The test executes a long query many times to simulate heavy queries.
 
 The average overhead time is measured as the average time difference between running the entire test with the read-write plugin and running the entire test without any plugins. The baseline overhead time is 0 because there are no plugins in this scenario and thus there is no plugin overhead.
```
congoamz, line 133: We will have to change the connected image as well since it has a bar for r/w with load balancing. Remind me where the test code for these results? I'm not sure if we should run the test with internal connection pooling instead of load balancing

congoamz:
Can you remove the change to docs/images/jdbc_wrapper_init_release_pipelines_benchmarks.png? I already replaced this image in #359 and included performance results from using internal connection pools

PR #416, opened 2023-04-28
Fix bug where topology query could execute indefinitely

aaron-congo:
The topology query is not monitored by the EFM plugin. If the user has no socket timeout and the topology query is executed against a down instance, it could execute indefinitely. This PR temporarily sets a socket timeout for the topology query if none was set by the user.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
@@ -30,6 +30,8 @@ import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Logger;
@@ -75,6 +77,7 @@ public class AuroraHostListProvider implements DynamicHostListProvider {
               + "This pattern is required to be specified for IP address or custom domain connections to AWS RDS "
               + "clusters. Otherwise, if unspecified, the pattern will be automatically created for AWS RDS clusters.");
 
+  private static final Executor NETWORK_TIMEOUT_EXECUTOR = Executors.newSingleThreadExecutor();
   private final HostListProviderService hostListProviderService;
   private final String originalUrl;
   private RdsUrlType rdsUrlType;
@@ -94,6 +97,7 @@ public class AuroraHostListProvider implements DynamicHostListProvider {
   public static final CacheMap<String, Boolean> primaryClusterIdCache = new CacheMap<>();
 
   private final ReentrantLock lock = new ReentrantLock();
+  private final int topologyQueryTimeoutMs = 5000;
   protected String clusterId;
   protected HostSpec clusterInstanceTemplate;
   protected ConnectionUrlParser connectionUrlParser;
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -333,11 +337,27 @@ protected List<HostSpec> queryForTopology(final Connection conn) throws SQLExcep
       this.topologyAwareDialect = (TopologyAwareDatabaseCluster) this.hostListProviderService.getDialect();
     }
 
+    int networkTimeout = -1;
+    try {
+      networkTimeout = conn.getNetworkTimeout();
+      // The topology query is not monitored by the EFM plugin, so it needs a socket timeout
+      if (networkTimeout == 0) {
```
congoamz, line 344: Should we override the user's socket timeout here even if it is set to a non-zero value?

karenc-bq: What is the motivation for that?

sergiyvamz: We should apply default 5000ms socket timeout just in case the current one is set to indefinite wait (it doesn't matter if this indefinite was a user explicit setting or just lack of setting)

congoamz: If they've set their socket timeout to accommodate their longest running queries it could be awhile before a problem is detected with the topology query. The topology query itself should be pretty quick to return. However, I lean towards keeping the code as is, I'd prefer to abide by the settings they gave us as much as possible. Just thought I'd bring it up to see if you agree

congoamz: @sergiyvamz won't the value returned by getNetworkTimeout be 0 regardless of if the user explicitly set it to zero or just didn't set it at all?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -333,11 +337,27 @@ protected List<HostSpec> queryForTopology(final Connection conn) throws SQLExcep
...
+        conn.setNetworkTimeout(NETWORK_TIMEOUT_EXECUTOR, defaultTopologyQueryTimeoutMs);
```
sergiyvamz, line 345: Is it synchronous? We shouldn't need to check executor to check it timeout is applied?

sergiyvamz (approved):

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
sergiyvamz: I assume it is. But it makes sense to verify.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -93,6 +103,8 @@ public class AuroraHostListProvider implements DynamicHostListProvider {
   public static final CacheMap<String, String> suggestedPrimaryClusterIdCache = new CacheMap<>();
   public static final CacheMap<String, Boolean> primaryClusterIdCache = new CacheMap<>();
 
+  private static final int defaultTopologyQueryTimeoutMs = 5000;
+  private static final int setNetworkTimeoutMs = 10000;
```
congoamz, line 107: Is this a reasonable timeout value for waiting for setNetworkTimeout to complete?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
congoamz: I just checked and the default value is 0 with all drivers

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
sergiyvamz: Setting timeout is a call to set a socket setting. It doesn't require any network activity like sending/receiving data. I'd expect that it takes few millis to set it but 10s is more than enough.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -333,11 +345,42 @@ protected List<HostSpec> queryForTopology(final Connection conn) throws SQLExcep
...
+    } finally {
+      if (networkTimeout == 0 && !conn.isClosed()) {
+        setNetworkTimeout(conn, networkTimeout);
+      }
+    }
+  }
+
+  private void setNetworkTimeout(final Connection conn, final int timeoutMs) throws SQLException {
+    Future<Boolean> future = NETWORK_TIMEOUT_COMPLETION_SERVICE.submit(() -> {
+      conn.setNetworkTimeout(NETWORK_TIMEOUT_EXECUTOR, timeoutMs);
```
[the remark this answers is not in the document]
sergiyvamz: That looks a bit over-engineered. Why not to use something similar to this?
https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/task/SyncTaskExecutor.java

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
congoamz: Should I lower it to a different value then?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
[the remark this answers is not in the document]
congoamz: Runnable#run and Callable#call do not allow you to set a timeout for the task. Is it okay to use them without a timeout here?

sergiyvamz (approved):

PR #432, opened 2023-05-05
Improve internal connection pool key defaults

congoamz:

### DIFF
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
@@ -39,7 +39,7 @@ The wrapper driver currently uses [Hikari](https://github.com/brettwooldridge/Hi
 - username
 - password
 
-You can optionally pass in a `HikariPoolMapping` function as a second parameter to the `HikariPooledConnectionProvider`. This allows you to decide when new connection pools should be created by defining what is included in the pool map key. A new pool will be created each time a new connection is requested with a unique key. By default, a new pool will be created for each unique instance-user combination. If you would like to define a different key system, you should pass in a `HikariPoolMapping` function defining this logic. Note that the user will always be automatically included in the key for security reasons.
+You can optionally pass in a `HikariPoolMapping` function as a second parameter to the `HikariPooledConnectionProvider`. This allows you to decide when new connection pools should be created by defining what is included in the pool map key. A new pool will be created each time a new connection is requested with a unique key. By default, a new pool will be created for each unique instance-user combination. If you would like to define a different key system, you should pass in a `HikariPoolMapping` function defining this logic. Note that the user will always be automatically included in the key for security reasons. Please see [ReadWriteSplittingPostgresExample.java](../../../examples/AWSDriverExample/src/main/java/software/amazon/ReadWriteSplittingPostgresExample.java) for an example of how to configure the pool map key.
 
 2. Call `ConnectionProviderManager.setConnectionProvider`, passing in the `HikariPooledConnectionProvider` you created in step 1.
 
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
@@ -39,7 +39,7 @@ The wrapper driver currently uses [Hikari](https://github.com/brettwooldridge/Hi
 - username
 - password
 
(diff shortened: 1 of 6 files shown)


wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
```diff
@@ -140,6 +140,14 @@ public Connection connect(
     return null;
   }
 
+  // The pool key should always be retrieved using this method, because the username/password
+  // must always be included to avoid sharing privileged connections with other users.
+  private String getPoolKey(HostSpec hostSpec, Properties props) {
+    return poolMapping.getKey(hostSpec, props)
```
sergiyvamz, line 146: StringBuilder or String.format() would be more efficient in this case.

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
```diff
@@ -140,6 +140,14 @@ public Connection connect(
...
+        + props.getProperty(PropertyDefinition.USER.name)
+        + props.getProperty(PropertyDefinition.PASSWORD.name);
```
sergiyvamz, line 148: I wouldn't include password. In case of IAM token (that is changing every15min) we might see more and more connection pools are created.

wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
```diff
@@ -758,4 +760,75 @@ public void test_pooledConnection_failoverInTransaction()
       ConnectionProviderManager.resetProvider();
     }
   }
+
+  @TestTemplate
+  @EnableOnTestFeature(TestEnvironmentFeatures.FAILOVER_SUPPORTED)
```
sergiyvamz, line 765: This test practically test how well getPoolKey() works. I don't think we need such test as integration test with real database. A simple unit test with mocked connection pools would be more than enough.

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
congoamz: This stack overflow post recommends using string concatenation instead of string builder for simple strings as the compiler automatically converts concatenation to string builder internally. Take a look and let me know if you disagree

https://stackoverflow.com/questions/11942368/why-use-stringbuilder-explicitly-if-the-compiler-converts-string-concatenation-t#:~:text=%40JohanSj%C3%B6berg%20yes%2C%20in%20fact%20it,append(result).

joyc-bq (approved):

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -762,6 +761,25 @@ public String executeInstanceIdQuery(DatabaseEngine databaseEngine, Statement st
     return null;
   }
 
+  public void createUser(Connection conn, String username, String password) throws SQLException {
+    DatabaseEngine engine = TestEnvironment.getCurrent().getInfo().getRequest().getDatabaseEngine();
+    String dropUserSql = getCreateUserSql(engine, username, password);
+    try (Statement stmt = conn.createStatement()) {
+      stmt.execute(dropUserSql);
```
karenc-bq, line 768: you can also put .execute in the try-with-resources block to auto close the result set

karenc-bq: Same comment for other places

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
sergiyvamz: I decompiled class code from jar and I don't see any StringBuilder there.

sergiyvamz (approved):

docs/using-the-jdbc-driver/using-plugins/UsingTheReadWriteSplittingPlugin.md
```diff
@@ -39,7 +39,7 @@ The wrapper driver currently uses [Hikari](https://github.com/brettwooldridge/Hi
...
+You can optionally pass in a `HikariPoolMapping` function as a second parameter to the `HikariPooledConnectionProvider`. This allows you to decide when new connection pools should be created by defining what is included in the pool map key. A new pool will be created each time a new connection is requested with a unique key. By default, a new pool will be created for each unique instance-user combination. If you would like to define a different key system, you should pass in a `HikariPoolMapping` function defining this logic. Note that the user will always be automatically included in the key for security reasons.
```
karenc-bq, line 42: Can we provide an example?

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
congoamz: Weird, if you google "does java automatically use stringbuilder" there are a bunch of results saying its automatically converted to use StringBuilder. Regardless I'll just change it at this point

karenc-bq (approved):

karenc-bq (approved):

PR #435, opened 2023-05-10
Remove isMultiWriterCluster flag and ignore stale records when querying the topology

crystall-bitquill:
### Summary

Remove the isMultiWriterCluster flag from the AuroraTopologyService and add checks for stale topology records.

### Description

Multi writer clusters are not longer supported, so the isMultiWriterCluster flag that was used to disable failover has been removed. Additionally, checks have been added to ensure stale writer records that are obtained post writer-failover are not being used.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
@@ -16,6 +16,9 @@
 
 package software.amazon.jdbc;
 
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
```diff
@@ -39,6 +42,8 @@ public class HostSpec {
   protected Set<String> allAliases = ConcurrentHashMap.newKeySet();
   protected long weight; // Greater or equal 0. Lesser the weight, the healthier node.
   protected String hostId;
+  protected String lastUpdateTime;
```
sergiyvamz, line 45: Can we use date object instead of String?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraHostListProvider.java
```diff
@@ -361,22 +367,34 @@ private List<HostSpec> processQueryResults(final ResultSet resultSet) throws SQL
...
 
+    int writerCount = writers.size();
+
     if (writerCount == 0) {
       LOGGER.severe(
           () -> Messages.get(
               "AuroraHostListProvider.invalidTopology"));
       hosts.clear();
+    } else if (writerCount == 1) {
+      hosts.add(writers.get(0));
```
sergiyvamz, line 389: If the wrapper driver we don't have a rule that a writer node should be at position 0. You can just add a writer to the host collection.

sergiyvamz (approved):

PR #452, opened 2023-05-17
feat: introduce target driver dialects

sergiyvamz:
### Summary

Introduce target driver dialects

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@davecramer 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
@@ -78,6 +78,10 @@ public class AwsSecretsManagerConnectionPlugin extends AbstractConnectionPlugin 
   private Secret secret;
   protected PluginService pluginService;
 
+  static {
+    PropertyDefinition.registerPluginProperties(AwsSecretsManagerConnectionPlugin.class);
+  }
+
   public AwsSecretsManagerConnectionPlugin(final PluginService pluginService, final Properties props) {
 
     this(
--- a/wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.targetdriverdialect;
+
+import static software.amazon.jdbc.util.ConnectionUrlBuilder.buildUrl;
+import static software.amazon.jdbc.util.StringUtils.isNullOrEmpty;
+
+import java.sql.Driver;
+import java.sql.SQLException;
+import java.util.Properties;
+import java.util.logging.Logger;
+import javax.sql.DataSource;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.util.PropertyUtils;
+
+public class GenericTargetDriverDialect implements TargetDriverDialect {
+
+  private static final Logger LOGGER =
+      Logger.getLogger(GenericTargetDriverDialect.class.getName());
+
+  @Override
+  public boolean isDialect(Driver driver) {
+    return true;
+  }
+
+  @Override
+  public boolean isDialect(String dataSourceClass) {
+    return true;
+  }
+
+  @Override
+  public ConnectInfo prepareConnectInfo(final @NonNull String protocol,
+      final @NonNull HostSpec hostSpec,
+      final @NonNull Properties props) throws SQLException {
+
+    final String databaseName =
+        PropertyDefinition.DATABASE.getString(props) != null
+            ? PropertyDefinition.DATABASE.getString(props)
+            : "";
+    String urlBuilder = protocol + hostSpec.getUrl() + databaseName;
+
+    // keep unknown properties (the ones that don't belong to AWS Wrapper Driver)
+    // and use them to make a connection
+    PropertyDefinition.removeAllExceptCredentials(props);
+    return new ConnectInfo(urlBuilder, props);
+  }
+
+  @Override
+  public void prepareDataSource(
+      final @NonNull DataSource dataSource,
+      final @NonNull String protocol,
+      final @NonNull HostSpec hostSpec,
+      final @NonNull Properties props,
+      final @Nullable String serverPropertyName,
+      final @Nullable String portPropertyName,
+      final @Nullable String urlPropertyName,
+      final @Nullable String databasePropertyName) throws SQLException {
+
+    final Properties customProps = new Properties();
+
+    if (!isNullOrEmpty(serverPropertyName)) {
+      customProps.setProperty(serverPropertyName, hostSpec.getHost());
+    }
+
+    if (hostSpec.isPortSpecified() && !isNullOrEmpty(portPropertyName)) {
+      customProps.put(portPropertyName, hostSpec.getPort());
+    }
+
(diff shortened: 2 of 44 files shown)


wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgDataSourceHelper.java
```diff
@@ -0,0 +1,59 @@
...
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.PropertyUtils;
+
+public class PgDataSourceHelper {
+
+  public void prepareDataSource(
+      final @NonNull DataSource dataSource,
+      final @NonNull String protocol,
```
aaronchung-bitquill, line 33: `protocol` is unused here.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
```diff
@@ -0,0 +1,117 @@
...
+          hostSpec,
+          serverPropertyName,
+          portPropertyName,
+          databasePropertyName,
+          urlProperties);
+      LOGGER.finest(() -> "Connecting to " + finalUrl);
+      customProps.setProperty(urlPropertyName, finalUrl);
+    }
+
+    LOGGER.finest(() -> PropertyUtils.logProperties(customProps, "Connecting with properties: \n"));
```
karenc-bq, line 110: Can we log the properties and the url in one message? It helps the debugging process in multithreaded applications

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/MariadbDataSourceHelper.java
```diff
@@ -0,0 +1,75 @@
...
+  public void prepareDataSource(
+      final @NonNull DataSource dataSource,
+      final @NonNull String protocol,
+      final @NonNull HostSpec hostSpec,
+      final @NonNull Properties props) throws SQLException {
+
+    if (!(dataSource instanceof MariaDbDataSource)) {
+      throw new SQLException(Messages.get(
+          "TargetDriverDialectManager.unexpectedClass",
+          new Object[] {"org.mariadb.jdbc.MariaDbDataSource", dataSource.getClass().getName()}));
```
karenc-bq, line 47: MariaDbDataSource.getClass().getName() in case the package name ever changes?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
```diff
@@ -0,0 +1,63 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.HostSpec;
+
+public class PgTargetDriverDialect extends GenericTargetDriverDialect {
+
+  private static final Set<String> dataSourceClassMap = new HashSet<>(Arrays.asList(
+          "org.postgresql.ds.PGSimpleDataSource",
+          "org.postgresql.ds.PGPoolingDataSource",
+          "org.postgresql.ds.PGConnectionPoolDataSource"));
```
aaronchung-bitquill, line 35: Similar to Karens comment. Can we use `.getClass().getName()` in case package name or whatever changes?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
```diff
@@ -0,0 +1,63 @@
...
+      final @NonNull String protocol,
+      final @NonNull HostSpec hostSpec,
+      final @NonNull Properties props,
+      final @Nullable String serverPropertyName,
+      final @Nullable String portPropertyName,
+      final @Nullable String urlPropertyName,
+      final @Nullable String databasePropertyName) throws SQLException {
+
+    // The logic is isolated to a separated class since it uses
+    // direct reference to org.postgresql.ds.common.BaseDataSource
```
aaronchung-bitquill, line 59: Referring to the code comments above, is the idea to decouple the Dialect class from `org.postgresql.ds.common.BaseDataSource`?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
sergiyvamz: Hmm they are different properties. Url may be missed. Let's find a separate solution for multithreaded logs.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
sergiyvamz: That's correct. The idea is to decouple different dialect classes. Since `PgTargetDriverDialect` is instantiated in `TargetDriverDialectManager`, it's impossible to use any classes from `org.postgresql` package in `PgTargetDriverDialect`. In this case AWS JDBC Driver would have a dependency on `org.postgresql` package. That's why all logic is isolated in `PgDataSourceHelper` class. This helper class has direct references to `org.postgresql` package but it's instantiated when it's needed only, i.e. when the target driver is actually `pgjdbc` driver.

aaronchung-bitquill: Thank you for the explanation!

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/TargetDriverDialectManager.java
```diff
@@ -0,0 +1,124 @@
...
+  public static final AwsWrapperProperty TARGET_DRIVER_DIALECT = new AwsWrapperProperty(
+      "wrapperTargetDriverDialect", "",
+      "A unique identifier for the target driver dialect.");
+
+  /**
+   * Every Dialect implementation SHOULD BE stateless!!!
+   * Dialect objects are shared between different connections.
+   * The order of entries in this map is the order in which dialects are called/verified.
+   */
+  protected static final Map<String, TargetDriverDialect> knownDialectsByCode =
```
aaronchung-bitquill, line 47: Nit:  would it be better to have this mapping located in the `TargetDriverDialectCodes` class? That way if a new dialect is introduced, it is one less class that needs to be updated, and it is harder for the developer to forget adding this mapping. 

Also sorry about the lateness of the comment.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/TargetDriverDialectCodes.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.targetdriverdialect;
+
+public class TargetDriverDialectCodes {
+  public static final String PG_JDBC = "pgjdbc";
+  public static final String MYSQL_CONNECTOR_J = "mysql-connector-j";
+  public static final String MARIADB_CONNECTOR_J_VER_3 = "mariadb-connector-j-3"; // ver 3.0.3+
```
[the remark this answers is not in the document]
sergiyvamz: There's a chance that different versions of the same driver may require different dialects. MariaDb driver changed behaviour in version 3+. So I think it makes sense to be explicit in defining dialect codes.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
```diff
@@ -0,0 +1,117 @@
...
+      final @Nullable String databasePropertyName) throws SQLException {
+
+    final Properties customProps = new Properties();
+
+    if (!isNullOrEmpty(serverPropertyName)) {
+      customProps.setProperty(serverPropertyName, hostSpec.getHost());
+    }
+
+    if (hostSpec.isPortSpecified() && !isNullOrEmpty(portPropertyName)) {
+      customProps.put(portPropertyName, hostSpec.getPort());
```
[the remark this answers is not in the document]
sergiyvamz: Technically it's the same call to put(). I'll fix it for consistency.

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -78,6 +78,10 @@ public class AwsSecretsManagerConnectionPlugin extends AbstractConnectionPlugin
   private Secret secret;
   protected PluginService pluginService;
 
+  static {
+    PropertyDefinition.registerPluginProperties(AwsSecretsManagerConnectionPlugin.class);
```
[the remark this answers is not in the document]
sergiyvamz: Registering properties is direct responsibility of a particular plugin. Register properties for all know plugins in ConnectionPluginManager doesn't look like direct responsibilities of the this class. Also, such approach doesn't cover user custom plugins. I'd prefer to let plugins be responsible for such registration.

PR #454, opened 2023-05-18
add connection pooling test to rw splitting heavy query test: draft only

joyc-bq:
Code will not be merged

Removed reader load balancing testing 

Added connection pooling test to rw splitting heavy query test


benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingConnectionPoolingBenchmarks.java
```diff
@@ -0,0 +1,165 @@
...
+      Driver.register();
+    }
+  }
+
+  protected Connection connectToInstance(final String url, final Properties props)
+      throws SQLException {
+    return DriverManager.getConnection(url, props);
+  }
+
+  protected Properties initNoPluginPropsWithTimeouts() {
```
congoamz, line 83: nit: the method name mentions there are timeouts but there aren't any in the implementation. We should either change the name or add timeouts. Since the previous version of this code did not have timeouts here either I'm guessing they are not needed

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingConnectionPoolingBenchmarks.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  private Thread getThread_PGReadWriteSplitting(final Properties props) {
+    return new Thread(() -> {
+      try (Connection conn = connectToInstance(POSTGRESQL_CONNECTION_STRING, props);
+           Statement stmt1 = conn.createStatement()) {
+        stmt1.executeQuery(QUERY_1);
+
+        // switch to reader if read-write splitting plugin is enabled
+        conn.setReadOnly(true);
```
congoamz, line 103: I'd say let's execute one query on the reader as well

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/ReadWriteSplittingConnectionPoolingBenchmarks.java
joyc-bq: thanks!

PR #466, opened 2023-05-25
[Do Not Merge] feat: refresh IAM tokens with a background thread

karenc-bq:


wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+              new Object[] {token}));
+      PropertyDefinition.PASSWORD.set(props, token);
+      tokenCache.put(
+          cacheKey,
+          new TokenInfo(token, Instant.now().plus(tokenExpirationSec, ChronoUnit.SECONDS)));
+      final String threadHost = host;
+      final int threadPort = port;
+      cacheThead.put(
+          cacheKey,
+          executorService.submit(() -> {
```
raymond-lum, line 124: Is this adding a new thread for every connection? If so, that wouldn't be ideal. We should rather have a single thread for the entire instance that is spawned on instantiation that loops every prefetch time (OR some time less than iamExpiration) and goes through every key in cacheThread on each loop and checks for expirations).

Then when the destructor for this class is called the "while (true)" should be replaced with a "while (isDestroyed)" that is set to true by the destructor of this class to clean up the thread so it doesn't leak.

Lastly, when doing while loops within a thread, you should sleep otherwise this will hammer the CPU time of the client. I would expect to sleep < tokenExpirationSec and that the trigger to call generateAuthenticationToken is also less than when the token actually expires. If you made those the same there would be a possible race condition so I think this needs to wake up just a bit more frequent than expirations could occur.

What the above does:
1. Uses less resources because only a single thread.
2. Ensures we aren't leaking threads on close.
3. Ensure we aren't wasting CPU cycles (because we would sleep the thread at some time just < expiration).

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.JdbcCallable;
+import software.amazon.jdbc.PluginService;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.StringUtils;
+
+public class IamAuthMultithreadedConnectionPlugin extends IamAuthConnectionPlugin {
```
sergiyvamz, line 38: IamAuthPrefetchConnectionPlugin ?

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+          props,
+          host,
+          port,
+          region);
+      LOGGER.finest(
+          () -> Messages.get(
+              "IamAuthConnectionPlugin.generatedNewIamToken",
+              new Object[] {token}));
+      PropertyDefinition.PASSWORD.set(props, token);
+      tokenCache.put(
```
sergiyvamz, line 117: putIfAbsent() ?

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+          cacheKey,
+          new TokenInfo(token, Instant.now().plus(tokenExpirationSec, ChronoUnit.SECONDS)));
+      final String threadHost = host;
+      final int threadPort = port;
+      cacheThead.put(
```
sergiyvamz, line 122: putIfAbsent() ?

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+          cacheKey,
+          executorService.submit(() -> {
+            while (true) {
```
sergiyvamz, line 125: I'd do a simpler logic:

while(true) {
  sleep(tokenExpirationTime)
  generateNewToken();
  putTokenToCache()
}

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+                    hostSpec,
+                    props,
+                    threadHost,
+                    threadPort,
+                    region);
+                LOGGER.finest(
+                    () -> Messages.get(
+                        "IamAuthConnectionPlugin.generatedNewIamTokenInThread",
+                        new Object[] {threadToken}));
+                PropertyDefinition.PASSWORD.set(props, threadToken);
```
sergiyvamz, line 137: I'd let the main thread to update password in properties.

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+import software.amazon.jdbc.JdbcCallable;
+import software.amazon.jdbc.PluginService;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.StringUtils;
+
+public class IamAuthMultithreadedConnectionPlugin extends IamAuthConnectionPlugin {
+  private static final Logger LOGGER = Logger.getLogger(IamAuthMultithreadedConnectionPlugin.class.getName());
+
+  static final ConcurrentHashMap<String, TokenInfo> tokenCache = new ConcurrentHashMap<>();
```
sergiyvamz, line 41: We can use just usual HashMap. It's not important for items in this cache to expire. We can rely on background thread that refresh them and keep valid.

sergiyvamz, line 41: ConcurrentHashMap<String, String> ?

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthPrefetchConnectionPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.StringUtils;
+
+public class IamAuthPrefetchConnectionPlugin extends IamAuthConnectionPlugin implements
+    CanReleaseResources {
+
+  private static final Logger LOGGER = Logger.getLogger(IamAuthPrefetchConnectionPlugin.class.getName());
+
+  static final Map<String, String> tokenCache = new HashMap<>();
+  static final ConcurrentHashMap<String, Future<?>> cacheThead = new ConcurrentHashMap<>();
```
raymond-lum, line 48:
```suggestion
  static final ConcurrentHashMap<String, Future<?>> cacheThread = new ConcurrentHashMap<>();
```

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthPrefetchConnectionPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.StringUtils;
+
+public class IamAuthPrefetchConnectionPlugin extends IamAuthConnectionPlugin implements
+    CanReleaseResources {
+
+  private static final Logger LOGGER = Logger.getLogger(IamAuthPrefetchConnectionPlugin.class.getName());
+
+  static final Map<String, String> tokenCache = new HashMap<>();
+  static final ConcurrentHashMap<String, Future<?>> cacheThread = new ConcurrentHashMap<>();
```
raymond-lum, line 48: Need to adjust all other cacheThead.

wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthMultithreadedConnectionPlugin.java
hsuamz: 1. To clarify, the IAM plugin uses 1 thread per cache key, not for every connection. This cache key is the connection URL. So if the user decides to use an instance endpoint, then monitoring thread creates 1 thread for that instance endpoint. If users makes more connections to the same endpoint, the plugin uses the same thread.

2, 3. Before we make any conclusion, we're investigating through a bug identified that is impacting performance measurements right now.

PR #472, opened 2023-05-30
[wip] Support internal connection pools for cluster URLs

congoamz:

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
@@ -140,7 +140,7 @@ public class HikariPooledConnectionProvider implements PooledConnectionProvider,
   public boolean acceptsUrl(
       @NonNull String protocol, @NonNull HostSpec hostSpec, @NonNull Properties props) {
     final RdsUrlType urlType = rdsUtils.identifyRdsType(hostSpec.getHost());
-    return RdsUrlType.RDS_INSTANCE.equals(urlType);
+    return urlType.isRds();
   }
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
@@ -51,6 +51,7 @@ public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
   private static final Set<String> subscribedMethods =
       Collections.unmodifiableSet(new HashSet<String>() {
         {
+          // All network methods need to be added so that the plugin can monitor them for failover exceptions
           addAll(SubscribedMethodHelper.NETWORK_BOUND_METHODS);
           add("initHostProvider");
           add("connect");
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.jdbc.AwsWrapperProperty;
+import software.amazon.jdbc.ConnectionProviderManager;
 import software.amazon.jdbc.HostListProviderService;
 import software.amazon.jdbc.HostRole;
 import software.amazon.jdbc.HostSpec;
@@ -40,6 +41,7 @@ import software.amazon.jdbc.plugin.AbstractConnectionPlugin;
 import software.amazon.jdbc.plugin.failover.FailoverSQLException;
 import software.amazon.jdbc.util.Messages;
 import software.amazon.jdbc.util.SqlState;
+import software.amazon.jdbc.util.SubscribedMethodHelper;
 import software.amazon.jdbc.util.WrapperUtils;
 
 public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
@@ -49,10 +51,11 @@ public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
   private static final Set<String> subscribedMethods =
       Collections.unmodifiableSet(new HashSet<String>() {
         {
+          addAll(SubscribedMethodHelper.NETWORK_BOUND_METHODS);
           add("initHostProvider");
           add("connect");
+          add("forceConnect");
           add("notifyConnectionChanged");
-          add("Connection.setReadOnly");
         }
       });
   static final String METHOD_SET_READ_ONLY = "Connection.setReadOnly";
@@ -214,6 +217,7 @@ public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
         LOGGER.finer(
             () -> Messages.get("ReadWriteSplittingPlugin.failoverExceptionWhileExecutingCommand",
                 new Object[] {methodName}));
+        ConnectionProviderManager.releaseResources();
         closeIdleConnections();
       } else {
         LOGGER.finest(
--- a/wrapper/src/test/java/integration/refactored/container/tests/AutoscalingTests.java
+++ b/wrapper/src/test/java/integration/refactored/container/tests/AutoscalingTests.java
@@ -109,6 +109,9 @@ public class AutoscalingTests {
   @TestTemplate
   public void test_pooledConnectionAutoScaling_setReadOnlyOnOldConnection()
       throws SQLException, InterruptedException {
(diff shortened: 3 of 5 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -214,6 +218,7 @@ public <T, E extends Exception> T execute(
         LOGGER.finer(
             () -> Messages.get("ReadWriteSplittingPlugin.failoverExceptionWhileExecutingCommand",
                 new Object[] {methodName}));
+        ConnectionProviderManager.releaseResources();
```
congoamz, line 221: We could move this to the failover plugin as an alternative

sergiyvamz, line 221: It doesn't look like a responsibility of this plugin. Need to discuss.

sergiyvamz: Is there any interference here with connection tracker plugin?

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
```diff
@@ -140,7 +140,7 @@ public HikariPooledConnectionProvider(
   public boolean acceptsUrl(
       @NonNull String protocol, @NonNull HostSpec hostSpec, @NonNull Properties props) {
     final RdsUrlType urlType = rdsUtils.identifyRdsType(hostSpec.getHost());
-    return RdsUrlType.RDS_INSTANCE.equals(urlType);
+    return urlType.isRds();
```
sergiyvamz, line 143: What would be a connection pool key in case of cluster endpoint?

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
aaron-congo: Would it make more sense to move it to the failover plugin then?
Can you remind me what the connection tracker does and how this might interfere with it?

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
aaron-congo: The key would be similar to an instance endpoint. PoolKey.url would be the cluster URL and PoolKey.extraKey would be the user (unless a custom HikariPoolMapping is set)

wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
```diff
@@ -645,30 +658,63 @@ public void test_pooledConnectionFailover() throws SQLException, InterruptedExce
...
     ConnectionProviderManager.setConnectionProvider(provider);
 
     final Connection initialWriterConn;
     final Connection newWriterConn;
+    final String originalWriterId;
+    final String nextWriterId;
+
+    List<Connection> setupConnections = new ArrayList<>();
+    for (int i = 0; i < maxPoolSize; i++) {
+      Connection conn = DriverManager.getConnection(ConnectionStringHelper.getWrapperClusterEndpointUrl(), props);
```
congoamz, line 680: Cannot use try-with-resources for the setup connections because then there would only be one connection in the pool (the same connection would be returned each time getConnection is called). Doing it this way instead allows the pool to be filled before closing the connections

wrapper/src/test/java/integration/refactored/container/tests/AutoscalingTests.java
```diff
@@ -109,6 +109,9 @@ protected HikariPoolConfigurator getHikariConfig(int maxPoolSize) {
   @TestTemplate
   public void test_pooledConnectionAutoScaling_setReadOnlyOnOldConnection()
       throws SQLException, InterruptedException {
+    // Timeout controlling whether a cached connection should be checked for validity before being returned.
+    // Replicate worst-case scenario: hikari does not check for validity before returning cached connections.
+    System.setProperty("com.zaxxer.hikari.aliveBypassWindowMs", String.valueOf(TimeUnit.HOURS.toMillis(1)));
```
karenc-bq, line 114: Why is this timeout 1 hour compared to the 10 minutes in the r/w splitting tests?

wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
```diff
@@ -645,30 +658,63 @@ public void test_pooledConnectionFailover() throws SQLException, InterruptedExce
...
+        assertThrows(FailoverSuccessSQLException.class, () -> auroraUtil.queryInstanceId(conn));
+
+        nextWriterId = auroraUtil.queryInstanceId(conn);
+        assertNotEquals(originalWriterId, nextWriterId);
+        assertEquals(0, provider.getHostCount(), "Internal connection pool should be empty after failover.");
+        newWriterConn = conn.unwrap(Connection.class);
+        assertNotSame(initialWriterConn, newWriterConn);
+      }
+
+      List<Connection> verificationConnections = new ArrayList<>();
```
karenc-bq, line 706:
```suggestion
      final List<Connection> verificationConnections = new ArrayList<>();
```

wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
```diff
@@ -645,30 +658,63 @@ public void test_pooledConnectionFailover() throws SQLException, InterruptedExce
...
+    final String nextWriterId;
+
+    List<Connection> setupConnections = new ArrayList<>();
+    for (int i = 0; i < maxPoolSize; i++) {
+      Connection conn = DriverManager.getConnection(ConnectionStringHelper.getWrapperClusterEndpointUrl(), props);
+      setupConnections.add(conn);
+    }
+    for (Connection conn : setupConnections) {
+      conn.close();
+    }
```
karenc-bq, line 685: What are you testing for here when you populate these connections then closing them immediately?

wrapper/src/test/java/integration/refactored/container/tests/AutoscalingTests.java
congoamz: We want to set this setting to some value greater than the amount of time the tests take so that we can replicate that worst-case scenario. The r/w splitting tests take <5 minutes each usually, but the autoscaling tests take ~20 minutes each

wrapper/src/test/java/integration/refactored/container/tests/ReadWriteSplittingTests.java
congoamz: They are not actually closed since it is a pooled connection, calling close actually returns them to the pool so that they are cached for later use. I'm filling the pool here so that I can verify that none of the connections added in this block are returned after failover.

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
sergiyvamz: If you move it to failover plugin, I still can't understand how to gently release certain connection pools. There's no designed communication between these two component. Calling ConnectionProviderManager.releaseResources() will effectively close all connection pools including those that contains readers connections that haven't been affected by failover.

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
sergiyvamz: I can assume that in case of reader cluster endpoint such connection pool will contain connections to different nodes. What should we do with such connections in case of failover?

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
congoamz: Let's discuss how/where to call releaseResources. My testing shows that all connections that existed before failover are broken after failover (regardless of any instance URL or cluster URL used) so I think all connection pools will need to be closed.

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
congoamz: My testing shows that all connections that existed before failover are broken after failover (regardless of any instance URL or cluster URL used) so I think all connection pools will need to be closed.

congoamz:
Closing due to concerns over using a pool for cluster URLs, particularly because they can resolve to different instances over time

PR #476, opened 2023-06-03
feat: ELB URL support

crystall-bitquill:
### Summary

Elastic Load Balancer URL support

### Description

Added an ELB pattern to match URL formatting and to obtain the AWS region from the URL.


wrapper/src/main/java/software/amazon/jdbc/util/RdsUrlType.java
```diff
@@ -23,6 +23,7 @@ public enum RdsUrlType {
   RDS_CUSTOM_CLUSTER(true, true),
   RDS_PROXY(true, false),
   RDS_INSTANCE(true, false),
+  ELB(false, false),
```
karenc-bq, line 26: Do we actually need an ELB type? Can it just go into the "OTHER" category

PR #514, opened 2023-07-04
feat: setLoginTimeout implementation

karenc-bq:
### Summary

Implement setLoginTimeout as suggested in #509 

### Additional Reviewers

@jrobe


wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -284,12 +295,19 @@ public void setLogWriter(final PrintWriter out) throws SQLException {
 
   @Override
   public void setLoginTimeout(final int seconds) throws SQLException {
-    throw new SQLFeatureNotSupportedException();
+    if (seconds <= 0) {
```
sergiyvamz, line 298: I'd validate user input here.

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -284,12 +295,19 @@ public void setLogWriter(final PrintWriter out) throws SQLException {
...
+      return;
+    }
+    loginTimeout = seconds;
   }
 
   @Override
   public int getLoginTimeout() throws SQLException {
-    throw new SQLFeatureNotSupportedException();
+    if (loginTimeout <= 0) {
```
sergiyvamz, line 306: Probably we shouldn't raise an exception in the getter method.

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -153,6 +154,16 @@ public Connection getConnection(final String username, final String password) th
     if (!StringUtils.isNullOrEmpty(this.targetDataSourceClassName)) {
 
       final DataSource targetDataSource = createTargetDataSource();
+      try {
```
sergiyvamz, line 157: I'm not sure about this try-catch block. What's the purpose? Does setLogintimeout() require a special handling?

karenc-bq: this calls the setLoginTimeout method in the target datasource, which may potentially throw a SQLException due to database access errors

PR #542, opened 2023-07-12
sort plugins by default

sergiyvamz:
### Summary

Sort plugins by default to prevent plugin misconfiguration. Allows a user to provide a custom plugin order if needed.

### Description

Plugin chain initialization has been separated to `ConnectionPluginChainBuilder `class. Depending on user settings, a plugin factory list is sorted based on plugin assigned weights. 

### Additional Reviewers

@karenc-bq @aaron-congo @aaronchung-bitquill @crystall-bitquill 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
@@ -1,0 +1,229 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import software.amazon.jdbc.plugin.AuroraConnectionTrackerPluginFactory;
+import software.amazon.jdbc.plugin.AuroraHostListConnectionPluginFactory;
+import software.amazon.jdbc.plugin.AwsSecretsManagerConnectionPluginFactory;
+import software.amazon.jdbc.plugin.ConnectTimeConnectionPluginFactory;
+import software.amazon.jdbc.plugin.DataCacheConnectionPluginFactory;
+import software.amazon.jdbc.plugin.DefaultConnectionPlugin;
+import software.amazon.jdbc.plugin.DriverMetaDataConnectionPluginFactory;
+import software.amazon.jdbc.plugin.ExecutionTimeConnectionPluginFactory;
+import software.amazon.jdbc.plugin.IamAuthConnectionPluginFactory;
+import software.amazon.jdbc.plugin.LogQueryConnectionPluginFactory;
+import software.amazon.jdbc.plugin.dev.DeveloperConnectionPluginFactory;
+import software.amazon.jdbc.plugin.efm.HostMonitoringConnectionPluginFactory;
+import software.amazon.jdbc.plugin.failover.FailoverConnectionPluginFactory;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPluginFactory;
+import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsPluginFactory;
+import software.amazon.jdbc.profile.DriverConfigurationProfiles;
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.SqlState;
+import software.amazon.jdbc.util.StringUtils;
+import software.amazon.jdbc.util.WrapperUtils;
+
+public class ConnectionPluginChainBuilder {
+
+  private static final Logger LOGGER = Logger.getLogger(ConnectionPluginChainBuilder.class.getName());
+
+  private static final int STICK_TO_PRIOR_PLUGIN = -1;
+
+  protected static final Map<String, Class<? extends ConnectionPluginFactory>> pluginFactoriesByCode =
+      new HashMap<String, Class<? extends ConnectionPluginFactory>>() {
+        {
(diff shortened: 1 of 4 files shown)


wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+        pluginFactories = this.sortPluginFactories(pluginFactories);
+
+        final List<Class<? extends ConnectionPluginFactory>> tempPluginFactories = pluginFactories;
+        LOGGER.finest(() ->
+            "Plugins order has been rearranged. The following order is in effect: "
+                + tempPluginFactories.stream()
+                  .map(Class::getSimpleName)
+                  .collect(Collectors.joining(", ")));
+      }
+
```
brunos-bq, line 167: is it worth to separate the code below into a function loadFactories() that returns the factories list?

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    // Determine weights for plugin factories with STICK_TO_PRIOR_PLUGIN
+    int lastWeight = 0;
+    for (PluginFactoryInfo info : weights) {
+      if (info.weight != STICK_TO_PRIOR_PLUGIN) {
+        lastWeight = info.weight;
+        continue;
+      }
+      lastWeight++;
```
brunos-bq, line 220: this bit is a bit confusing - why do we need to increment and reassign weights?

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -48,6 +48,12 @@ public class PropertyDefinition {
       new AwsWrapperProperty(
           "wrapperPlugins", null, "Comma separated list of connection plugin codes");
 
+  public static final AwsWrapperProperty PRESERVE_PLUGINS_ORDER =
```
brunos-bq, line 51: This is an interesting one.
How do we put this plugin order thing in the documentation?
If we want the sorting to be on by default, why not revert the wording from PRESERVE_PLUGINS_ORDER to AUTOMATICALLY_SORT_PLUGINS/AUTO_SORT_PLUGIN_ORDER ?

In this way, we could add something in the docs like this:
"This flag is enabled by default, meaning that the plugins order will be auto generated according to the order that we expect the plugins to be ordered. Disable it at your own risk or if you really need some plugins to be executed in a particular order"

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+        final ConnectionPluginFactory[] factories =
+            WrapperUtils.loadClasses(
+                    pluginFactories,
+                    ConnectionPluginFactory.class,
+                    "ConnectionPluginManager.unableToLoadPlugin")
+                .toArray(new ConnectionPluginFactory[0]);
+
+        // make a chain of connection plugins
+
+        plugins = new ArrayList<>(factories.length + 1);
```
karenc-bq, line 178: What is the value of initialize the plugins arraylist here with a fixed size? Can we just initialize it on line 118 
`final List<ConnectionPlugin> plugins = new ArrayList<>();` and remove the if-else clause.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+        new DefaultConnectionPlugin(pluginService, defaultConnProvider, pluginManagerService);
+    plugins.add(defaultPlugin);
+
+    return plugins;
+  }
+
+  protected List<Class<? extends ConnectionPluginFactory>> sortPluginFactories(
+      final List<Class<? extends ConnectionPluginFactory>> unsortedPluginFactories) {
+
+    ArrayList<PluginFactoryInfo> weights = new ArrayList<>();
```
karenc-bq, line 202:
```suggestion
    final List<PluginFactoryInfo> weights = new ArrayList<>();
```

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+          put(AuroraHostListConnectionPluginFactory.class, 300);
+          put(AuroraConnectionTrackerPluginFactory.class, 400);
+          put(AuroraStaleDnsPluginFactory.class, 500);
+          put(ReadWriteSplittingPluginFactory.class, 600);
+          put(FailoverConnectionPluginFactory.class, 700);
+          put(HostMonitoringConnectionPluginFactory.class, 800);
+          put(IamAuthConnectionPluginFactory.class, 900);
+          put(AwsSecretsManagerConnectionPluginFactory.class, 1000);
+          put(LogQueryConnectionPluginFactory.class, 1100);
+          put(ConnectTimeConnectionPluginFactory.class, STICK_TO_PRIOR_PLUGIN);
```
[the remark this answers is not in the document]
sergiyvamz: I need a weight as a number. In this case it's easier to sort elements in the list. 000 is not good for numbers. I use a constant value -1 to distinguish elements with absolute weights and elements with relative weight. The code below support such notation.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
sergiyvamz: Reworked this piece. Replaced two loops with a single one. Added more comments to explain the algorithm.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
sergiyvamz: Using `new ArrayList()` will initialize the list with 0 capacity. Every time we need to add a new element to the list will require the list to expand. Since `ArrayList` is based on internal array it means that an internal copy from one array to a new array will occur. By specifying an initial size we can avoid unnecessary overhead with copying.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -0,0 +1,233 @@
...
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.SqlState;
+import software.amazon.jdbc.util.StringUtils;
+import software.amazon.jdbc.util.WrapperUtils;
+
+public class ConnectionPluginChainBuilder {
+
+  private static final Logger LOGGER = Logger.getLogger(ConnectionPluginChainBuilder.class.getName());
+
+  private static final int WEIGHT_RELATIVE_TO_PRIOR_PLUGIN = -1;
```
brunos-bq, line 53: NIT: i would just add a comment that plugins that have this value in the map will not be affected by any reordering

brunos-bq (approved):

karenc-bq (approved):

PR #548, opened 2023-07-14
Close reader connection after switching to read/write mode

firebike:
### Summary

A fix for https://github.com/awslabs/aws-advanced-jdbc-wrapper/issues/547

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
@@ -363,6 +363,7 @@ public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
     } else {
       switchCurrentConnectionTo(this.writerConnection, writerHost);
     }
+    this.closeConnectionIfIdle(this.readerConnection);
 
     LOGGER.finer(() -> Messages.get("ReadWriteSplittingPlugin.switchedFromReaderToWriter",
         new Object[] {writerHost.getUrl()}));
--- a/wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
+++ b/wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
@@ -21,11 +21,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.mockito.AdditionalMatchers.not;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
 import java.sql.Connection;
 import java.sql.ResultSet;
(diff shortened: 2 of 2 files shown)


crystall-bitquill:
Hi @firebike,

Thanks for opening this PR!

I would like to suggest an addition to your changes, which is to close connections depending on whether internal connection pooling is used. This is because when internal connection pooling is not used, it can become expensive to close and reopen connections, especially if the `setReadOnly` method is called frequently. When using internal connection pooling, however, it makes sense to return connections to the pool as you mentioned in the issue.

I would suggest adding two flags to the `ReadWriteSplittingPlugin`: `isReaderConnFromInternalPool` and `isWriterConnFromInternalPool`, which would be used to check whether to close connections after switching to read-write or read-only.

When new writer or reader connections are made in the `getNewWriterConnection` or `getNewReaderConnection` methods, the `isReaderConnFromInternalPool` and `isWriterConnFromInternalPool` flags can be set by checking whether the connection provider is an instance of a `PooledConnectionProvider`. That would look something like this:
```java
  private void getNewWriterConnection(final HostSpec writerHostSpec) throws SQLException {
    final Connection conn = this.pluginService.connect(writerHostSpec, this.properties);
    this.isWriterConnFromInternalPool = this.connProviderManager.getConnectionProvider(
        this.pluginService.getDriverProtocol(),
        writerHostSpec,
        this.properties)
        instanceof PooledConnectionProvider;
    setWriterConnection(conn, writerHostSpec);
    switchCurrentConnectionTo(this.writerConnection, writerHostSpec);
  }
```
The `getNewReaderConnection` method could be handled similarly.

When switching connections in the `switchToWriterConnection` or `switchToReaderConnection` methods, instead of automatically closing the connections that are now idle, a check can be performed. If `isReaderConnFromInternalPool` or `isWriterConnFromInternalPool` is true, the now idle reader or writer connection can be returned to the connection pool.

We would also need an instance of a `ConnectionProviderManager`, which could be initialized in the constructor:
```java
  ReadWriteSplittingPlugin(final PluginService pluginService, final Properties properties) {
    this.pluginService = pluginService;
    this.properties = properties;
    this.readerSelectorStrategy = READER_HOST_SELECTOR_STRATEGY.getString(properties);
    this.connProviderManager = new ConnectionProviderManager(pluginService.getConnectionProvider());
  }
```

Since the connection provider and driver protocol are needed from the plugin service, they can be retrieved by extending the `PluginService` interface to include two methods: `getConnectionProvider` and `getDriverProtocol`. The methods could be implemented in `PluginServiceImpl` like this:
```java
  @Override
  public ConnectionProvider getConnectionProvider() {
    return this.pluginManager.defaultConnProvider;
  }

  @Override
  public String getDriverProtocol() {
    return this.driverProtocol;
  }
```

Please note that after altering the `PluginService` interface, the added methods would need to be implemented in `ConcurrencyTests.java`, in the `TestPluginService` class. They can return empty values.

Does this addition sound reasonable to you?

firebike:
Hi @crystall-bitquill 

Thank you for your detailed answer.

Before jumping on the implementation, I'd like to better understand the mechanics of the driver.

### Connection cleanup

There is a thing that bothers me a bit. 
Let's assume, that we have the following setup in place:
1. We are using the current implementation, without my fix.
2. We are using HikariCP in our application, means `aws-advanced-jdbc-wrapper` connections are pooled by the application.
3. We are switching connections to read-only mode, so effectively reader and writer connections are paired
4. We are **not** using the internal reader pool.

What happens when the application code returns `aws-advanced-jdbc-wrapper` connections back to the app connection pool? Especially, are the method calls that the app level HikariCP makes on a returned connection, propagated to the underlying writer and reader connections? If that would be the case, then we'd never see the OOM in our app, as all warnings should have been cleared (Normally HikariCP clears warnings on returned connection). Could it be, that the root cause of the problem is laying in this part of the driver code?

### Writer connections pooling

You suggest to implement `isWriterConnFromInternalPool`. Is it possible to use internal pools for writer connections too?

firebike:
Hi @crystall-bitquill 

Basing on our discussion, I've decided to:
1. Open a new PR https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/578 that implements `clearWarnings()` calls propagation to the underlying reader connection.
2. Close this PR, as closing the reader connection is not the proper solution for the memory leak problem caused by warnings accumulation. You may consider adding this functionality in the future to improve resource management. Imo, it will be useful (if not a must) in a topology with many reader nodes and a single writer node, in which the writer connections will be scarce.

PR #578, opened 2023-07-28
Propagate Connection.clearWarnings() to writer and reader connection

firebike:
### Summary

Propagate `Connection.clearWarnings()` to writer and reader connection

### Description

`ReadWriteSplittingPlugin` does not propagate `Connection.clearWarnings()` calls to the underlying reader connection, hence the connection object may accumulate warning messages, and in effect cause a memory leak in the application.

This PR fixes the memory leak problem described in https://github.com/awslabs/aws-advanced-jdbc-wrapper/issues/547

### DIFF
--- a/wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
+++ b/wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
@@ -23,6 +23,7 @@ import static org.mockito.AdditionalMatchers.not;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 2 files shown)


wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
```diff
@@ -473,4 +474,68 @@ public void testConnect_errorUpdatingHostSpec() throws SQLException {
...
+            "Connection.clearWarnings",
+            mockSqlFunction,
+            new Object[] {}
+    );
+    verify(mockWriterConn, never()).clearWarnings();
+    verify(mockReaderConn1, never()).clearWarnings();
+  }
+
+  @Test
+  public void testExecuteClearWarningsOnNullConnectionsIsNotCalled() throws SQLException {
```
crystall-bitquill, line 524: Does this test verify whether `clearWarnings` is or is not called?

crystall-bitquill (approved):

wrapper/src/test/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPluginTest.java
firebike: Good point. Assertion added.

sergiyvamz (approved):

PR #589, opened 2023-08-02
Feature: add dialect and host list provider for RDS MultiAZ DB Cluster

fangli:
### Summary

This PR:

1. add `RdsMysqlClusterDialect`
2. add `RdsMultiAzDbClusterListProvider`
3. refactor `AuroraHostListProvider`
4. test fixes


wrapper/src/main/java/software/amazon/jdbc/dialect/DialectCodes.java
```diff
@@ -20,9 +20,11 @@ public class DialectCodes {
   public static final String AURORA_MYSQL = "aurora-mysql";
   public static final String RDS_MYSQL = "rds-mysql";
   public static final String MYSQL = "mysql";
+  public static final String RDS_MULTIAZ_MYSQL_CLUSTER = "rds-maz-mysql-cluster";
```
sergiyvamz, line 23: "rds-mysql-taz" for consistency with previous codes?

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMysqlClusterDialect.java
```diff
@@ -0,0 +1,70 @@
...
+package software.amazon.jdbc.dialect;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.List;
+import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
+
+public class RdsMysqlClusterDialect extends MysqlDialect {
```
sergiyvamz, line 26: It seems that this class is specific to TAZ clusters so I'd recommend to rename it to RdsTazMysqlClusterDialect

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsPgClusterDialect.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.dialect;
+
+/**
+ * Suitable for the following AWS PG configurations.
+ * - Regional Cluster
+ */
+public class RdsPgClusterDialect extends PgDialect {
```
sergiyvamz, line 23: What's the purpose of the class?

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectCodes.java
```diff
@@ -20,9 +20,11 @@ public class DialectCodes {
   public static final String AURORA_MYSQL = "aurora-mysql";
   public static final String RDS_MYSQL = "rds-mysql";
   public static final String MYSQL = "mysql";
+  public static final String RDS_MULTI_AZ_MYSQL_CLUSTER = "rds-multi-az-mysql-cluster";
```
sergiyvamz, line 23: Can we be a more consistent with previous names and remove "_cluster"?

fangli: `Cluster` is part of the product name and we can't remove it.
[MultiAZ DB Cluster](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html) consists of 3 nodes, while [MultiAZ DB](https://aws.amazon.com/rds/features/multi-az/) consists of 1 active node and 1 standby node.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+
+    try {
+      final Statement stmt = conn.createStatement();
+      String writerNodeId = processWriterNodeId(stmt.executeQuery(this.fetchWriterNodeQuery));
+      if (writerNodeId.isEmpty()) {
+        final ResultSet nodeIdResultSet = stmt.executeQuery(this.nodeIdQuery);
+        while (nodeIdResultSet.next()) {
+          writerNodeId = nodeIdResultSet.getString(1);
+        }
+      }
```
sergiyvamz, line 93: This block is unclear. If writer is empty, it fetches topology and consider the last node in this topology as a writer. Is that correct?

fangli: This part is still being discussed and subject to change. For current implementation:
* If `SHOW REPLICA STATUS` returns a Source node ID, it implies that the current node is read replica.
* If `SHOW REPLICA STATUS` returns nothing, it means that the current node is not replicating from other nodes, implies that the current node is writer.

Issue #593, opened 2023-08-09
Unable to use reader endpoint to connect


lily-es:
### Describe the bug

Attempting to create a data pool using the reader endpoint (`XXXXXXXX.cluster-ro-XXXXXX.eu-north-1.rds.amazonaws.com`) using Hikari fails. This does not happen when using the writer endpoint(`XXXXXXXX.cluster-XXXXXX.eu-north-1.rds.amazonaws.com`).

### Expected Behavior

Connection to succeed just like when using the writer endpoint.

### What plugins are used? What other connection properties were set?

auroraConnectionTracker,failover,efm,iam

### Current Behavior

This exception is thrown

    12:10:50.628 ERROR | com.zaxxer.hikari.pool.PoolBase - backend - Failed to execute isValid() for connection, configure connection test query (java.lang.RuntimeException: Unable to identify the given connection: 'xxxxxxxxx.cluster-ro-xxxxxxxxx.eu-north-1.rds.amazonaws.com', please ensure the correct host list provider is specified. The host list provider in use is: 'software.amazon.jdbc.hostlistprovider.AuroraHostListProvider@60297f36'.). traceId=
    12:10:50.637 ERROR | com.zaxxer.hikari.pool.HikariPool - backend - Error thrown while acquiring connection from data source traceId=
    java.sql.SQLException: java.lang.RuntimeException: Unable to identify the given connection: 'xxxxxxxxx.cluster-ro-xxxxxxxxx.eu-north-1.rds.amazonaws.com', please ensure the correct host list provider is specified. The host list provider in use is: 'software.amazon.jdbc.hostlistprovider.AuroraHostListProvider@60297f36'.
	    at jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(Unknown Source) ~[?:?]
	    at java.lang.reflect.Constructor.newInstanceWithCaller(Unknown Source) ~[?:?]
	    at java.lang.reflect.Constructor.newInstance(Unknown Source) ~[?:?]
	    at software.amazon.jdbc.util.WrapperUtils.createInstance(WrapperUtils.java:444) ~[aws-advanced-jdbc-wrapper-2.2.3.jar:?]
	    at software.amazon.jdbc.util.WrapperUtils.wrapExceptionIfNeeded(WrapperUtils.java:562) ~[aws-advanced-jdbc-wrapper-2.2.3.jar:?]
[...]

PR #595, opened 2023-08-11
fix: update aurora pg topology query to also return instances with null last_update_timestamp

crystall-bitquill:
### Summary

Update aurora pg topology query to also return instances with null last_update_timestamp

### Description

In Aurora PostgreSQL versions 13.9 and above, the instance a user is currently connected to will not include a `last_update_timestamp` when retrieved from `aurora_replica_status()`.  This PR adjusts the topology query to account for the change. 

For example, querying the `last_update_timestamp` for all instance from `aurora_replica_status()` while connecting through the reader cluster url will result in this:
```
          server_id          | case |  last_update_timestamp  
-----------------------------+------+------------------------
 database-instance-1         | t    | 2023-08-15 00:29:11+00
 database-instance-2         | f    | 

```

This PR also bumps the integration test engine version to 13.9

Addresses #593.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
@@ -42,7 +42,8 @@ public class AuroraPgDialect extends PgDialect {
           + "CPU, COALESCE(REPLICA_LAG_IN_MSEC, 0), LAST_UPDATE_TIMESTAMP "
           + "FROM aurora_replica_status() "
           // filter out nodes that haven't been updated in the last 5 minutes
-          + "WHERE EXTRACT(EPOCH FROM(NOW() - LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID' ";
+          + "WHERE EXTRACT(EPOCH FROM(NOW() - LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID' "
+          + "OR LAST_UPDATE_TIMESTAMP IS NULL";
 
   private static final String NODE_ID_QUERY = "SELECT aurora_db_instance_identifier()";
   private static final String IS_READER_QUERY = "SELECT pg_is_in_recovery()";
--- a/wrapper/src/test/java/integration/container/tests/AuroraConnectivityTests.java
+++ b/wrapper/src/test/java/integration/container/tests/AuroraConnectivityTests.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package integration.container.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import integration.DatabaseEngineDeployment;
+import integration.DriverHelper;
+import integration.TestEnvironmentFeatures;
+import integration.container.ConnectionStringHelper;
+import integration.container.TestDriver;
+import integration.container.TestDriverProvider;
+import integration.container.TestEnvironment;
+import integration.container.condition.DisableOnTestFeature;
+import integration.container.condition.EnableOnDatabaseEngineDeployment;
+import integration.container.condition.EnableOnNumOfInstances;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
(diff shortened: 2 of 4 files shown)


wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
```diff
@@ -65,7 +65,6 @@
 import software.amazon.jdbc.HikariPooledConnectionProvider;
 import software.amazon.jdbc.PropertyDefinition;
 import software.amazon.jdbc.hostlistprovider.AuroraHostListProvider;
-import software.amazon.jdbc.hostlistprovider.ConnectionStringHostListProvider;
```
brunos-bq, line 68: this PR doesn't remove any code - is it unused?

wrapper/src/test/java/integration/container/tests/AuroraConnectivityTests.java
```diff
@@ -0,0 +1,89 @@
...
+        PropertyDefinition.USER.name,
+        TestEnvironment.getCurrent().getInfo().getDatabaseInfo().getUsername());
+    props.setProperty(
+        PropertyDefinition.PASSWORD.name,
+        TestEnvironment.getCurrent().getInfo().getDatabaseInfo().getPassword());
+    DriverHelper.setConnectTimeout(testDriver, props, 10, TimeUnit.SECONDS);
+    DriverHelper.setSocketTimeout(testDriver, props, 10, TimeUnit.SECONDS);
+    props.setProperty(PropertyDefinition.PLUGINS.name, "efm");
+
+    String url =
```
brunos-bq, line 74: NIT: what's the setting for max length in a line? seems like a break shouldn't be needed here

wrapper/src/test/java/integration/container/tests/AuroraConnectivityTests.java
```diff
@@ -0,0 +1,89 @@
...
+    TestEnvironmentFeatures.PERFORMANCE,
+    TestEnvironmentFeatures.RUN_HIBERNATE_TESTS_ONLY,
+    TestEnvironmentFeatures.RUN_AUTOSCALING_TESTS_ONLY})
+public class AuroraConnectivityTests {
+
+  private static final Logger LOGGER = Logger.getLogger(BasicConnectivityTests.class.getName());
+
+  @TestTemplate
+  @ExtendWith(TestDriverProvider.class)
+  public void test_WrapperConnectionReaderClusterWithEfmEnabled(TestDriver testDriver) throws SQLException {
```
brunos-bq, line 60: curious about this test - if all we do is connecting and fetching a result? isn't this covered in many other tests?

not exactly understanding what we are testing here that is new

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
```diff
@@ -42,7 +42,8 @@ public class AuroraPgDialect extends PgDialect {
           + "CPU, COALESCE(REPLICA_LAG_IN_MSEC, 0), LAST_UPDATE_TIMESTAMP "
           + "FROM aurora_replica_status() "
           // filter out nodes that haven't been updated in the last 5 minutes
-          + "WHERE EXTRACT(EPOCH FROM(NOW() - LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID' ";
+          + "WHERE EXTRACT(EPOCH FROM(NOW() - LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID' "
+          + "OR LAST_UPDATE_TIMESTAMP IS NULL";
```
brunos-bq, line 46: can LAST_UPDATE_TIMESTAMP be NULL? should it be OK for it to be null?

what is the value of the field when a node come in?

wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
crystall-bitquill: Yes it's an unused import

wrapper/src/test/java/integration/container/tests/AuroraConnectivityTests.java
crystall-bitquill: This particular case (connecting with the reader cluster url, with efm enabled) hasn't been covered. There's one test in the ReadWriteSplittingTests that uses the reader cluster url but does not enabled efm. Without the updated topology query this connection would fail.

brunos-bq: that's interesting. nice catch. do you think this test could not go somewhere else?
i am just wondering that the name "AuroraConnectivityTests" might be misleading

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
brunos-bq: i was wondering the case of a cluster scaling up , to see what would happen if a node comes in and there is no LAST_UPDATE_TIMESTAMP.

overall, is it safe to consider that every time a node has this field NULL we should consider it?

crystall-bitquill: New nodes seem to have a last_update_timestamp from creation, and when nodes are deleted the entire record is removed. Even if a new node has a brief period of time where the last_update_timestamp is null, it should still be returned as part of the topology query results. It should be safe to consider all nodes with null last_update_timestamp values.

aaronchung-bitquill:
LGTM!

aaronchung-bitquill (approved):
LGTM!

karenc-bq (approved):

PR #601, opened 2023-08-14
fix: expand ReentrantLock scope in MonitorThreadContainer for better …

aaronchung-bitquill:
…thread safety

### Summary
Remove race condition in `MonitorThreadContainer` between `getInstance()` and `releaseInstance()` methods.

### Description

In the `MonitorThreadContainer`, there are possible race conditions between the `getInstance()` and `releaseInstance()` methods. 

In one of the race conditions is in the `getInstance()`. 
```
...
      } finally {
        LOCK_OBJECT.unlock();
      }
    }
    CLASS_USAGE_COUNT.getAndIncrement();
    return singleton;
  }
```
Just after the lock object is unlocked, is an opportunity for a race condition with another thread in the `releaseInstance()` method. 
```
...
    if (CLASS_USAGE_COUNT.decrementAndGet() <= 0) {
      LOCK_OBJECT.lock();
      try {
        if (singleton != null) {
          singleton.releaseResources();
          singleton = null;
          CLASS_USAGE_COUNT.set(0);
...
```
If the locked block is entered, a race condition may result.  `getInstance()` may  return a null `singleton` which will be followed by a NPE. Or another scenario is that a non-null singleton will be returned, but `singleton.releaseResources()` is called so the `MonitorThreadContainer`'s `threadPool` will be shutdown and subsequent usage of the threadPool will result in a `RejectedExecutionException`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### :magic_wand: Added
 - Host Availability Strategy to help keep host health status up to date ([PR #530](https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/530)).
 
+### :bug: Fixed
+- Race condition issues between `MonitorThreadContainer#getInstance()` and `MonitorThreadContainer#releaseInstance()` ([PR #601](https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/601)).
+
 ### :crab: Changed
 - Dynamically sets the default host list provider based on the dialect used. User applications no longer need to manually set the AuroraHostListProvider when connecting to Aurora Postgres or Aurora MySQL databases.
 - Deprecated AuroraHostListConnectionPlugin. 
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorThreadContainer.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorThreadContainer.java
@@ -54,19 +54,19 @@ public class MonitorThreadContainer {
   }
 
   static MonitorThreadContainer getInstance(final ExecutorServiceInitializer executorServiceInitializer) {
-    if (singleton == null) {
-      LOCK_OBJECT.lock();
-      try {
-        if (singleton == null) {
-          singleton = new MonitorThreadContainer(executorServiceInitializer);
-          CLASS_USAGE_COUNT.set(0);
-        }
-      } finally {
-        LOCK_OBJECT.unlock();
+    MonitorThreadContainer singletonToReturn;
+    LOCK_OBJECT.lock();
+    try {
+      if (singleton == null) {
+        singleton = new MonitorThreadContainer(executorServiceInitializer);
+        CLASS_USAGE_COUNT.set(0);
       }
+      singletonToReturn = singleton;
+      CLASS_USAGE_COUNT.getAndIncrement();
+    } finally {
+      LOCK_OBJECT.unlock();
     }
-    CLASS_USAGE_COUNT.getAndIncrement();
-    return singleton;
+    return singletonToReturn;
   }
 
   /**
@@ -77,18 +77,15 @@ public class MonitorThreadContainer {
     if (singleton == null) {
       return;
     }
-
-    if (CLASS_USAGE_COUNT.decrementAndGet() <= 0) {
-      LOCK_OBJECT.lock();
-      try {
-        if (singleton != null) {
-          singleton.releaseResources();
-          singleton = null;
-          CLASS_USAGE_COUNT.set(0);
-        }
-      } finally {
-        LOCK_OBJECT.unlock();
+    LOCK_OBJECT.lock();
+    try {
+      if (singleton != null && CLASS_USAGE_COUNT.decrementAndGet() <= 0) {
+        singleton.releaseResources();
+        singleton = null;
+        CLASS_USAGE_COUNT.set(0);
       }
+    } finally {
+      LOCK_OBJECT.unlock();
     }
   }
 
--- a/wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
+++ b/wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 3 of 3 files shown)


aaronchung-bitquill:
Just realized that perhaps using `synchronized` on both methods may be a better solution.

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
```diff
@@ -0,0 +1,68 @@
+/*
```
brunos-bq, line 1: NIT: can we rename it to ConcurrencyEFMTest ?

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### :magic_wand: Added
 - Host Availability Strategy to help keep host health status up to date ([PR #530](https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/530)).
 
+### :bug: Fixed
+- Race conditions between `MonitorThreadContainer#getInstance()` and `MonitorThreadContainer#releaseInstance()` ([PR #601](https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/601)).
```
brunos-bq, line 11: NIT: Fixed a case of race condition between...

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
```diff
@@ -0,0 +1,68 @@
...
+    Runnable getInstanceTask = () -> {
+      MonitorThreadContainer.getInstance(mockExecutorServiceInitializer);
+    };
+
+    Runnable releaseInstanceTask = MonitorThreadContainer::releaseInstance;
+
+    MonitorThreadContainer.getInstance(mockExecutorServiceInitializer);
+
+    ExecutorService executorService = Executors.newFixedThreadPool(2);
+    executorService.execute(getInstanceTask);
```
brunos-bq, line 63: should we execute getInstance and releaseInstance more times here?

davecramer:
Instead of using synchronized it would be better to use the re-entrant lock before the first check. synchronized is still a heavier lock.

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
aaronchung-bitquill: `ConcurrencyEFMTest` may be too general. There is also another test class `MultiThreadedDefaultMonitorServiceTest` that also tests concurrency in EFM.

aaronchung-bitquill: Is the nit b/c `MultiThreadedMonitorThreadContainerTest` is extremely verbose? :D 

I was thinking of replacing `MultiThreaded` with `Concurrency` so the test class name would be `ConcurrencyMonitorTheadContainerTest` but that still seems a bit wordy. 

What do you think?

CHANGELOG.md
aaronchung-bitquill: None of the Fixed changelog items start with "Fixed ...." and instead either state the bug, or what was done to fix the bug. I did this once and was called out and IIRC the reason was that "Fixed..."  is redundant with the header.

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
aaronchung-bitquill: Actually, since this test a `@RepeatedTest`, I think this should suffice. Is that ok with you?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorThreadContainer.java
```diff
@@ -54,41 +54,35 @@ public static MonitorThreadContainer getInstance() {
...
-      LOCK_OBJECT.lock();
-      try {
-        if (singleton != null) {
-          singleton.releaseResources();
-          singleton = null;
-          CLASS_USAGE_COUNT.set(0);
-        }
-      } finally {
-        LOCK_OBJECT.unlock();
+    LOCK_OBJECT.lock();
```
brunos-bq, line 77: just double checking - do we enter the lock section even if singleton is null when method is called?

aaronchung-bitquill: Yes. I originally thought there could b a race condition in the current code where after the null check, and before it returns, another thread could call getInstance, which would make the singleton not null, and the original thread thats calling releaseInstance wouldnt know about that.
But now im wondering if thats an issue at all.

CHANGELOG.md
aaronchung-bitquill: Done!

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorThreadContainer.java
aaronchung-bitquill: I will add the null check and early return back. 
This isnt the area that we've seen issues with, so I think the safer approach here is to leave it as it was.

crystall-bitquill (approved):

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.RepeatedTest;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+public class MultiThreadedMonitorThreadContainerTest {
+
+  @Mock
+  ExecutorServiceInitializer mockExecutorServiceInitializer;
```
karenc-bq, line 34:
```suggestion
  @Mock ExecutorServiceInitializer mockExecutorServiceInitializer;
```

wrapper/src/test/java/software/amazon/jdbc/plugin/efm/MultiThreadedMonitorThreadContainerTest.java
```diff
@@ -0,0 +1,64 @@
...
+  @Mock
+  ExecutorService mockExecutorService;
```
karenc-bq, line 36:
```suggestion
  @Mock ExecutorService mockExecutorService;
```

karenc-bq (approved):

PR #632, opened 2023-09-07
feat: optional auto reconfigure connection post failover

crystall-bitquill:
### Summary

Optional auto reconfigure connection post failover

### Description

This PR adds the `keepSessionStateOnFailover` property to the failover plugin. It allows a partial session state to be recovered after a failover. This includes the autoCommit and ReadOnly values.

This is necessary in cases where manually reconfiguring a connection is not possible. One example case is with Spring Boot and HikariCP. When autoCommit is set to false, if failover occurs and the autoCommit value is reset to true, there will be errors due to the unexpected autoCommit value. Connections will not be usable in that case.


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -70,6 +70,8 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
         }
       });
 
+  private static final String METHOD_SET_READ_ONLY = "Connection.setReadOnly";
+  private static final String METHOD_SET_AUTO_COMMIT = "Connection.SetAutoCommit";
```
sergiyvamz, line 74: Connection.setAutoCommit  (small "s")

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -727,6 +771,12 @@ private Connection connectInternal(String driverProtocol, HostSpec hostSpec, Pro
         this.staleDnsHelper.getVerifiedConnection(isInitialConnection, this.hostListProviderService,
             driverProtocol, hostSpec, props, connectFunc);
 
+    if (this.keepSessionStateOnFailover) {
+      this.savedReadOnlyStatus = this.savedReadOnlyStatus == null ? conn.isReadOnly() : this.savedReadOnlyStatus;
```
sergiyvamz, line 775: Do we update these stored values in case you call setReadOnly() or setAutoCommit() after connection is open?

sergiyvamz (approved):

karenc-bq (approved):

PR #659, opened 2023-09-25
chore: update performance test to use failover mode

karenc-bq:
### Summary

Performance tests were using the "readonly" flag to determine whether the driver should reconnect to the writer or connect to a reader during failover.
This approach is outdated after the failoverMode connection parameter is introduced, so tests will wait for TASK A to fail instead of finishing failover as soon as a reader connection is established.


crystall-bitquill (approved):

wrapper/src/test/java/integration/container/tests/PerformanceTest.java
```diff
@@ -279,9 +281,10 @@ private void execute_FailoverTime_SocketTimeout(int socketTimeout, int sleepDela
                 .getProxyDatabaseInfo()
                 .getInstanceEndpointSuffix());
     props.setProperty("failoverTimeoutMs", Integer.toString(PERF_FAILOVER_TIMEOUT_MS));
+    FailoverConnectionPlugin.FAILOVER_MODE.set(props, "strict-reader");
```
brunos-bq, line 284: this is equivalent to setting that parameter to true?

wrapper/src/test/java/integration/container/tests/PerformanceTest.java
```diff
@@ -163,7 +164,7 @@ private void execute_FailureDetectionTime_EnhancedMonitoringEnabled(
     PLUGINS.set(props, "efm");
 
     final PerfStatMonitoring data = new PerfStatMonitoring();
-    doMeasurePerformance(sleepDelayMillis, REPEAT_TIMES, props, false, data);
+    doMeasurePerformance(sleepDelayMillis, REPEAT_TIMES, props, data);
```
brunos-bq, line 167: does that mean that what was false is now true?

sergiyvamz (approved):

wrapper/src/test/java/integration/container/tests/PerformanceTest.java
crystall-bitquill: No, the parameter was removed altogether from the doMeasurePerformance method and the failover mode does not default to being read onlt.

wrapper/src/test/java/integration/container/tests/PerformanceTest.java
crystall-bitquill: Yes

Issue #675, opened 2023-10-10
Memory Leak in EFM Plugin


chr-fritz:
### Describe the bug

In our application we've two aurora databases each with an own jdbc connection. Through our monitoring we found that our pods randomly restarted due to out of memory. After some investigation we found out that the `software.amazon.jdbc.plugin.efm.MonitorImpl` of one of our jdbc connections stopped working and we don't saw any exceptions. In this case it collects `software.amazon.jdbc.plugin.efm.MonitorConnectionContext` objects in the `newContexts` queue. In one of our heap dumps we found out that in the `newContexts` queue were over 4000000 objects.

This screenshot from one of our heap show that the first instance is `stopped=true` with over 411MB of objects in the `newContexts` queue. The second one is the one which is working it no objects in the `newContexts` queue.
<img width="1263" alt="grafik" src="https://github.com/awslabs/aws-advanced-jdbc-wrapper/assets/1180400/e0469a70-e559-4df4-b585-3e8473f2b17a">

### Expected Behavior

No memory leak.

### What plugins are used? What other connection properties were set?

auroraConnectionTracker,failover,efm,iam

### Current Behavior

Randomly the Monitor stopps and objects are stored in `newContexts` but they won't ever be removed.

### Reproduction Steps

We couldn't find any reproducer as we just observed this behavior randomly without any pattern.

### Possible Solution

At least catch and log `RuntimeException` and return an appropriate `ConnectionStatus` object in `software.amazon.jdbc.plugin.efm.MonitorImpl#checkConnectionStatus` to avoid that uncaught runtime exceptions stop the monitoring thread without shutting down the monitor completly.

### The AWS Advanced JDBC Driver version used

2.2.2

### JDK version used

zulu 11.0.17

### Operating System and version

Distroless Docker Image at AWS EKS

crystall-bitquill:
Hi @chr-fritz,

Thanks for reaching out and raising this issue.

We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

crystall-bitquill:
Hi @chr-fritz,

[...]

PR #676, opened 2023-10-10
fix: continue monitoring if unhandled Exception is thrown

crystall-bitquill:
### Summary

Add additional logging for efm

### Description

- Adds additional logging in the `MonitorImpl` class for previously unhandled exceptions and for scenarios where monitoring has stopped but the `startMonitoring` method was called again.
- Changes the monitoring loop to continue running unless an InterruptedException is thrown.

Related to #675 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
@@ -100,7 +100,7 @@ public class MonitorImpl implements Monitor {
     this.properties = properties;
     this.monitorDisposalTimeMillis = monitorDisposalTimeMillis;
     this.monitorService = monitorService;
-    
+
     this.contextLastUsedTimestampNano = this.getCurrentTimeNano();
     this.contextsSizeGauge = telemetryFactory.createGauge("efm.activeContexts.queue.size",
         () -> (long) activeContexts.size());
@@ -113,6 +113,9 @@ public class MonitorImpl implements Monitor {
 
   @Override
   public void startMonitoring(final MonitorConnectionContext context) {
+    if (this.stopped) {
+      LOGGER.warning(() -> Messages.get("MonitorImpl.monitorIsStopped", new Object[] {this.hostSpec.getHost()}));
+    }
     final long currentTimeNano = this.getCurrentTimeNano();
     context.setStartMonitorTimeNano(currentTimeNano);
     this.contextLastUsedTimestampNano = currentTimeNano;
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -220,14 +223,17 @@ public void run() {
...
+    } catch (final Exception ex) {
       // do nothing; exit thread
+      LOGGER.warning(() -> Messages.get("MonitorImpl.exceptionDuringMonitoring", new Object[] {ex.getMessage()}));
     } finally {
       if (this.monitoringConn != null) {
         try {
           this.monitoringConn.close();
         } catch (final SQLException ex) {
           // ignore
+          LOGGER.warning(
```
sergiyvamz, line 235: Hmm I'm not sure this message adds extra clarity.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -94,6 +94,9 @@ public MonitorImpl(
 
   @Override
   public void startMonitoring(final MonitorConnectionContext context) {
+    if (this.stopped) {
+      LOGGER.warning(() -> Messages.get("MonitorImpl.monitorIsStopped"));
```
sergiyvamz, line 98: Can you add monitoring host name?

davecramer:
Do we have an issue to refer to that prompted this change ?

crystall-bitquill:
> Do we have an issue to refer to that prompted this change ?

Yes, #675. I've updated the PR description.

wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
```diff
@@ -188,6 +188,9 @@ MonitorThreadContainer.emptyNodeKeys=Provided node keys are empty.
 
 # Monitor Impl
 MonitorImpl.contextNullWarning=Parameter 'context' should not be null.
+MonitorImpl.interruptedExceptionDuringMonitoring=Monitoring thread for node {0} was interrupted: {1}
+MonitorImpl.exceptionDuringMonitoring=Unhandled exception in monitoring thread for node {0}: {1}
+MonitorImpl.monitorIsStopped=Monitoring has already stopped for node {0}.
```
brunos-bq, line 193: NIT: Monitor was or is already stopped?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -244,6 +247,16 @@ public void run() {
       }
     } catch (final InterruptedException intEx) {
       // do nothing; exit thread
+      LOGGER.warning(
+          () -> Messages.get(
+              "MonitorImpl.interruptedExceptionDuringMonitoring",
+              new Object[] {this.hostSpec.getHost(), intEx.getMessage()}));
+    } catch (final Exception ex) {
```
brunos-bq, line 254: are we expecting any other kind of exception? (we did not catch it beforehand)

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -244,6 +247,16 @@ public void run() {
...
+      // do nothing; exit thread
+      LOGGER.warning(
```
sergiyvamz, line 256: I think we need to change code so such unhandled exceptions to be logged but the monitoring thread/loop keeps running.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -113,6 +113,9 @@ public MonitorImpl(
 
   @Override
   public void startMonitoring(final MonitorConnectionContext context) {
+    if (this.stopped) {
+      LOGGER.warning(() -> Messages.get("MonitorImpl.monitorIsStopped", new Object[] {this.hostSpec.getHost()}));
```
davecramer, line 117: This seems confusing, why are we logging that it is stopped. Presumably it would be stopped if we call startMonitoring?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -113,6 +113,9 @@ public MonitorImpl(
...
+    }
     final long currentTimeNano = this.getCurrentTimeNano();
     context.setStartMonitorTimeNano(currentTimeNano);
```
davecramer, line 120: what stops someone calling startMonitoring more than once ?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
crystall-bitquill: We're logging that the monitor is stopped in case the run method has finished and stopped is false but startMonitoring has been called again. This is because of a reported issue where the newContexts queue grew to the point of causing an out of memory error, which implies that items are being added to the queue through startMonitoring and not being removed, which would normally happen in the run method.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
crystall-bitquill: The startMonitoring method is called every time the execute method runs, so will be called more than once.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
```diff
@@ -137,125 +140,142 @@ public void clearContexts() {
...
-              }
+            while ((monitorContext = this.activeContexts.poll()) != null) {
 
-              if (firstAddedMonitorContext == monitorContext) {
-                // this context has already been processed by this loop
-                // add it to the queue and exit this loop
-                this.activeContexts.add(monitorContext);
-                break;
-              }
+              synchronized (monitorContext) {
```
davecramer, line 192: Does this really have to be synchronized? It is a local variable

sergiyvamz: It's a local variable but we add it to queues in MonitorImpl that runs in a separate thread. Monitoring thread may change an internal state of the context. The idea was to synchronize on it to avoid multi-threading collisions. It seems this intent isn't properly implemented here. Need to address it in a separate PR.

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorImpl.java
crystall-bitquill: After some investigation, it seems like the current implementation is working and a fix isn't required.

PR #683, opened 2023-10-12
fix: update topology for specific methods

crystall-bitquill:
### Summary

Update topology for specific methods

### Description

Specify a list of methods where it's safe to update the topology before executing to prevent error where connection objects have more than one streaming result set open at the same time.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
@@ -50,4 +50,31 @@ public class SubscribedMethodHelper {
       "CallableStatement.executeUpdate",
       "CallableStatement.executeWithFlags"
   );
+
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java
```diff
@@ -50,4 +51,32 @@ public class SubscribedMethodHelper {
       "CallableStatement.executeUpdate",
       "CallableStatement.executeWithFlags"
   );
+
+  public static final List<String> METHODS_REQUIRING_UPDATED_TOPOLOGY = Arrays.asList(
+      "Connection.commit",
+      "Connection.connect",
+      "Connection.isValid",
+      "Connection.rollback",
```
sergiyvamz, line 59: I'd remove rollback from the list.

PR #690, opened 2023-10-19
Add support for AWS RDS Multi-AZ DB Cluster, MySQL and Postgres.

fangli:
### Summary

This CR is to add support for AWS RDS Multi-AZ DB Cluster, both MySQL and Postgres.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
@@ -52,8 +52,7 @@ public class AuroraMysqlDialect extends MysqlDialect {
 
   @Override
   public List</* dialect code */ String> getDialectUpdateCandidates() {
-    return Collections.singletonList(
-        DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER);
+    return Collections.singletonList(DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER);
   }
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectCodes.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectCodes.java
@@ -20,10 +20,12 @@ public class DialectCodes {
   public static final String AURORA_MYSQL = "aurora-mysql";
   public static final String RDS_MYSQL = "rds-mysql";
   public static final String MYSQL = "mysql";
+  // https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html
   public static final String RDS_MULTI_AZ_MYSQL_CLUSTER = "rds-multi-az-mysql-cluster";
 
   public static final String AURORA_PG = "aurora-pg";
   public static final String RDS_PG = "rds-pg";
+  // https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html
   public static final String RDS_MULTI_AZ_PG_CLUSTER = "rds-multi-az-pg-cluster";
   public static final String PG = "pg";
 
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
@@ -166,6 +166,7 @@ public class DialectManager implements DialectProvider {
     if (driverProtocol.contains("postgresql")) {
       RdsUrlType type = this.rdsHelper.identifyRdsType(host);
       if (type.isRdsCluster()) {
+        this.canUpdate = true;
         this.dialectCode = DialectCodes.AURORA_PG;
         this.dialect = knownDialectsByCode.get(DialectCodes.AURORA_PG);
         return this.dialect;
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
@@ -27,9 +27,7 @@ import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.jdbc.AwsWrapperProperty;
 import software.amazon.jdbc.HostSpec;
-import software.amazon.jdbc.HostSpecBuilder;
 import software.amazon.jdbc.PluginService;
-import software.amazon.jdbc.hostlistprovider.AuroraHostListProvider;
 import software.amazon.jdbc.util.CacheMap;
 import software.amazon.jdbc.util.ConnectionUrlParser;
 import software.amazon.jdbc.util.Messages;
@@ -59,7 +57,9 @@ public class DialectManager implements DialectProvider {
           put(DialectCodes.PG, new PgDialect());
           put(DialectCodes.MARIADB, new MariaDbDialect());
           put(DialectCodes.RDS_MYSQL, new RdsMysqlDialect());
+          put(DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER, new RdsMultiAzDbClusterMysqlDialect());
           put(DialectCodes.RDS_PG, new RdsPgDialect());
+          put(DialectCodes.RDS_MULTI_AZ_PG_CLUSTER, new RdsMultiAzDbClusterPgDialect());
           put(DialectCodes.AURORA_MYSQL, new AuroraMysqlDialect());
           put(DialectCodes.AURORA_PG, new AuroraPgDialect());
           put(DialectCodes.UNKNOWN, new UnknownDialect());
@@ -144,6 +144,7 @@ public class DialectManager implements DialectProvider {
     if (driverProtocol.contains("mysql")) {
       RdsUrlType type = this.rdsHelper.identifyRdsType(host);
       if (type.isRdsCluster()) {
+        this.canUpdate = true;
         this.dialectCode = DialectCodes.AURORA_MYSQL;
         this.dialect = knownDialectsByCode.get(DialectCodes.AURORA_MYSQL);
         return this.dialect;
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
@@ -32,7 +32,9 @@ import software.amazon.jdbc.hostlistprovider.ConnectionStringHostListProvider;
 public class MysqlDialect implements Dialect {
 
   private static final List<String> dialectUpdateCandidates = Arrays.asList(
-      DialectCodes.AURORA_MYSQL, DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER, DialectCodes.RDS_MYSQL);
+      DialectCodes.AURORA_MYSQL,
+      DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER,
+      DialectCodes.RDS_MYSQL);
   private static MySQLExceptionHandler mySQLExceptionHandler;
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
@@ -32,7 +32,7 @@ import software.amazon.jdbc.hostlistprovider.ConnectionStringHostListProvider;
 public class MysqlDialect implements Dialect {
 
   private static final List<String> dialectUpdateCandidates = Arrays.asList(
-      DialectCodes.AURORA_MYSQL, DialectCodes.RDS_MYSQL);
+      DialectCodes.AURORA_MYSQL, DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER, DialectCodes.RDS_MYSQL);
   private static MySQLExceptionHandler mySQLExceptionHandler;
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
@@ -26,7 +26,7 @@ import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
 public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
 
   private static final String TOPOLOGY_QUERY =
-      "SELECT id, endpoint, port FROM rds.rds_topology";
+      "SELECT id, endpoint, port FROM mysql.rds_topology";
 
   private static final String FETCH_WRITER_NODE_QUERY =
       "SHOW REPLICA STATUS";
@@ -39,7 +39,7 @@ public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
   @Override
   public boolean isDialect(final Connection connection) {
     try (final Statement stmt = connection.createStatement();
-         final ResultSet rs = stmt.executeQuery("SELECT id FROM rds.rds_topology LIMIT 1")) {
+         final ResultSet rs = stmt.executeQuery("SELECT id FROM mysql.rds_topology LIMIT 1")) {
       if (rs.next()) {
         // If the table is presented it means it's a Multi-AZ DB cluster
         return true;
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
@@ -20,28 +20,28 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.Collections;
 import java.util.List;
-import software.amazon.jdbc.hostlistprovider.AuroraHostListProvider;
+import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
 
-public class AuroraMysqlDialect extends MysqlDialect {
+public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
 
   private static final String TOPOLOGY_QUERY =
-      "SELECT SERVER_ID, CASE WHEN SESSION_ID = 'MASTER_SESSION_ID' THEN TRUE ELSE FALSE END, "
-          + "CPU, REPLICA_LAG_IN_MILLISECONDS, LAST_UPDATE_TIMESTAMP "
-          + "FROM information_schema.replica_host_status "
-          // filter out nodes that haven't been updated in the last 5 minutes
-          + "WHERE time_to_sec(timediff(now(), LAST_UPDATE_TIMESTAMP)) <= 300 OR SESSION_ID = 'MASTER_SESSION_ID' ";
+      "SELECT id, endpoint, port FROM rds.rds_topology";
 
-  private static final String NODE_ID_QUERY = "SELECT @@aurora_server_id";
-  private static final String IS_READER_QUERY = "SELECT @@innodb_read_only";
+  private static final String FETCH_WRITER_NODE_QUERY =
+      "SHOW REPLICA STATUS";
+
+  private static final String FETCH_WRITER_NODE_QUERY_HEADER = "Source_Server_Id";
+
+  private static final String NODE_ID_QUERY = "SELECT @@server_id";
+  private static final String IS_READER_QUERY = "SELECT @@read_only";
 
   @Override
   public boolean isDialect(final Connection connection) {
     try (final Statement stmt = connection.createStatement();
-        final ResultSet rs = stmt.executeQuery("SHOW VARIABLES LIKE 'aurora_version'")) {
+         final ResultSet rs = stmt.executeQuery("SELECT id FROM rds.rds_topology LIMIT 1")) {
       if (rs.next()) {
-        // If variable with such name is presented then it means it's an Aurora cluster
+        // If the table is presented it means it's a Multi-AZ DB cluster
         return true;
       }
     } catch (final SQLException ex) {
@@ -52,18 +52,19 @@ public class AuroraMysqlDialect extends MysqlDialect {
 
   @Override
   public List</* dialect code */ String> getDialectUpdateCandidates() {
-    return Collections.singletonList(
-        DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER);
+    return null;
   }
 
   @Override
   public HostListProviderSupplier getHostListProvider() {
-    return (properties, initialUrl, hostListProviderService) -> new AuroraHostListProvider(
+    return (properties, initialUrl, hostListProviderService) -> new RdsMultiAzDbClusterListProvider(
         properties,
         initialUrl,
         hostListProviderService,
         TOPOLOGY_QUERY,
         NODE_ID_QUERY,
-        IS_READER_QUERY);
+        IS_READER_QUERY,
+        FETCH_WRITER_NODE_QUERY,
+        FETCH_WRITER_NODE_QUERY_HEADER);
   }
 }
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
@@ -25,11 +25,9 @@ import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
 
 public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
 
-  private static final String TOPOLOGY_QUERY =
-      "SELECT id, endpoint, port FROM mysql.rds_topology";
+  private static final String TOPOLOGY_QUERY = "SELECT id, endpoint, port FROM mysql.rds_topology";
 
-  private static final String FETCH_WRITER_NODE_QUERY =
-      "SHOW REPLICA STATUS";
+  private static final String FETCH_WRITER_NODE_QUERY = "SHOW REPLICA STATUS";
 
   private static final String FETCH_WRITER_NODE_QUERY_HEADER = "Source_Server_Id";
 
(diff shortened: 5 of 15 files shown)


wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
```diff
@@ -51,7 +52,8 @@ public boolean isDialect(final Connection connection) {
 
   @Override
   public List</* dialect code */ String> getDialectUpdateCandidates() {
-    return null;
+    return Collections.singletonList(
```
brunos-bq, line 55: NIT: This could fit in one single line, i think?

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectCodes.java
```diff
@@ -20,9 +20,11 @@ public class DialectCodes {
   public static final String AURORA_MYSQL = "aurora-mysql";
   public static final String RDS_MYSQL = "rds-mysql";
   public static final String MYSQL = "mysql";
+  public static final String RDS_MULTI_AZ_MYSQL_CLUSTER = "rds-multi-az-mysql-cluster";
```
brunos-bq, line 23: NIT: Should we add a comment on the difference between this new line and RDS_MYSQL?

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
```diff
@@ -144,6 +144,7 @@ public Dialect getDialect(
     if (driverProtocol.contains("mysql")) {
       RdsUrlType type = this.rdsHelper.identifyRdsType(host);
       if (type.isRdsCluster()) {
+        this.canUpdate = true;
```
brunos-bq, line 147: Is this canUpdate valid for every RDS Cluster, or just for the new Multi AZ ones?

wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
```diff
@@ -29,7 +29,7 @@
 public class MysqlDialect implements Dialect {
 
   private static final List<String> dialectUpdateCandidates = Arrays.asList(
-      DialectCodes.AURORA_MYSQL, DialectCodes.RDS_MYSQL);
+      DialectCodes.AURORA_MYSQL, DialectCodes.RDS_MULTI_AZ_MYSQL_CLUSTER, DialectCodes.RDS_MYSQL);
```
brunos-bq, line 32: NIT: For readability, it could be better to have one item per line here

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -0,0 +1,70 @@
...
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.List;
+import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
+
+public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
+
+  private static final String TOPOLOGY_QUERY =
```
brunos-bq, line 28: NIT: Single line

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -0,0 +1,70 @@
...
+      "SELECT id, endpoint, port FROM mysql.rds_topology";
+
+  private static final String FETCH_WRITER_NODE_QUERY =
```
brunos-bq, line 31: NIT: Single line

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final String NODE_ID_QUERY = "SELECT @@server_id";
+  private static final String IS_READER_QUERY = "SELECT @@read_only";
+
+  @Override
+  public boolean isDialect(final Connection connection) {
+    try (final Statement stmt = connection.createStatement();
+         final ResultSet rs = stmt.executeQuery("SELECT id FROM mysql.rds_topology LIMIT 1")) {
+      if (rs.next()) {
+        // If the table is presented it means it's a Multi-AZ DB cluster
```
brunos-bq, line 44: NIT: // If the table exists ...

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final String FETCH_WRITER_NODE_QUERY_HEADER = "Source_Server_Id";
+
+  private static final String NODE_ID_QUERY = "SELECT @@server_id";
+  private static final String IS_READER_QUERY = "SELECT @@read_only";
+
+  @Override
+  public boolean isDialect(final Connection connection) {
+    try (final Statement stmt = connection.createStatement();
+         final ResultSet rs = stmt.executeQuery("SELECT id FROM mysql.rds_topology LIMIT 1")) {
```
brunos-bq, line 42: This query could also go as a class variable

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
```diff
@@ -0,0 +1,87 @@
...
+      exceptionHandler = new MultiAzDbClusterPgExceptionHandler();
+    }
+    return exceptionHandler;
+  }
+
+  @Override
+  public boolean isDialect(final Connection connection) {
+    try (final Statement stmt = connection.createStatement();
+         final ResultSet rs = stmt.executeQuery(
+             "SELECT * FROM rds_tools.multi_az_db_cluster_source_dbi_resource_id()")) {
```
brunos-bq, line 59: This query could also go as a class variable

wrapper/src/main/java/software/amazon/jdbc/dialect/UnknownDialect.java
```diff
@@ -19,9 +19,6 @@
 import java.sql.Connection;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Properties;
```
brunos-bq, line 22: NIT: Were they unused?

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public static final List<String> ACCESS_ERROR = Collections.singletonList(
+      "28P01"
+  );
+
+  @Override
+  public boolean isNetworkException(final Throwable throwable) {
+    Throwable exception = throwable;
+
+    while (exception != null) {
```
brunos-bq, line 47: Does that mean that we inspect all the layers of errors until either we reach the end, either get a SQLException?

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  public boolean isNetworkException(final String sqlState) {
+    if (sqlState == null) {
+      return false;
+    }
+
+    for (final String pgSqlState : NETWORK_ERRORS) {
+      if (sqlState.startsWith(pgSqlState)) {
```
brunos-bq, line 65: Is this checking prefixes? If that's the case, maybe it would be handy to add in the NETWORK_ERRORS item comments which ones are errors and which ones are prefixes

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+      "57P02", // crash shutdown
+      "57P03", // cannot connect now
+      "58", // system error (backend)
+      "08", // connection error
+      "99", // unexpected error
+      "F0", // configuration file error (backend)
+      "XX" // internal error (backend)
+  );
+
+  public static final List<String> ACCESS_ERROR = Collections.singletonList(
```
brunos-bq, line 39: Is it worth to have a list with one element? (If a new element appears, the code will have to be changed anyway)

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+    return false;
+  }
+
+  @Override
+  public boolean isLoginException(final String sqlState) {
+    if (sqlState == null) {
+      return false;
+    }
+    return ACCESS_ERROR.contains(sqlState);
```
brunos-bq, line 96: Wouldn't it be easier to check equals?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Obtain a cluster topology from database.
+   *
+   * @param conn A connection to database to fetch the latest topology.
+   * @return a list of {@link HostSpec} objects representing the topology
+   * @throws SQLException if errors occurred while retrieving the topology.
+   */
+  protected List<HostSpec> queryForTopology(final Connection conn) throws SQLException {
+
```
brunos-bq, line 72: NIT: Remove extra line

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /**
+   * Creates an instance of HostSpec which captures details about a connectable host.
+   *
+   * @param resultSet the result set from querying the topology
+   * @return a {@link HostSpec} instance for a specific instance from the cluster
+   * @throws SQLException If unable to retrieve the hostName from the result set
+   */
+  private HostSpec createHost(final ResultSet resultSet, final String writerNodeId) throws SQLException {
```
brunos-bq, line 179: NIT: createHostSpec ?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -128,9 +129,11 @@ public <T, E extends Exception> T execute(final Class<T> resultClass, final Clas
 
     } catch (final Exception e) {
       if (e instanceof FailoverSQLException) {
-        tracker.invalidateAllConnections(this.currentWriter);
-        tracker.logOpenedConnections();
-        this.needUpdateCurrentWriter = true;
+        if (!Objects.equals(this.getWriter(this.pluginService.getHosts()), this.currentWriter)) {
```
brunos-bq, line 132: Not sure if we should use Objects.equals - don't we have another way of checking the equality here?

wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
```diff
@@ -48,6 +51,15 @@ public static void applyProperties(final Object target, final Properties propert
     }
   }
 
+  public static void addConnAttrForClientIdentification(Dialect dialect, Properties props) {
```
brunos-bq, line 54: Name of the function abbreviates some words and not others, very confusing.
Would suggest addConnectionAttributesForClientIdentification()

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
```diff
@@ -0,0 +1,87 @@
...
+import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
+import software.amazon.jdbc.util.DriverInfo;
+
+public class RdsMultiAzDbClusterPgDialect extends PgDialect {
+
+  private static MultiAzDbClusterPgExceptionHandler exceptionHandler;
+
+  private static final String TOPOLOGY_QUERY =
+      "SELECT LOWER(id) AS id, LOWER(endpoint) AS endpoint, port FROM "
+      + "rds_tools.show_topology('aws_jdbc_driver-" + DriverInfo.DRIVER_VERSION + "')";
```
sergiyvamz, line 35: What's the reason to pass driver name and version?

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+package software.amazon.jdbc.exceptions;
+
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public class MultiAzDbClusterPgExceptionHandler implements ExceptionHandler {
```
sergiyvamz, line 24: What's the differences comparing to PgExceptionHandler? Does it make sense to use an existing handler?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
sergiyvamz: What this condition does?

wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
sergiyvamz: This logic is dialect-specific. For the moment it's required just for Mysql TAZ. I'd suggest to extend Dialect interface with the following method (just a row idea):
void prepareProperties(Properties props)

This method does nothing for all implementations of Dialect interface. RdsMultiAzDbClusterMysqlDailect will be the only one that add extra properties in such method.

prepareProperties() can be executed in DriverConnectionProvider and DataSourceConnectionProvider.

wrapper/src/main/java/software/amazon/jdbc/DriverConnectionProvider.java
```diff
@@ -114,9 +114,11 @@ public Connection connect(
       final @NonNull Properties props)
       throws SQLException {
 
-    final Properties copy = PropertyUtils.copyProperties(props);
+    final Properties copiedProps = PropertyUtils.copyProperties(props);
+    PropertyUtils.addConnAttrForClientIdentification(dialect, copiedProps);
```
sergiyvamz, line 118: Need to call 
dialect.prepareProperties(copiedProps)

See my comments for PropertyUtils

sergiyvamz: Need to add similar changes to DataSourceConnectionProvider

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
```diff
@@ -187,14 +188,16 @@ public Connection connect(
       @NonNull HostSpec hostSpec,
       @NonNull Properties props)
       throws SQLException {
+    final Properties copiedProps = PropertyUtils.copyProperties(props);
+    PropertyUtils.addConnAttrForClientIdentification(dialect, copiedProps);
```
sergiyvamz, line 192: Need similar changes as for DriverConnectionProvider/DataSourceConnectionProvider

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
```diff
@@ -0,0 +1,87 @@
...
+import software.amazon.jdbc.exceptions.MultiAzDbClusterPgExceptionHandler;
+import software.amazon.jdbc.hostlistprovider.RdsMultiAzDbClusterListProvider;
+import software.amazon.jdbc.util.DriverInfo;
+
+public class RdsMultiAzDbClusterPgDialect extends PgDialect {
+
+  private static MultiAzDbClusterPgExceptionHandler exceptionHandler;
+
+  private static final String TOPOLOGY_QUERY =
+      "SELECT LOWER(id) AS id, LOWER(endpoint) AS endpoint, port FROM "
```
davecramer, line 34: do we really need LOWER here ? I think all of the data is already in lower case

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Get writer node ID.
+   *
+   * @param fetchWriterNodeResultSet A ResultSet of writer node query
+   * @return String The ID of a writer node
+   * @throws SQLException if errors occurred while retrieving the topology
+   */
+  private String processWriterNodeId(final ResultSet fetchWriterNodeResultSet) throws SQLException {
+    String writerNodeId = "";
```
davecramer, line 113: better to use NULL as it avoids creating an object.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    while (fetchWriterNodeResultSet.next()) {
```
davecramer, line 114: is it possible that there are two writers?

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
```diff
@@ -0,0 +1,87 @@
...
+      + "rds_tools.show_topology('aws_jdbc_driver-" + DriverInfo.DRIVER_VERSION + "')";
+
+  private static final String FETCH_WRITER_NODE_QUERY =
+      "SELECT LOWER(multi_az_db_cluster_source_dbi_resource_id) AS multi_az_db_cluster_source_dbi_resource_id FROM "
+      + "rds_tools.multi_az_db_cluster_source_dbi_resource_id()";
+
+  private static final String FETCH_WRITER_NODE_QUERY_HEADER = "multi_az_db_cluster_source_dbi_resource_id";
```
davecramer, line 41: QUERY_HEADER seems like the wrong name

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
davecramer: Yes, HostSpec implements equals so no need to use Object.equals()

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
fangli: The `Multi-AZ DB Cluster` is a product name with very specific supported feature: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html It's a bit unclear on whether `canUpdate` is valid for every RDS cluster because we can hardly define `every RDS cluster`.

The overall "update chain" is Aurora -> MultiAZDbCluster -> Other generic rds DBs -> generic DBs, from explicits to implicits.

wrapper/src/main/java/software/amazon/jdbc/dialect/UnknownDialect.java
fangli: Yes

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
fangli: > Does that mean that we inspect all the layers of errors until either we reach the end, either get a SQLException?
Yes. This is align with the default PgExceptionhandler.

> how do we handle this in other dialects ?
The only difference between MultiAZ handler and default Pg handler is, they take the error code `28000` differently.

* In other clusters (like Aurora), the `28000` means "access denied by IAM auth", leading to a fatal error.
* But in MultiAZ DB Cluster, The PG may raise `28000` during cluster rebooting, meaning "deny access and queries due to rebooting". The driver should retry and keep on the failover logic.

As a result, the IAM auth plugin will be incompatitable with MultiAZ db cluster. Because the driver can not distinguish the scenario of `28000` in a decent way in the current design.

Afterall, a refactor is expected because the existing class `PgExceptionHandler` and `MultiAzDbClusterPgExceptionHandler` are essentially the same. I would keep the refactor out of currrent PR for easier review.

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
fangli: This will be addressed in follow-up refactor.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
fangli: The `host` in the context is pretty much `HostSpec`. I'll keep it align with the default hostlist provider.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
fangli: > What this condition does?

In case the cluster gets rebooted without failover.

> Not sure if we should use Objects.equals - don't we have another way of checking the equality here?

Is there any concerns here? like efficiency or something?

`Objects.equals` is used because it handles `null` value. `HostSpec.equals()` may throw Nullpointer if the variable is `Null`.

wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
fangli: Good idea. I would keep this change out of this PR since it'll noticeably modifies existing classes.

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
fangli: `28000` is the only change. we should refactor it as a whole.

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
fangli: No, we don't, now. and removed.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
fangli: Not something I'm aware of, see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterPgDialect.java
fangli: Will use `FETCH_WRITER_NODE_QUERY_COLUMN_NAME`.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -0,0 +1,197 @@
...
+
+    int writerCount = writers.size();
+
+    if (writerCount == 0) {
+      LOGGER.severe(() -> Messages.get("RdsHostListProvider.invalidTopology"));
+      hosts.clear();
+    } else if (writerCount == 1) {
+      hosts.add(writers.get(0));
+    } else {
+      // Take the latest updated writer node as the current writer. All others will be ignored.
```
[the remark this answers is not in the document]
fangli: yeah the logic can be cleaned up. it's copied from the base class, a bit redundant.

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
davecramer: update is probably the wrong word. Should be upgrade

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
davecramer: if that is the case why do we have a loop here ?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
davecramer: No it won't return a NullPointerException `null instanceof HostSpec` returns false

wrapper/src/main/java/software/amazon/jdbc/exceptions/MultiAzDbClusterPgExceptionHandler.java
fangli: refactored with latest PR.

wrapper/src/main/java/software/amazon/jdbc/dialect/DialectManager.java
```diff
@@ -165,6 +166,7 @@ public Dialect getDialect(
     if (driverProtocol.contains("postgresql")) {
```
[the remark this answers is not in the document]
fangli: This has been out for very long. I wouldn't have concern.

If we decide to loose the letter case for better user experience, we'll also need to recheck all occurrences.

sergiyvamz (approved):

brunos-bq (approved):

karenc-bq (approved):

sergiyvamz (approved):

Issue #612, opened 2023-08-22
Prevent possible race condition on IAM token expiry


chlunde:
### Describe the feature

We had some DB connections fail after when using this driver and IAM authentication. This was fixed by [2.1.2](https://github.com/awslabs/aws-advanced-jdbc-wrapper/releases/tag/2.1.2) and newer via https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/456 and https://github.com/awslabs/aws-advanced-jdbc-wrapper/pull/457.

But I don't think this code path should have triggered at all, because tokens should be refreshed in time.

In the logs, we still see a few `"pam_authenticate failed: Permission denied"` every hour. This will be very confusing when diagnosing other issues. I also expect

### Use Case

No unnecessary reconnects, stable latency. No confusing log messages about expired token.

### Proposed Solution

I would expect there to be a small grace period to prevent "TOCTTOU" race conditions, where the client code thinks it is valid when checking it, but the token is invalid when the server side receives and validates it.

This code uses exactly 900 seconds as a default. Subtracting somewhere between 2-30 seconds should be enough.

https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/708d8a240ef2cca9a7fc993f940688369374f26b/wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthConnectionPlugin.java#L53-L54

This should also improve P99 latencies.

You could also consider background refreshes, but I'm not sure if that's a good idea, it certainly adds complexity.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

2.1.2+

### JDK version used

17

### Operating System and version

bottlerocket

crystall-bitquill:
Hi @chlunde,

Thanks for reaching out and raising this issue.

We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

davecramer:
@chlunde Interestingly enough when I tested how long an IAM token was valid for it was more like 20 minutes. Have you done any experiments around this ? I am very curious how you are getting these errors if this is really the case.

chlunde:
[...]

PR #706, opened 2023-10-26
refactor: add buffer to iam token expiry and create token expiry time earlier

crystall-bitquill:
### Summary

Update the IAM plugin 

### Description

- made the default `iamExpiration` time 14 minutes and 30 seconds
- created the token cache expiry time before generating an IAM token instead of after

Related to #612 


wrapper/src/main/java/software/amazon/jdbc/plugin/IamAuthConnectionPlugin.java
```diff
@@ -127,7 +127,7 @@ private Connection connectInternal(String driverProtocol, HostSpec hostSpec, Pro
         ? getRdsRegion(host)
         : Region.of(iamRegion);
 
-    final int tokenExpirationSec = IAM_EXPIRATION.getInteger(props);
+    final int tokenExpirationSec = Math.min(IAM_EXPIRATION.getInteger(props), DEFAULT_TOKEN_EXPIRATION_SEC);
```
sergiyvamz, line 130: This logic will override user settings. I'd preserve user choice in any case.

sergiyvamz (approved):

Issue #630, opened 2023-09-05
Add endpoint configurability to AWS Secrets Manager Plugin


pennytextures:
### Describe the feature

It would be nice if the secrets manager plugin would take some configuration via properties to use an endpoint, like so:

secretsManagerEndpoint: http://localhost:4566

This would allow it to work with secrets manager running from localstack.

### Use Case

We'd like to use the jdbc wrapper, but in development environments against localstack we'd have to reimplement our own aws secrets manager plugin and make the endpoint configurable, it'd be much nicer if we could just do this via a property.

### Proposed Solution

`  public static final AwsWrapperProperty ENDPOINT_PROPERTY = new AwsWrapperProperty(
      "secretsManagerEndpoint", null,
      "The secrets manager endpoint of the secret to retrieve.");`
      
 Then if this property isn't null, include it in the client builder

### Other Information

This wouldn't be an issue if we developed against real aws resources, but at the scale of our project this is not financially viable.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

2.2.4

### JDK version used

17.0.8

### Operating System and version

amazoncorretto:17.0.8-al2

crystall-bitquill:
Hi @pennytextures,

Thanks for reaching out and raising this issue.

We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

crystall-bitquill:
Hi @pennytextures,

We've recently merged some changes to add this feature (#707). Could you kindly check out our latest [snapshot build](https://github.com/awslabs/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#using-a-snapshot-of-the-driver) and let us know if the changes work for you?

Thank you!

pennytextures:
i’ll try and revive our attempt to integrate it with our codebase next week :)On 17 Nov 2023, at 21:15, crystall-bitquill ***@***.***> wrote:﻿
Hi @pennytextures,
I wanted to check in and see if you were able to verify the endpoint configurability was working. If there are no further updates on this ticket in the next few days, it will be closed. However, if you have any other questions or concerns please feel free to reopen this issue or to create a new one.
Thank you!

PR #707, opened 2023-10-27
feat: allow endpoint override for secrets manager plugin

crystall-bitquill:
### Summary

Allow endpoint override for secrets manager plugin

### Description

- adds the configuration parameter `secretsManagerEndpoint` to allow users to set the endpoint override for secret retrieval in the secrets manager plugin

Addresses #630 

### DIFF
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
@@ -12,10 +12,11 @@ The following properties are required for the AWS Secrets Manager Connection Plu
 
 > **Note:** To use this plugin, you will need to set the following AWS Secrets Manager specific parameters.
 
-| Parameter                | Value  |                         Required                         | Description                                             | Example     | Default Value |
-|--------------------------|:------:|:--------------------------------------------------------:|:--------------------------------------------------------|:------------|---------------|
-| `secretsManagerSecretId` | String |                           Yes                            | Set this value to be the secret name or the secret ARN. | `secretId`  | `null`        |
-| `secretsManagerRegion`   | String | Yes unless the `secretsManagerSecretId` is a Secret ARN. | Set this value to be the region your secret is in.      | `us-east-2` | `us-east-1`   |
+| Parameter                | Value  |                         Required                         | Description                                                                                                                                                                                                                      | Example                 | Default Value |
+|--------------------------|:------:|:--------------------------------------------------------:|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:------------------------|---------------|
+| `secretsManagerSecretId` | String |                           Yes                            | Set this value to be the secret name or the secret ARN.                                                                                                                                                                          | `secretId`              | `null`        |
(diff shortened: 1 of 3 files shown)


sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -92,14 +97,30 @@ public class AwsSecretsManagerConnectionPlugin extends AbstractConnectionPlugin
...
+          final String endpoint = ENDPOINT_PROPERTY.getString(props);
+          if (endpoint != null && !endpoint.isEmpty()) {
+            try {
+              final URI endpointURI = new URI(endpoint);
+              return SecretsManagerClient.builder()
+                  .credentialsProvider(AwsCredentialsManager.getProvider(hostSpec, props))
+                  .endpointOverride(endpointURI)
+                  .region(region)
+                  .build();
+            } catch (URISyntaxException e) {
```
brunos-bq, line 113: is this the only exception we can have from the endpoint override?
What happens if the syntax is ok but the endpoint doesn't lead anywhere?
is there a way to see all the exceptions this call can throw?

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
```diff
@@ -12,10 +12,11 @@ The following properties are required for the AWS Secrets Manager Connection Plu
...
+| Parameter                | Value  |                         Required                         | Description                                                              | Example                 | Default Value |
+|--------------------------|:------:|:--------------------------------------------------------:|:-------------------------------------------------------------------------|:------------------------|---------------|
+| `secretsManagerSecretId` | String |                           Yes                            | Set this value to be the secret name or the secret ARN.                  | `secretId`              | `null`        |
+| `secretsManagerRegion`   | String | Yes unless the `secretsManagerSecretId` is a Secret ARN. | Set this value to be the region your secret is in.                       | `us-east-2`             | `us-east-1`   |
+| `secretsManagerEndpoint` | String |                            No                            | Set this value to be the endpoint override to retrieve your secret from. | `http://localhost:1234` | `null`        |
```
brunos-bq, line 19: We need to indicate that it is a URL in the description, and also indicate which formats are OK or not (ex: port mandatory or not)

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -72,6 +74,9 @@ public class AwsSecretsManagerConnectionPlugin extends AbstractConnectionPlugin
   public static final AwsWrapperProperty REGION_PROPERTY = new AwsWrapperProperty(
       "secretsManagerRegion", "us-east-1",
       "The region of the secret to retrieve.");
+  public static final AwsWrapperProperty ENDPOINT_PROPERTY = new AwsWrapperProperty(
```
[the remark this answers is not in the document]
crystall-bitquill: I think it's ok without the prefix, since these properties are only used in the plugin and there isn't any ambiguity on what the they are.

brunos-bq (approved):

PR #718, opened 2023-10-31
Fix stopped monitoring threads

sergiyvamz:
### Summary

Fix stopped monitoring threads

### Description

- Improve exception logging in monitoring thread
- Increase monitoring thread idle time to 10min
- Make sure getting an existing monitor, creating a new monitor, and destroying a monitor are well-synchronized. 
- Avoid caching a reference to monitor object since it may be destroyed; use weak pointer instead.
- Avoid storing and then reusing of idle/stopped monitors

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@crystall-bitquill 
@aaronchung-bitquill 


wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorServiceImpl.java
```diff
@@ -100,16 +100,14 @@ public MonitorConnectionContext startMonitoring(
           new Object[] {hostSpec}));
     }
 
-    final Monitor monitor;
-    if (this.cachedMonitor == null
+    Monitor monitor = this.cachedMonitor == null ? null : this.cachedMonitor.get();
```
karenc-bq, line 103:
```suggestion
    final Monitor monitor = this.cachedMonitor == null ? null : this.cachedMonitor.get();
```

crystall-bitquill (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/MonitorServiceImpl.java
sergiyvamz: Can't do that. Monitor maybe reassigned later (line 109).

karenc-bq (approved):

PR #723, opened 2023-11-06
fix: using masked properties for logging

brunos-bq:
### Summary

fix: using masked properties for logging

### Additional Reviewers

@sergiyv-improving 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
@@ -19,16 +19,22 @@ package software.amazon.jdbc.util;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
```diff
@@ -96,7 +96,8 @@ public static void setPropertyOnTarget(
       } else {
         writeMethod.invoke(target, propValue);
       }
-      LOGGER.finest(() -> String.format("Set property '%s' with value: %s", propName, propValue));
+      Object loggedProperty = propValue.equals(PropertyDefinition.PASSWORD.name) ? "***" : propValue;
```
sergiyvamz, line 99: I'd extract a code that check if the property contains a secret into a separate function and use it everywhere here in PropertyUtils

wrapper/src/main/java/software/amazon/jdbc/util/PropertyUtils.java
```diff
@@ -127,6 +134,10 @@ public static void setPropertyOnTarget(
     return copy;
   }
 
+  public static boolean isSecretProperty(final Object propertyKey) {
```
sergiyvamz, line 137: private?

sergiyvamz (approved):

karenc-bq (approved):

PR #705, opened 2023-10-25
Allow user to set up a lambda to initialize new connections

sergiyvamz:
### Summary

Allow user to set up a lambda to initialize new connections.

### Description

User needs to set an connection initialization function with ConnectionProviderManager.setConnectionInitFunc()
Related to https://github.com/awslabs/aws-advanced-jdbc-wrapper/issues/679

### Additional Reviewers

@karenc-bq 
@crystall-bitquill 
@brunos-bq 
@aaronchung-bitquill 

### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #729, opened 2023-11-07
Track session states

sergiyvamz:
### Summary

- track session states
- allow user to provide custom function to handle session state transfers

### Additional Reviewers

@karenc-bq 
@davecramer 
@hsuamz 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
@@ -560,23 +560,23 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
   }
 
   /**
-   * Transfers basic session state from one connection to another.
+   * Transfers session state from one connection to another.
    *
-   * @param from The connection to transfer state from
-   * @param fromHostSpec The connection {@link HostSpec} to transfer state from
-   * @param to   The connection to transfer state to
-   * @param toHostSpec The connection {@link HostSpec} to transfer state to
+   * @param src The connection to transfer state from
+   * @param srcHostSpec The connection {@link HostSpec} to transfer state from
+   * @param dest   The connection to transfer state to
+   * @param destHostSpec The connection {@link HostSpec} to transfer state to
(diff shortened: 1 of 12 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -538,22 +563,38 @@ private void switchCurrentConnectionTo(final HostSpec host, final Connection con
    * Transfers basic session state from one connection to another.
```
brunos-bq, line 563: NIT: add a line about session states in the method doc

wrapper/src/main/java/software/amazon/jdbc/states/SessionStateHelper.java
```diff
@@ -0,0 +1,99 @@
...
+   * Transfers basic session state from one connection to another.
+   *
+   * @param sessionState Session state of from-connection
+   * @param from The connection to transfer state from
+   * @param to   The connection to transfer state to
+   * @throws SQLException if a database access error occurs, this method is called on a closed connection, this
+   *                      method is called during a distributed transaction, or this method is called during a
+   *                      transaction
+   */
+  public void transferSessionState(
```
brunos-bq, line 37: NIT: wondering if it wouldn't be clearer to have `src` and `dest` instead of `from` and `to` (ex: `dest.setReadOnly(src.isReadOnly());`)

karenc-bq (approved):

scott-long-adp:
thanks for this -- we will be using ConnectionProviderManager.setConnectionInitFunc() from #705 to initialize new connections that were not able to go through PoolBase.setupConnection() 

Regarding SessionStateTransferCallable, RestoreSessionStateCallable - 

The default behavior for SessionStateTransferCallable is to propagate all properties from one connection to the other, including transaction isolation -- so this new default behavior may actually be sufficient. 

But the default behavior for RestoreSessionStateCallable is only to restore a locally-saved readonly and autocommit property.  

What is the reason for the limited number of properties restored by RestoreSessionStateCallable compared to SessionStateTransferCallable?  Should we consider a RestoreSessionStateCallable event similar to ConnectionProviderManager.setConnectionInitFunc() and just reset all properties from the ground up?

thanks for any insight

PR #739, opened 2023-11-16
fix: use existing entries to update the round robin cache

crystall-bitquill:
### Summary

Use existing entries to update the round robin cache

### Description

The round robin cache is updated if all hosts have a cache entry, and if there are no existing cache entries for any host, new entries will be made for all hosts. This is missing the case where only some hosts have cache entries available. This PR amends the update cache conditions to include this scenario.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/RoundRobinHostSelector.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/RoundRobinHostSelector.java
@@ -107,27 +107,26 @@ public class RoundRobinHostSelector implements HostSelector {
       final @NonNull List<HostSpec> hosts,
       final @Nullable Properties props)
       throws SQLException {
-    final List<HostSpec> hostsMissingCacheEntry = new ArrayList<>();
     final List<HostSpec> hostsWithCacheEntry = new ArrayList<>();
     for (final HostSpec host : hosts) {
       if (roundRobinCache.get(host.getHost()) != null) {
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/RoundRobinHostSelector.java
```diff
@@ -107,27 +107,24 @@ private void createCacheEntryForHosts(
...
             host.getHost(),
             roundRobinCache.get(hostsWithCacheEntry.get(0).getHost()),
             DEFAULT_ROUND_ROBIN_CACHE_EXPIRE_NANO);
       }
-    } else if (hostsWithCacheEntry.isEmpty()) {
+    } else {
       final RoundRobinClusterInfo roundRobinClusterInfo = new RoundRobinClusterInfo();
       updateCachePropertiesForRoundRobinClusterInfo(roundRobinClusterInfo, props);
-      for (final HostSpec host : hostsMissingCacheEntry) {
+      for (final HostSpec host : hosts) {
```
brunos-bq, line 127: does that mean that a new cache entry will be added for every host? including those that already had a cache entry?

let's add comments explaining each part of the conditions

crystall-bitquill: Yes, each host will have a cache entry. If the host already had an entry, it will just be updated.

sergiyvamz (approved):

PR #742, opened 2023-11-20
Disable failing integration test for PG driver

sergiyvamz:
### Summary

Disable failing integration test for PG driver

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@crystall-bitquill 
@brunos-bq 

### DIFF
--- a/wrapper/src/test/java/integration/container/TestDriverProvider.java
+++ b/wrapper/src/test/java/integration/container/TestDriverProvider.java
@@ -165,6 +165,10 @@ public class TestDriverProvider implements TestTemplateInvocationContextProvider
                       try {
                         instanceIDs = auroraUtil.getAuroraInstanceIds();
                       } catch (SQLException ex) {
+                        if ("28P01".equals(ex.getSQLState())) {
+                          // Authentication error for PG
+                          throw ex;
+                        }
                         instanceIDs = new ArrayList<>();
                       }
                     }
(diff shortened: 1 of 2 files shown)


karenc-bq (approved):

wrapper/src/test/java/integration/container/TestDriverProvider.java
```diff
@@ -165,6 +165,10 @@ public void beforeEach(ExtensionContext context) throws Exception {
                       try {
                         instanceIDs = auroraUtil.getAuroraInstanceIds();
                       } catch (SQLException ex) {
+                        if ("28P01".equals(ex.getSQLState())) {
```
karenc-bq, line 168: Could this error be thrown by RDS databases during switchover?

brunos-bq, line 168: NIT: can we do a global static variable with the name of the error code? like POSTGRES_AUTH_ERROR_CODE

brunos-bq (approved):

wrapper/src/test/java/integration/container/TestDriverProvider.java
sergiyvamz: It's hard to say. The goal here is to detect authentication errors (i.e. test environment errors that are rare but may happen) and stop the tests.

sergiyvamz: We know that error 28000 (not 28P01) is raised when TAZ cluster is switching over.

PR #748, opened 2023-11-21
feat: Autoregister a target driver

sergiyvamz:
### Summary

Automatically register the target driver in the class path

### Additional Reviewers

@karenc-bq 
@brunos-bq 
@crystall-bitquill 
@aaron-congo 
@aaronchung-bitquill 


docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -47,19 +47,19 @@ The AWS JDBC Driver also has a parameter, [`wrapperLoggerLevel`](#aws-advanced-j
...
+| `socketTimeout`                   | `Integer` | No       | Socket timeout in milliseconds.                                                                                                                                                                                                                                                               | `null`        |
+| `tcpKeepAlive`                    | `Boolean` | No       | Enable or disable TCP keep-alive probe.                                                                                                                                                                                                                                                       | `false`       |
+| `wrapperTargetDriverAutoRegister` | `Boolean` | No       | Allows the AWS JDBC Driver to register a target driver based on `wrapperTargetDriverDialect` configuration parameter or, if it's missed, on a connection url protocol.                                                                                                                        | `true`        |
```
brunos-bq, line 62: i don't think we should use or add the `wrapper` prefix to any property

`targetDriverAutoRegister` or even just `autoRegister`

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -232,6 +214,47 @@ public DriverPropertyInfo[] getPropertyInfo(final String url, final Properties i
     return props;
   }
 
+  private java.sql.Driver getTargetDriver(
+      final @NonNull String driverUrl,
```
brunos-bq, line 218: NIT: `driverURL`

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -232,6 +214,47 @@ public DriverPropertyInfo[] getPropertyInfo(final String url, final Properties i
...
+      final @NonNull Properties props)
+      throws SQLException {
+
+    final ConnectionUrlParser connectionUrlParser = new ConnectionUrlParser();
```
brunos-bq, line 222: what about
```
final ConnectionUrlParser parser = new ConnectionUrlParser();
```

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -259,6 +260,49 @@ ConnectionWrapper createConnectionWrapper(
         telemetryFactory);
   }
 
+  private java.sql.Driver getTargetDriver(
```
brunos-bq, line 263: this function is the same than the one from Driver - is there a way to have it only once and both places use it?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/TargetDriverDialectManager.java
```diff
@@ -39,6 +40,10 @@ public class TargetDriverDialectManager implements TargetDriverDialectProvider {
       "wrapperTargetDriverDialect", "",
       "A unique identifier for the target driver dialect.");
 
+  public static final AwsWrapperProperty TARGET_DRIVER_AUTO_REGISTER = new AwsWrapperProperty(
+      "wrapperTargetDriverAutoRegister", "true",
```
brunos-bq, line 44: let's remove the wrapper prefix

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
sergiyvamz: I extracted this logic to TargetDriverHelper class.

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -134,26 +135,8 @@ public Connection connect(final String url, final Properties info) throws SQLExc
...
-      if (driver == null) {
-        final List<String> registeredDrivers = Collections.list(DriverManager.getDrivers())
-            .stream()
-            .map(x -> x.getClass().getName())
-            .collect(Collectors.toList());
-        LOGGER.severe(() -> Messages.get("Driver.missingDriver", new Object[] {driverUrl, registeredDrivers}));
-        return null;
-      }
+      TargetDriverHelper helper = new TargetDriverHelper();
+      java.sql.Driver driver = helper.getTargetDriver(driverUrl, props);
```
brunos-bq, line 139: NIT: is `getTargetDriver` the right name for this function?
we first try to find if the driver is registered, then try to register the driver of the dialect

after this line, we either had a driver ready, either registered a new driver, either crashed
would `instantiateDriver` a better fit?

wrapper/src/main/java/software/amazon/jdbc/TargetDriverHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Properties;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import software.amazon.jdbc.targetdriverdialect.TargetDriverDialectManager;
+import software.amazon.jdbc.util.ConnectionUrlParser;
+import software.amazon.jdbc.util.Messages;
+
+public class TargetDriverHelper {
+
+  public java.sql.Driver getTargetDriver(
```
brunos-bq, line 32: NIT: it's a really minor thing but could we add some comments in some parts of the conditional just to explain the situation and what we're doing in that case

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/TargetDriverDialectManager.java
```diff
@@ -121,4 +139,40 @@ private void logDialect(final String dialectCode, final TargetDriverDialect targ
         "TargetDriverDialectManager.useDialect",
         new Object[] {dialectCode, targetDriverDialect}));
   }
+
+  public boolean registerDriver(
```
brunos-bq, line 143: NIT: can we add some comments on what we're doing in some conditions?

brunos-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/Driver.java
sergiyvamz: This method either return a target driver, or raises an exception. 

What could be alternatives for `getTargetDriver` method name?

PR #753, opened 2023-11-24
Handle empty string sql query

fisco-unimatic:
Don't throw  IndexOutOfBoundsException in SqlMethodAnalyzer when executing empty sql query

### Summary

As mentioned in [the discussion about JBoss EAP datasource migration](https://github.com/awslabs/aws-advanced-jdbc-wrapper/discussions/217#discussioncomment-7661887), the [PostgreSQLValidConnectionChecker](https://github.com/oloake/IronJacamar/blob/master/adapters/src/main/java/org/jboss/jca/adapters/jdbc/extensions/postgres/PostgreSQLValidConnectionChecker.java#L61C10-L61C27) tries to execute an empty string statement. This results in an IndexOutOfBoundsException.

### Description

When given an empty SQL statement, I've made `SqlMethodAnalyzer#getFirstSqlStatement` return an empty string in a singleton list instead of an empty list. So now it is safe to get and invoke String methods on the first element of the returned list.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/SqlMethodAnalyzer.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/SqlMethodAnalyzer.java
@@ -16,10 +16,13 @@
 
 package software.amazon.jdbc.util;
 
+import static java.util.Collections.singletonList;
+
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/SqlMethodAnalyzer.java
```diff
@@ -55,14 +58,14 @@ private String getFirstSqlStatement(final String sql) {
 
   private List<String> parseMultiStatementQueries(String query) {
     if (StringUtils.isNullOrEmpty(query)) {
-      return new ArrayList<>();
+      return singletonList("");
```
karenc-bq, line 61: Can you please use `Collections.singletonList("");` instead to avoid static imports?

aaronchung-bitquill:
Hi @fisco-unimatic 
Thank you for creating this pull request and contributing to the AWS Advanced JDBC Wrapper driver. 
We have a fix in the [latest version of the driver](https://search.maven.org/artifact/software.amazon.jdbc/aws-advanced-jdbc-wrapper/2.3.3/jar?eh=) that addresses the issue you have raised.
Can you check it out and let us know if the issue still persists? 
Thank you

PR #764, opened 2023-11-30
Enhanced host monitoring plugin ver.2

sergiyvamz:
### Summary

Enhanced host monitoring plugin ver.2 (redesign)

### Description

- Redesigned version of Host Monitoring Plugin (aka EFM)
- Fixes for performance integration tests

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@aaronchung-bitquill 
@crystall-bitquill 
@davecramer 


docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -157,20 +157,20 @@ ConnectionProviderManager.setConnectionInitFunc((connection, protocol, hostSpec,
...
...                                                                                                                               | None                                                                                                                                                                                                          || Data Cache Connection Plugin                                                                   | `dataCache`               | Any database           | Caches results from SQL queries matching the regular expression specified in the  `dataCacheTriggerCondition` configuration parameter.                                                                                                                                                                                                                                                                                                                                                                                 | None                                                                                                                                                                                                          |
```
crystall-bitquill, line 164:
```suggestion
| [Host Monitoring Connection Plugin v2](./using-plugins/UsingTheHostMonitoringPlugin.md#host-monitoring-plugin-v2)        | `efm2`                    | Aurora, RDS Multi-AZ DB Cluster | Enables enhanced host connection failure monitoring, allowing faster failure detection rates. This plugin is an alternative implementation for host health status monitoring. It is functionally the same as the `efm` plugin and uses the same configuration parameters. This plugin is experimental.                                                                                                                                                                                                                                | None                                                                                                                                                                                                          |
| Data Cache Connection Plugin                                                                   | `dataCache`               | Any database           | Caches results from SQL queries matching the regular expression specified in the  `dataCacheTriggerCondition` configuration parameter.                                                                                                                                                                                                                                                                                                                                                                                 | None                                                                                                                                                                                                          |
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+
+# Host Monitoring Plugin v2
+This plugin is an alternative implementation of host health status monitoring and it's functionally equal to Host Monitoring Plugin described above. Both plugins share the same set of configuration parameters as mentioned above. `efm2` plugin could be used in any scenarios where `efm` plugin is mentioned. Since these two plugin are two different plugins, user may decide to use them together within a single connection. Such usage won't suppose to make any negative side effects. However, it's not recommended. It's recommended to use either `efm` plugin, or `efm2` plugin where it's needed.
```
crystall-bitquill, line 77:
```suggestion
This plugin is an alternative implementation of host health status monitoring and it is functionally equal to the Host Monitoring Plugin described above. Both plugins share the same set of [configuration parameters](#enhanced-failure-monitoring-parameters). The `efm2` plugin can be used in any scenario where the `efm` plugin is mentioned. Since these two plugins are separate plugins, users may decide to use them together with a single connection. While this should not  have any negative side effects, it is not recommended. It is recommended to use either the `efm` plugin, or the `efm2` plugin where it's needed. 
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+
+# Host Monitoring Plugin v2
+This plugin is an alternative implementation of host health status monitoring and it's functionally equal to Host Monitoring Plugin described above. Both plugins share the same set of configuration parameters as mentioned above. `efm2` plugin could be used in any scenarios where `efm` plugin is mentioned. Since these two plugin are two different plugins, user may decide to use them together within a single connection. Such usage won't suppose to make any negative side effects. However, it's not recommended. It's recommended to use either `efm` plugin, or `efm2` plugin where it's needed. 
+
+`efm2` plugin is designed to address some of the issue that have been reported by multiple users:
```
crystall-bitquill, line 79:
```suggestion
The `efm2` plugin is designed to address some of the issues that have been reported by multiple users. The following changes have been made:
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+- Use of weak pointers to ease GC
```
crystall-bitquill, line 80:
```suggestion
- Used weak pointers to ease garbage collection
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+- Split monitoring logic into two separate threads to increase overall monitoring stability
+- Reviewed locks for monitoring context
+- Stopping of idle monitoring threads is reviewed and redesigned
```
crystall-bitquill, line 83:
```suggestion
- Reviewed and redesigned stopping of idle monitoring threads
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+- Overall, monitoring logic is reviewed and simplified
```
crystall-bitquill, line 84:
```suggestion
- Reviewed and simplified monitoring logic
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,15 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+
+>:warning: This plugin is experimental and should be verified before using it in production environment.
```
crystall-bitquill, line 86: Maybe this should be moved to just under the `Host Monitoring Plugin v2` title?

.github/workflows/mysql_advanced_performance.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  aurora-mysql-performance-tests:
+    concurrency: AvancedPerformanceTests-Aurora
+    name: 'Run Aurora MySQL container advanced performance tests'
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Clone repository'
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 50
```
karenc-bq, line 15: I think we can just use the default fetch-depth 1 for the testing workflows.

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,17 @@ properties.setProperty("monitoring-socketTimeout", "10");
 > We recommend you either disable the Host Monitoring Connection Plugin or avoid using RDS Proxy endpoints when the Host Monitoring Connection Plugin is active.
 >
 > Although using RDS Proxy endpoints with the AWS Advanced JDBC Driver with Enhanced Failure Monitoring doesn't cause any critical issues, we don't recommend this approach. The main reason is that RDS Proxy transparently re-routes requests to a single database instance. RDS Proxy decides which database instance is used based on many criteria (on a per-request basis). Switching between different instances makes the Host Monitoring Connection Plugin useless in terms of instance health monitoring because the plugin will be unable to identify which instance it's connected to, and which one it's monitoring. This could result in false positive failure detections. At the same time, the plugin will still proactively monitor network connectivity to RDS Proxy endpoints and report outages back to a user application if they occur.
+
+# Host Monitoring Plugin v2
+
+>:warning: This plugin is experimental and should be verified before using it in production environment.
```
karenc-bq, line 78:
```suggestion
> [!WARNING] This plugin is experimental and should be verified before using it in production environment.
```

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,17 @@ properties.setProperty("monitoring-socketTimeout", "10");
...
+
+This plugin is an alternative implementation of host health status monitoring and it is functionally equal to the Host Monitoring Plugin described above. Both plugins share the same set of [configuration parameters](#enhanced-failure-monitoring-parameters). The `efm2` plugin can be used in any scenario where the `efm` plugin is mentioned. Since these two plugins are separate plugins, users may decide to use them together with a single connection. While this should not  have any negative side effects, it is not recommended. It is recommended to use either the `efm` plugin, or the `efm2` plugin where it's needed.
```
karenc-bq, line 80: Can we use `Enhanced Failure Monitoring` instead of `host health status monitoring` since we mentioned in earlier? It feels like we are introducing a new concept with `host health status monitoring`

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -72,3 +72,17 @@ properties.setProperty("monitoring-socketTimeout", "10");
 > We recommend you either disable the Host Monitoring Connection Plugin or avoid using RDS Proxy endpoints when the Host Monitoring Connection Plugin is active.
 >
 > Although using RDS Proxy endpoints with the AWS Advanced JDBC Driver with Enhanced Failure Monitoring doesn't cause any critical issues, we don't recommend this approach. The main reason is that RDS Proxy transparently re-routes requests to a single database instance. RDS Proxy decides which database instance is used based on many criteria (on a per-request basis). Switching between different instances makes the Host Monitoring Connection Plugin useless in terms of instance health monitoring because the plugin will be unable to identify which instance it's connected to, and which one it's monitoring. This could result in false positive failure detections. At the same time, the plugin will still proactively monitor network connectivity to RDS Proxy endpoints and report outages back to a user application if they occur.
+
+# Host Monitoring Plugin v2
```
karenc-bq, line 76:
```
# **Experimental** Host Monitoring Plugin v2

> [!WARNING] This plugin is experimental and should be verified before using it in production environment.

This plugin is an alternative implementation of the enhanced failure monitoring feature and it is functionally equivalent to the Host Monitoring Plugin described above. Both plugins share the same set of [configuration parameters](#enhanced-failure-monitoring-parameters). The `efm2` plugin is designed to be a drop-in replacement for the `efm` plugin.

The `efm2` plugin is designed to address [some of the issues](https://github.com/awslabs/aws-advanced-jdbc-wrapper/issues/675) reported by multiple users. The following changes have been made:
- use weak pointers to ease garbage collection
- split monitoring logic into two separate threads to increase overall monitoring stability
- review locks for monitoring context
- redesign stopping of idle monitoring threads
- simplify monitoring logic

> [!NOTE] Since `efm` and `efm2` are separate plugins, it is technically possible to specify both plugins for a single connection. While this should not have any negative side effects, it is not recommended. It is recommended to use either the `efm` plugin, or the `efm2` plugin where it's needed.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitoringConnectionPlugin.java
```diff
@@ -0,0 +1,285 @@
...
+      final @NonNull PluginService pluginService, final @NonNull Properties properties) {
+    this(pluginService, properties, () -> new MonitorServiceImpl(pluginService), new RdsUtils());
+  }
+
+  HostMonitoringConnectionPlugin(
+      final @NonNull PluginService pluginService,
+      final @NonNull Properties properties,
+      final @NonNull Supplier<MonitorService> monitorServiceSupplier,
+      final RdsUtils rdsHelper) {
+    if (pluginService == null) {
```
karenc-bq, line 109: Do we need both the `@NonNull` annotation and the null checks

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorConnectionContext.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Monitoring context for each connection. This contains each connection's criteria for whether a
+ * server should be considered unhealthy. The context is shared between the main thread and the monitor thread.
+ */
+public class MonitorConnectionContext {
+
+  private final AtomicReference<WeakReference<Connection>> connectionToAbortRef;
```
karenc-bq, line 30: Thoughts on simplifying this to a volatile?
```suggestion
  `private volatile WeakReference<Connection> connectionToAbortRef;
```

https://stackoverflow.com/questions/281132/java-volatile-reference-vs-atomicreference?rq=4

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+    this.pluginService = pluginService;
+    this.telemetryFactory = pluginService.getTelemetryFactory();
+    this.hostSpec = hostSpec;
+    this.properties = properties;
+    this.failureDetectionTimeNano = TimeUnit.MILLISECONDS.toNanos(failureDetectionTimeMillis);
+    this.failureDetectionIntervalNano = TimeUnit.MILLISECONDS.toNanos(failureDetectionIntervalMillis);
+    this.failureDetectionCount = failureDetectionCount;
+    this.abortedConnectionsCounter = abortedConnectionsCounter;
+
+    final String nodeId = StringUtils.isNullOrEmpty(this.hostSpec.getHostId())
```
karenc-bq, line 113:
```suggestion
    final String hostId = StringUtils.isNullOrEmpty(this.hostSpec.getHostId())
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+        ? this.hostSpec.getHost()
```
karenc-bq, line 114: Does it make sense to have move this check inside getHostId so we don't have to do this every call?
```java
  public String getHostId() {
    return StringUtils.isNullOrEmpty(hostId) ? host : hostId;
  }
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+        () -> this.nodeUnhealthy ? 0L : 1L);
+
+    this.threadPool.submit(this::newContextRun); // task to handle new contexts
+    this.threadPool.submit(this); // task to handle active monitoring contexts
+    this.threadPool.shutdown(); // No more tasks are accepted by pool.
+  }
+
+  @Override
+  public boolean canDispose() {
+    return this.activeContexts.size() == 0;
```
karenc-bq, line 136:
```suggestion
    return this.activeContexts.isEmpty();
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+    }
+    LOGGER.finest(() -> Messages.get(
+        "MonitorImpl.stopped",
+        new Object[] {this.hostSpec.getHost()}));
+  }
+
+  protected long getActiveContextSize() {
+    return this.newContexts.values().stream()
+        .map(x -> (long) x.size())
+        .reduce(0L, Long::sum);
```
karenc-bq, line 155: You can also do this
```suggestion
    return this.newContexts.values().stream().mapToLong(Collection::size).sum();
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+    this.newContexts.clear();
+    this.activeContexts.clear();
+  }
+
+  // This method helps to organize unit tests.
+  long getCurrentTimeNano() {
+    return System.nanoTime();
+  }
+
+  public void newContextRun() {
```
karenc-bq, line 189:
```suggestion
  private void newContextRun() {
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+        ArrayList<Long> processedKeys = new ArrayList<>();
+        this.newContexts.entrySet().stream()
+            // Get entries with key (that is a time in nanos) less or equal than current time.
+            .filter(entry -> entry.getKey() < currentTimeNano)
+            .forEach(entry -> {
+              final Queue<WeakReference<MonitorConnectionContext>> queue = entry.getValue();
+              processedKeys.add(entry.getKey());
+              // Each value of found entry is a queue of monitoring contexts awaiting active monitoring.
+              // Add all contexts to a active monitoring contexts queue.
+              // Ignore contexts that already been disposed.
```
karenc-bq, line 209:
```suggestion
              // Add all contexts to an active monitoring contexts queue.
              // Ignore disposed contexts.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+    final TelemetryContext telemetryContext = telemetryFactory.openTelemetryContext(
+        "monitoring thread (new contexts)", TelemetryTraceLevel.TOP_LEVEL);
+    telemetryContext.setAttribute("url", this.hostSpec.getUrl());
+
+    try {
+      while (!this.stopped.get()) {
+
+        final long currentTimeNano = this.getCurrentTimeNano();
+
+        ArrayList<Long> processedKeys = new ArrayList<>();
```
karenc-bq, line 200:
```suggestion
        List<Long> processedKeys = new ArrayList<>();
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+              monitorContext.setInactive();
+              if (connectionToAbort != null) {
+                this.abortConnection(connectionToAbort);
+                this.abortedConnectionsCounter.inc();
+              }
+            }
+          }
+        } else {
+          // Remove inactive contexts from the queue
+          ArrayList<WeakReference<MonitorConnectionContext>> tmpActiveContexts = new ArrayList<>();
```
karenc-bq, line 279:
```suggestion
          final List<WeakReference<MonitorConnectionContext>> tmpActiveContexts = new ArrayList<>();
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+        if (this.nodeUnhealthy) {
+          // Kill all connections.
+          WeakReference<MonitorConnectionContext> monitorContextWeakRef;
+          while ((monitorContextWeakRef = this.activeContexts.poll()) != null) {
+            if (this.stopped.get()) {
+              break;
+            }
+            MonitorConnectionContext monitorContext = monitorContextWeakRef.get();
+            if (monitorContext != null) {
+              monitorContext.setNodeUnhealthy(true);
```
karenc-bq, line 268: Do we actually need to pass `true` to this method ? Will we ever explicitly call `ctx.setNodeUnhealthy(false)`?
```suggestion
              monitorContext.setNodeUnhealthy(this.nodeUnhealthy);
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+          continue;
+        }
+
+        final long statusCheckStartTimeNano = this.getCurrentTimeNano();
+        final boolean isValid = this.checkConnectionStatus();
+        final long statusCheckEndTimeNano = this.getCurrentTimeNano();
+
+        this.updateNodeHealthStatus(isValid, statusCheckStartTimeNano, statusCheckEndTimeNano);
+
+        if (this.nodeUnhealthy) {
```
karenc-bq, line 259: Thoughts on changing this if-else clause to
```java
        final List<WeakReference<MonitorConnectionContext>> tmpActiveContexts = new ArrayList<>();
        WeakReference<MonitorConnectionContext> monitorContextWeakRef;
        while ((monitorContextWeakRef = this.activeContexts.poll()) != null) {
          if (this.stopped.get()) {
            break;
          }
          MonitorConnectionContext monitorContext = monitorContextWeakRef.get();

          if (monitorContext == null) {
            continue;
          }

          if (this.nodeUnhealthy) {
            monitorContext.setNodeUnhealthy(true);
            final Connection connectionToAbort = monitorContext.getConnection();
            monitorContext.setInactive();
            if (connectionToAbort != null) {
              this.abortConnection(connectionToAbort);
              this.abortedConnectionsCounter.inc();
            }
          }

          if (!this.nodeUnhealthy && monitorContext.isActive()) {
            tmpActiveContexts.add(monitorContextWeakRef);
          }
        }

        this.activeContexts.addAll(tmpActiveContexts);
```
Less code duplication but not sure if it makes it harder to read.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -0,0 +1,418 @@
...
+
+    Queue<WeakReference<MonitorConnectionContext>> queue =
+        this.newContexts.computeIfAbsent(
+            startMonitoringTimeNano,
+            (key) -> new ConcurrentLinkedQueue<>());
+    queue.add(new WeakReference<>(context));
+  }
+
+  private long truncateNanoToSeconds(final long timeNano) {
+    return TimeUnit.SECONDS.toNanos(TimeUnit.NANOSECONDS.toSeconds(timeNano));
```
karenc-bq, line 176: Why are we converting nanos to seconds then back to nanos?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorConnectionContext.java
sergiyvamz: Volatile doesn't prevent multi-threading update collisions. It just makes a note to compiler to not optimize the final byte code and to expect the variable value may change at any time.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
sergiyvamz: This code is in class constructor and runs just once. It we had many places with similar place then it'd make sense to extract such logic into a new method.

.github/workflows/mysql_advanced_performance.yml
```diff
@@ -0,0 +1,55 @@
+name: Run Aurora Mysql Advanced Performance Tests
+
+on:
+  workflow_dispatch:
+
+jobs:
+  aurora-mysql-performance-tests:
+    concurrency: AvancedPerformanceTests-Aurora
```
brunos-bq, line 8: NIT: Advanced

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
brunos-bq: I am not sure "should be verified" makes much sense. Let's talk about "tested/

- This plugin is still in experimental stage. That means that users should test the plugin first for their use case scenarios before using it in their production environment.

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
brunos-bq, line 80: Host Monitoring Plugin v2, also known as `efm2` is an alternative, more modern implementation of the initial `efm` plugin for monitoring host health and detecting host failures faster.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -62,6 +62,7 @@ public class ConnectionPluginChainBuilder {
           put("logQuery", LogQueryConnectionPluginFactory.class);
           put("dataCache", DataCacheConnectionPluginFactory.class);
           put("efm", HostMonitoringConnectionPluginFactory.class);
+          put("efm2", software.amazon.jdbc.plugin.efm2.HostMonitoringConnectionPluginFactory.class);
```
brunos-bq, line 65: oooh, that looks a bit ugly. Users that would perhaps write custom code might get both plugins classes mixed up.

What about having a complete new name for the efm2 classes?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitoringConnectionPlugin.java
```diff
@@ -0,0 +1,285 @@
...
+
+    initMonitorService();
+
+    T result;
+    MonitorConnectionContext monitorContext = null;
+
+    try {
+      LOGGER.finest(
+          () -> Messages.get(
+              "HostMonitoringConnectionPlugin.activatedMonitoring",
```
brunos-bq, line 163: we need a clear logging if it's efm or efm2 - in case both are turned on

wrapper/src/main/java/software/amazon/jdbc/util/telemetry/OpenTelemetryFactory.java
```diff
@@ -57,11 +57,23 @@ public void postCopy(TelemetryContext telemetryContext, TelemetryTraceLevel trac
   }
 
   public TelemetryCounter createCounter(String name) {
+    if (name == null) {
+      throw new IllegalArgumentException("name");
+    }
+    if (name.length() > 63) {
+      name = name.substring(0, 63);
```
brunos-bq, line 64: do we have the full instance endpoint name in our efm counters? should we just use instance id instead?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+        // ignore
+        LOGGER.finest(
+            () -> Messages.get(
+                "MonitorConnectionContext.exceptionAbortingConnection",
+                new Object[] {sqlEx.getMessage()}));
+      }
+    } else {
+      context.setInactive();
+    }
+  }
```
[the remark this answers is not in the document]
sergiyvamz: I'd prefer to inactivate the context as soon as possible so it comes at the first line.

wrapper/src/test/java/integration/host/TestEnvironmentConfiguration.java
```diff
@@ -18,8 +18,10 @@
 
 public class TestEnvironmentConfiguration {
 
-  public boolean noDocker = Boolean.parseBoolean(System.getProperty("test-no-docker", "false"));
-  public boolean noAurora = Boolean.parseBoolean(System.getProperty("test-no-aurora", "false"));
+  public boolean noDocker =
+      Boolean.parseBoolean(System.getProperty("test-no-docker", "false"));
+  public boolean noAurora =
+      Boolean.parseBoolean(System.getProperty("test-no-aurora", "false"));
```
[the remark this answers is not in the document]
sergiyvamz: That's an expected behaviour. By default all tests are active.

karenc-bq (approved):
LGTM, some nits

wrapper/src/test/java/integration/host/TestEnvironmentConfiguration.java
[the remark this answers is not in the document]
karenc-bq: I don't think we need to set `false` as the default value if getProperty returns null, I think we can simplify the code, but it's just a nit

brunos-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
sergiyvamz: The idea is that efm2 to completely replace efm plugin. I hope such confusions will go away soon.

wrapper/src/main/java/software/amazon/jdbc/util/telemetry/OpenTelemetryFactory.java
sergiyvamz: Instance Id isn't available in some case so complete instance endpoint is used instead.

PR #765, opened 2023-12-04
Add example code for Read/Write Splitting sample

congoamz:

### DIFF
--- a/examples/ReadWriteSplittingSample/build.gradle.kts
+++ b/examples/ReadWriteSplittingSample/build.gradle.kts
@@ -1,0 +1,26 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+    implementation("org.postgresql:postgresql:42.6.0")
+    implementation("mysql:mysql-connector-java:8.0.33")
+    implementation("com.zaxxer:HikariCP:4.0.3")
+    implementation(project(":aws-advanced-jdbc-wrapper"))
+}
+
+tasks.withType<JavaExec> {
+    systemProperty("java.util.logging.config.file", "${project.buildDir}/resources/main/logging.properties")
+}
--- a/examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
+++ b/examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
@@ -185,9 +185,8 @@ public class ReadWriteSplittingSample {
 
     try (Statement stmt = conn.createStatement()) {
       for (int i = 0; i < NUM_READS; i++) {
-        stmt.addBatch(selectSQL);
+        stmt.execute(selectSQL);
       }
-      stmt.executeBatch();
     }
 
     long duration = System.nanoTime() - start;
--- a/examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
+++ b/examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon;
+
+import com.zaxxer.hikari.HikariConfig;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+import software.amazon.jdbc.ConnectionProviderManager;
+import software.amazon.jdbc.HikariPooledConnectionProvider;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
+
+public class ReadWriteSplittingSample {
+  // Adjust this value to run the sample with different driver configurations:
+  // 1: no r/w splitting
+  // 2: r/w splitting using r/w plugin
+  // 3: r/w plugin with internal connection pools
+  static final int APPROACH_ID = 1;
+  static final int NUM_THREADS = 250;
+  static final int NUM_WRITES = 2500;
+  static final int NUM_READS = 5000;
+  static final int MAX_SIMULTANEOUS = 40;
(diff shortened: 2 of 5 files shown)


examples/ReadWriteSplittingSample/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+/*
```
brunos-bq, line 1: do you need a new folder or a new package inside AWSDriverExample ?

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
+
+public class ReadWriteSplittingSample {
+  // Adjust this value to run the sample with different driver configurations:
+  // 1: no r/w splitting
+  // 2: r/w splitting using r/w plugin
+  // 3: r/w plugin with internal connection pools
+  static final int APPROACH_ID = 1;
```
brunos-bq, line 42: i think we should abandon the approach idea and have one sample file for each approach

settings.gradle.kts
```diff
@@ -26,7 +26,8 @@ include(
     "springwildfly",
     "springboothikariexample",
     "springtxfailover",
-    "vertxexample"
+    "vertxexample",
+    "read-write-sample"
```
brunos-bq, line 30: readwriteexample

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+import software.amazon.jdbc.ConnectionProviderManager;
+import software.amazon.jdbc.HikariPooledConnectionProvider;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
+
+public class ReadWriteSplittingSample {
```
brunos-bq, line 37: this class is too big - what about you make an Utils class with the database methods

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+    for (int i = 0; i < NUM_READS; i++) {
+      stmt.execute(selectSQL);
+    }
+
+    long duration = System.nanoTime() - start;
+    long durationMs = TimeUnit.NANOSECONDS.toMillis(duration);
+    LOGGER.finest(String.format("Thread %d read duration: %dms", tableNum, durationMs));
+  }
+
+  static class Approach1Thread implements Callable<Void> {
```
brunos-bq, line 199: NoRWSplittingThread

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+    } else if (APPROACH_ID == 3) {
+      props = poolProps;
+    } else {
+      throw new RuntimeException(
+          String.format(
+              "The approach ID should be set to a value between 1 and 3 (inclusive). Detected value: %d", APPROACH_ID));
+    }
+
+    DriverManager.registerDriver(new software.amazon.jdbc.Driver());
+    DriverManager.registerDriver(new org.postgresql.Driver());
```
karenc-bq, line 96: Why is this neccessary?

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+  // 2: r/w splitting using r/w plugin
+  // 3: r/w plugin with internal connection pools
+  static final int APPROACH_ID = 1;
+  static final int NUM_THREADS = 250;
+  static final int NUM_WRITES = 2500;
+  static final int NUM_READS = 5000;
+  static final int MAX_SIMULTANEOUS = 40;
+  static final int THREAD_DELAY_MS = 2000;
+  static final int EXECUTOR_TIMEOUT_MINS = 90;
+  static final boolean DELETE_TABLES_ON_STARTUP = false;
```
karenc-bq, line 49: Should this default to true instead

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,269 @@
...
+      }
+
+      long duration = System.nanoTime() - start;
+      LOGGER.info(String.format("Test completed in %dms", TimeUnit.NANOSECONDS.toMillis(duration)));
+    } catch (InterruptedException e) {
+      LOGGER.severe("The main thread was interrupted.");
+      throw new RuntimeException(e);
+    } finally {
+      try (Connection conn = DriverManager.getConnection(WRITER_CLUSTER, props)) {
+        Statement stmt = conn.createStatement();
```
karenc-bq, line 145: Can you add statement to the try clause as well? Same comment for the others.

examples/ReadWriteSplittingSample/build.gradle.kts
congoamz: Sorry, I'm not sure what you are suggesting, can you clarify? I believe the ReadWriteSplittingSample directory matches our other example directory structures, eg HikariExample. Are you suggesting I move this to the AWSDriverExample directory instead of having it in its own directory? My concerns with doing that are that AWSDriverExample contains single files and the logging config in ReadWriteSplittingExample is pretty specific to just ReadWriteSplittingSample.java

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
congoamz: That would simplify the file, but there would also be more code duplication and users would no longer be able to view all the code at once in a single file. Do we still want to go ahead with splitting this into different files?

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
congoamz: Similar to above, that would clean up this file and be more along the lines of recommended Java structure, but users would no longer be able to view all code at once in a single file. Let me know if you think we should make the change anyways though!

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
brunos-bq: Multiple files follow some arbitrary guidelines that I have seen since working on the wrapper. We tend to have multiple small files that handle one single case so users can run/copy them. If you believe the "approach" strategy matches better the way people will read it and the blog post, that's completely fine by me too.

Maybe once the blog post is released we can think whether we need smaller snippets for people to copy

examples/ReadWriteSplittingSample/src/main/java/software/amazon/ReadWriteSplittingSample.java
```diff
@@ -0,0 +1,271 @@
...
+  private static void executeReads(Connection conn, int tableNum) throws SQLException {
+    long start = System.nanoTime();
+    String selectSQL = "select * from rw_sample_?";
+
+    try (PreparedStatement stmt = conn.prepareStatement(selectSQL)) {
+      for (int i = 0; i < NUM_READS; i++) {
+        stmt.setInt(1, tableNum);
+        stmt.execute();
+      }
+    }
```
karenc-bq, line 194: I didnt try it out but I think you can do the following so you only execute once:
```suggestion
    try (PreparedStatement stmt = conn.prepareStatement(selectSQL)) {
      for (int i = 0; i < NUM_READS; i++) {
        stmt.setInt(1, tableNum);
        stmt.addToBatch();
      }
      stmt.executeBatch();
    }
```

congoamz: What's the advantage to doing it this way? If its just for performance, I wonder if we should make the change or not. I was using this code to purposely overload the writer so that we could observe how r/w splitting improved performance for this scenario. Then again, users probably won't be overloading their writer when they execute this code, since it requires a pretty specific configuration of settings in this file + rds cluster configuration. What do you think, should we go ahead with the change?

karenc-bq: It is a more common practice to use batch execute because of performance improvements. I just think that since it is a sample code we would want to follow best practices in our code. If we use batchExecute do we no longer see a performance improvement?

karenc-bq (approved):

PR #799, opened 2023-12-21
fix: restructuring try blocks in dialects for exception handling

brunos-bq:
### Summary

[](fix: handling exceptions when multi AZ isDialect() function fails)

### Additional Reviewers

@sergiyv-improving 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
@@ -54,34 +54,67 @@ public class AuroraPgDialect extends PgDialect {
       return false;
     }
 
+    Statement stmt = null;
+    ResultSet rs = null;
     boolean hasExtensions = false;
     boolean hasTopology = false;
     try {
-      try (final Statement stmt = connection.createStatement();
-           final ResultSet rs = stmt.executeQuery(extensionsSql)) {
-        if (rs.next()) {
-          final boolean auroraUtils = rs.getBoolean("aurora_stat_utils");
-          LOGGER.finest(() -> String.format("auroraUtils: %b", auroraUtils));
-          if (auroraUtils) {
-            hasExtensions = true;
-          }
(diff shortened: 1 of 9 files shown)


wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -40,10 +40,15 @@ public class RdsMultiAzDbClusterMysqlDialect extends MysqlDialect {
...
+           final ResultSet rs = stmt.executeQuery(TOPOLOGY_QUERY)) {
+        return rs.next();
+      } catch (final SQLException ex) {
+        // ignore
+      }
+      return false;
+    } catch (Exception ex) {
       // ignore
     }
     return false;
```
sergiyvamz, line 54:
```
Statement stmt;
ResultSet rs;
try {
	Statement stmt = connection.createStatement();
	ResultSet rs = stmt.executeQuery(TOPOLOGY_QUERY)
	return rs.next();
} catch (final SQLException ex) {
	// ignore
} finally {
	if (rs != null) {
		rs.close();
	}
	if (stmt != null) {
		stmt.close();
	}
}
return false;
```

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
```diff
@@ -54,34 +54,64 @@ public boolean isDialect(final Connection connection) {
...
-
+      if (rs != null) {
+        try {
+          rs.close();
+        } catch (SQLException ex) {
+          // ignore
+        }
+      }
+    }
+    try {
```
sergiyvamz, line 89: I think, as optimization, we can add the following code here:
```
if (!hasExtensions) {
return false;
}
```

sergiyvamz (approved):

karenc-bq (approved):

Issue #789, opened 2023-09-25
Relax overly-strict query type checking of "Results Producing" QueryReturnType, or provide override flag for batchUpdate()


benkanlu:
### Describe the feature

Currently the QueryInfo.getQueryReturnType checks to see if the SQL may return any results. If it does and the underlying method called (specifically jdbcTemplate.batchUpdate(sql)) is called, it throws an error saying it is not allowed. 

"Caused by: org.springframework.dao.DataIntegrityViolationException: StatementCallback; SQL [
-- ]; Statement.executeUpdate() or Statement.executeLargeUpdate() cannot issue statements that produce result sets."

### Use Case

There are many use cases to use 'CALL/EXECUTE' or even other things like 'OPTIMIZE' or 'REPAIR' in an automated way by sending multiple statements through batchUpdate(), which reduces communication overhead. If it does return anything (like in an OPTIMIZE table for example), the results can just be ignored or discarded.

Previously, MySQL/MariaDB native drivers allowed this type of 'ignoring' results sets if a non-DML statement (i.e. DDL/DCL) was sent. 

### Proposed Solution

While tightening down the types of statements allowed in a batchUpdate() may work in many cases, an option/flag to override this functionality would help immensely with current code bases expecting such functionality. 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS JDBC Driver for MySQL version used

Any version greater than 0.4.0 causes this issue.

### JDK version used

OpenJDK Runtime Environment Corretto-17.0.6.10.1 (build 17.0.6+10-LTS)

### Operating System and version

Windows 10 / Ubunutu 18.04

benkanlu:
Any additional thoughts on my last comment/explanation above?

crystall-bitquill:
Hi @benkanlu,

[...]

PR #836, opened 2024-01-18
fix: add update candidates for the maria db dialect; fix rds mysql dialect incorrectly returning false in isDialect method

crystall-bitquill:
### Summary

Add update candidates for the MariaDB dialect and fix the RDS MySQL dialect incorrectly returning false in the isDialect method

### Description

- Added update candidates for the MariaDB dialect in order to swap to MySQL dialects in the case of using a MySQL database with the protocol `jdbc:aws-wrapper:mariadb://`.
- Fixed the isDialect method of the RDS MySQL dialect. The method would return false if the MySQL dialect `isDialect` method also returned false. The MySQL isDialect method depends on whether the `version_comment` variable contains the value `mysql` but the RDS MySQL isDialect methods depends on whether the `version_comment` variable is equal to `Source distribution`.

Related to #789 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMysqlDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMysqlDialect.java
@@ -30,7 +30,17 @@ public class RdsMysqlDialect extends MysqlDialect {
 
   @Override
   public boolean isDialect(final Connection connection) {
-    if (!super.isDialect(connection)) {
+    if (super.isDialect(connection)) {
+      // MysqlDialect and RdsMysqlDialect use the same server version query to determine the dialect.
+      // The `SHOW VARIABLES LIKE 'version_comment'` either outputs
+      // | Variable_name   | value                        |
+      // |-----------------|------------------------------|
+      // | version_comment | MySQL Community Server (GPL) |
+      // for community Mysql, or
+      // | Variable_name   | value               |
+      // |-----------------|---------------------|
+      // | version_comment | Source distribution |
+      // for RDS MySQL. If super.idDialect returns true there is no need to check for RdsMysqlDialect.
       return false;
     }
     Statement stmt = null;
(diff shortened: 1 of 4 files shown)


wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMysqlDialect.java
```diff
@@ -30,7 +30,7 @@ public class RdsMysqlDialect extends MysqlDialect {
 
   @Override
   public boolean isDialect(final Connection connection) {
-    if (!super.isDialect(connection)) {
+    if (super.isDialect(connection)) {
```
sergiyvamz, line 33: Do we have tests for this condition? It doesn't look good.

wrapper/src/test/java/software/amazon/jdbc/DialectDetectionTests.java
```diff
@@ -0,0 +1,269 @@
...
+    closeable = MockitoAnnotations.openMocks(this);
+    when(this.mockConnection.createStatement()).thenReturn(this.mockStatement);
+    when(this.mockHost.getUrl()).thenReturn("url");
+    when(this.failResultSet.next()).thenReturn(false);
+    when(mockConnection.createStatement()).thenReturn(mockStatement);
+  }
+
+  @AfterEach
+  void cleanUp() throws Exception {
+    System.out.println("running clean up");
```
karenc-bq, line 85: Are these print statements supposed to be here

wrapper/src/test/java/software/amazon/jdbc/DialectDetectionTests.java
```diff
@@ -0,0 +1,269 @@
...
+    closeable.close();
+    Mockito.reset(
```
karenc-bq, line 87: It is a little bit confusing to see the code calling reset after closing the mocks. I thought close should be suffice?

karenc-bq (approved):

PR #851, opened 2024-01-24
fix: statement cancel not working with mysql driver

crystall-bitquill:
### Summary

Statement cancel not working with MySQL driver

### Description

The Statement.cancel method currently does not function with the MySQL community driver due to locking the ConnectionPluginManager. The MySQL driver also synchronizes on the Statement.getConnection method, which blocks any Statement.cancel executions. This PR moves the lock location and skips executing Statement.getConnection when running Statement.cancel.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -45,6 +44,7 @@ import software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin;
 import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsPlugin;
 import software.amazon.jdbc.plugin.strategy.fastestresponse.FastestResponseStrategyPlugin;
 import software.amazon.jdbc.profile.ConfigurationProfile;
+import software.amazon.jdbc.util.AsynchronousMethodsHelper;
 import software.amazon.jdbc.util.Messages;
 import software.amazon.jdbc.util.SqlMethodAnalyzer;
 import software.amazon.jdbc.util.WrapperUtils;
@@ -91,8 +91,6 @@ public class ConnectionPluginManager implements CanReleaseResources, Wrapper {
   private static final String NOTIFY_CONNECTION_CHANGED_METHOD = "notifyConnectionChanged";
   private static final String NOTIFY_NODE_LIST_CHANGED_METHOD = "notifyNodeListChanged";
   private static final SqlMethodAnalyzer sqlMethodAnalyzer = new SqlMethodAnalyzer();
-  private final ReentrantLock lock = new ReentrantLock();
-
   protected Properties props = new Properties();
   protected List<ConnectionPlugin> plugins;
   protected final @NonNull ConnectionProvider defaultConnProvider;
@@ -146,14 +144,6 @@ public class ConnectionPluginManager implements CanReleaseResources, Wrapper {
     this.plugins = plugins;
     this.connectionWrapper = connectionWrapper;
     this.telemetryFactory = telemetryFactory;
-  }
-
-  public void lock() {
-    lock.lock();
-  }
-
-  public void unlock() {
-    lock.unlock();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


wrapper/src/test/java/integration/container/tests/MysqlTests.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@EnableOnDatabaseEngine({DatabaseEngine.MYSQL})
+@ExtendWith(TestDriverProvider.class)
+public class MysqlTests {
+
+  private static final Logger LOGGER = Logger.getLogger(MysqlTests.class.getName());
+
+  @TestTemplate
+  void testCancelStatement(TestDriver testDriver) {
```
sergiyvamz, line 46: I think we can test it with unit tests, not integration tests.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -305,7 +296,14 @@ public <T, E extends Exception> T execute(
       final Object[] jdbcMethodArgs)
       throws E {
 
-    final Connection conn = WrapperUtils.getConnectionFromSqlObject(methodInvokeOn);
+    final Connection conn;
+    if (AsynchronousMethodsHelper.ASYNCHRONOUS_METHODS.contains(methodName)) {
+      // The target driver may block on Statement.getConnection().
+      conn = this.pluginService.getCurrentConnection();
```
sergiyvamz, line 302: I believe we need to get a connection from the Statement object and compare it to a current connection that is stored in PluginService.

sergiyvamz: How target driver may block on Statement.getConnection()?

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -45,6 +45,7 @@
 import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsPlugin;
 import software.amazon.jdbc.plugin.strategy.fastestresponse.FastestResponseStrategyPlugin;
 import software.amazon.jdbc.profile.ConfigurationProfile;
+import software.amazon.jdbc.util.AsynchronousMethodsHelper;
```
karenc-bq, line 48: Can you also remove the unused import in this file

wrapper/src/test/java/software/amazon/jdbc/util/WrapperUtilsTest.java
```diff
@@ -144,6 +151,30 @@ void testExecutesWithPluginsWithExceptionIsSequential() {
...
+  void testCancelStatementIsNotBlockedExecute() {
+    List<CompletableFuture<Integer>> futures = new ArrayList<>();
+
+    futures.add(CompletableFuture.supplyAsync(this::callExecuteWithPlugins));
+    futures.add(CompletableFuture.supplyAsync(this::callCancelExecuteWithPlugins));
+
+    for (CompletableFuture<Integer> future : futures) {
+      future.join();
+    }
+  }
```
karenc-bq, line 164: If you don't care about the results just do
```suggestion
    CompletableFuture.allOf(
        CompletableFuture.supplyAsync(this::callExecuteWithPluginsWithException),
        CompletableFuture.supplyAsync(this::callCancelExecuteWithPluginsWithException)
    ).join();
```

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
```diff
@@ -251,7 +260,9 @@ public static <T, E extends Exception> T executeWithPlugins(
       }
 
     } finally {
-      pluginManager.unlock();
+      if (lock.isHeldByCurrentThread()) {
```
karenc-bq, line 263: I wonder if the unlock should happen after calling closeContext. Can you check with @sergiyvamz

wrapper/src/main/java/software/amazon/jdbc/util/AsynchronousMethodsHelper.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package software.amazon.jdbc.util;
+
+import java.util.Collections;
+import java.util.List;
+
+public class AsynchronousMethodsHelper {
+  public static final List<String> ASYNCHRONOUS_METHODS = Collections.singletonList(
+      "Statement.cancel"
```
karenc-bq, line 24: What about other variations of statement.cancel? e.g. PreparedStatement.cancel

crystall-bitquill: Good point, it's been added to the list

PR #854, opened 2024-01-26
refactor: reduce regex checks to improve performance

karenc-bq:
### Summary

Reduce number of calls to Matcher.find to improve performance.

### Description

Baseline performance:
image
Identifying the cluster type takes longer than executing statements.
After the refactoring:
image

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
@@ -18,6 +18,7 @@ package software.amazon.jdbc.util;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
```diff
@@ -148,56 +111,53 @@ public class RdsUtils {
           "^(([0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})?)"
               + "::(([0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})?)$");
 
+  private static final Map<String, Matcher> cachedPatterns = new HashMap<>();
```
sergiyvamz, line 114: They should be concurrent / threadsafe.

joyc-bq (approved):

sergiyvamz (approved):

PR #856, opened 2024-01-29
feat: update HostSelector implementations to account for HostAvailablity

aaronchung-bitquill:
### Summary

feat: update HostSelector implementations to account for HostAvailablity

### Description

For each implementation of HostSelector, it filters and creates a list of eligible hosts. Added an extra condition to only filter out ones that are `HostAvailability.NOT_AVAILABLE`

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/LeastConnectionsHostSelector.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/LeastConnectionsHostSelector.java
@@ -24,6 +24,7 @@ import java.util.Properties;
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.hostavailability.HostAvailability;
 import software.amazon.jdbc.util.Messages;
 import software.amazon.jdbc.util.SlidingExpirationCache;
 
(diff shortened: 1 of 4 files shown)


sergiyvamz (approved):

wrapper/src/test/java/RandomHostSelectorTests.java
```diff
@@ -0,0 +1,57 @@
...
+import software.amazon.jdbc.RandomHostSelector;
+import software.amazon.jdbc.hostavailability.HostAvailability;
+import software.amazon.jdbc.hostavailability.SimpleHostAvailabilityStrategy;
+
+class RandomHostSelectorTests {
+
+
+
+  @RepeatedTest(value = 50)
+  void testGetHostGivenUnavailbleHost() throws SQLException {
```
brunos-bq, line 35: since you have one function that select multiple eligible hosts, can you add one test where you have 2 unavailable and 2 available hosts and check that your return is what is expected?

wrapper/src/main/java/software/amazon/jdbc/LeastConnectionsHostSelector.java
```diff
@@ -42,7 +43,8 @@ public HostSpec getHost(
       @NonNull final HostRole role,
       @Nullable final Properties props) throws SQLException {
     final List<HostSpec> eligibleHosts = hosts.stream()
-        .filter(hostSpec -> role.equals(hostSpec.getRole()))
+        .filter(hostSpec ->
```
brunos-bq, line 46: NIT: add a comment above this level explaining the filter clause

brunos-bq (approved):

PR #863, opened 2024-01-29
fix: redacted credentials in saml assertion unit test

joyc-bq:


wrapper/src/test/java/software/amazon/jdbc/plugin/federatedauth/AdfsCredentialsProviderFactoryTest.java
```diff
@@ -98,7 +98,11 @@ public void init() throws IOException {
 
   @Test
   void test() throws IOException, SQLException {
-    this.adfsCredentialsProviderFactory.getSamlAssertion(props);
+    String correctSamlAssertion = IOUtils.toString(
```
aaronchung-bitquill, line 101: Can you add `final` to these variables.  And as the original author of this test, I humbly to ask if you can add it to the variables i missed, as well? 😅

aaronchung-bitquill (approved):

PR #879, opened 2024-02-09
chore: target driver dialect in closed connections

jasonli-improving:
### Summary

Add check for target dialect specific method calls allowed on closed connections

### Description

Used Hashset to store allowed methods. Change to check dialect specific methods instead of generic ones

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
@@ -417,9 +417,6 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
    */
   private boolean allowedOnClosedConnection(final String methodName) {
     TargetDriverDialect dialect = this.pluginService.getTargetDriverDialect();
-    if (dialect == null) {
-      return false;
-    }
     return dialect.getAllowedOnConnectionMethodNames().contains(methodName);
   }
 
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
@@ -42,6 +42,7 @@ import software.amazon.jdbc.PropertyDefinition;
 import software.amazon.jdbc.hostavailability.HostAvailability;
 import software.amazon.jdbc.plugin.AbstractConnectionPlugin;
 import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsHelper;
+import software.amazon.jdbc.targetdriverdialect.TargetDriverDialect;
 import software.amazon.jdbc.util.Messages;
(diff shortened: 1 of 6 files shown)


davecramer:
I don't understand why the change here? It appears that you are now calling `getAllowedOnConnectionMethodNames` which seems to be many of them?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
```diff
@@ -34,6 +37,70 @@ public class GenericTargetDriverDialect implements TargetDriverDialect {
 
   private static final Logger LOGGER =
       Logger.getLogger(GenericTargetDriverDialect.class.getName());
+  public static final String CONN_GET_AUTO_COMMIT = "Connection.getAutoCommit";
```
karenc-bq, line 40: Can you follow the pattern used for `NETWORK_BOUND_METHODS`? Instead of listing all of the methods one by one they are stored directly in a list `List<String> NETWORK_BOUND_METHODS`.

karenc-bq: This list should only contain the common methods that are required by all target driver dialects

jasonli-improving:
> I don't understand why the change here? It appears that you are now calling `getAllowedOnConnectionMethodNames` which seems to be many of them?

There are currently no checks for specific target driver dialect methods in the `allowedOnClosedConnection` method. These changes account for the differences in allowed methods between the different target driver dialects.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/GenericTargetDriverDialect.java
jasonli-improving: Added this. There aren't too many methods that are common between all of them though

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -419,11 +416,11 @@ protected void updateTopology(final boolean forceUpdate) throws SQLException {
    * @return true if the given method is allowed on closed connections
    */
   private boolean allowedOnClosedConnection(final String methodName) {
-    // TODO: consider to use target driver dialect
-    return methodName.equals(METHOD_GET_AUTO_COMMIT)
-        || methodName.equals(METHOD_GET_CATALOG)
-        || methodName.equals(METHOD_GET_SCHEMA)
-        || methodName.equals(METHOD_GET_TRANSACTION_ISOLATION);
+    TargetDriverDialect dialect = this.pluginService.getTargetDriverDialect();
+    if (dialect == null) {
```
sergiyvamz, line 420: PluginService gurantees non null dialect object.

jasonli-improving: removed null check

PR #910, opened 2024-03-01
chore: add new integration tests to ensure topology query doesn't change

jasonli-improving:
### Summary

Add integration tests to check for timestamp and types in the return values of topology queries

### Description

- Add function to update cluster certification
- Add integration tests to test Aurora topology query types/timestamps
- Add RdsMultiAz type test (disabled due to not being supported yet)


wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
```diff
@@ -26,7 +26,7 @@
 
 public class AuroraMysqlDialect extends MysqlDialect {
 
-  private static final String TOPOLOGY_QUERY =
+  public static final String TOPOLOGY_QUERY =
```
brunos-bq, line 29: what are the pros/cons of keeping the variable private but adding a getter function to the dialect to retrieve the topology query?

jasonli-improving: I think adding a getter would be better since we don't want to change the topology query, I can change that

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
```diff
@@ -83,4 +83,8 @@ public HostListProviderSupplier getHostListProvider() {
         NODE_ID_QUERY,
         IS_READER_QUERY);
   }
+
+  public static String getTopologyQuery() {
```
brunos-bq, line 87: you could check if the superclasses also need that function and define it there or just the subclasses

jasonli-improving: Looks like the `TOPOLOGY_QUERY` field isn't defined in the superclass, just the subclasses

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -923,4 +924,20 @@ public void addAuroraAwsIamUser(
...
+      String dbInstanceIdentifier, String certificateIdentifier) {
+    try {
+      ModifyDbInstanceRequest modifyDbInstanceRequest = ModifyDbInstanceRequest.builder()
+          .dbInstanceIdentifier(dbInstanceIdentifier)
+          .publiclyAccessible(true)
+          .applyImmediately(true)
+          .caCertificateIdentifier(certificateIdentifier)
+          .build();
+      rdsClient.modifyDBInstance(modifyDbInstanceRequest);
+    } catch (Exception e) {
```
crystall-bitquill, line 938: If this exception occurs, should the test fail? Does it need to be re thrown?

jasonlamz: It should now re throw and fail the test for an exception

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraMysqlDialect.java
sergiyvamz, line 87: What is the reason to make this method static?

wrapper/src/main/java/software/amazon/jdbc/dialect/AuroraPgDialect.java
```diff
@@ -127,4 +127,8 @@ public HostListProviderSupplier getHostListProvider() {
         NODE_ID_QUERY,
         IS_READER_QUERY);
   }
+
+  public static String getTopologyQuery() {
```
sergiyvamz, line 131: What is the reason to make this method static?

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsMultiAzDbClusterMysqlDialect.java
```diff
@@ -109,4 +109,8 @@ public void prepareConnectProperties(
             ? connectionAttributes
             : connectProperties.getProperty("connectionAttributes") + "," + connectionAttributes);
   }
+
+  public static String getTopologyQuery() {
```
sergiyvamz, line 113: What is the reason to make this method static?

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,273 @@
...
+                .getInfo()
+                .getDatabaseInfo()
+                .getInstances()
+                .get(0)
+                .getPort(),
+            TestEnvironment.getCurrent().getInfo().getDatabaseInfo().getDefaultDbName());
+    LOGGER.finest("Connecting to " + url);
+
+    String query = null;
+    if (TestEnvironment.getCurrent().getCurrentDriver() == TestDriver.PG) {
```
sergiyvamz, line 101: What's about MariaDb? Is it supported by the test?

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,273 @@
...
+    String dbInstanceIdentifier = TestEnvironment.getCurrent().getInfo()
+        .getDatabaseInfo()
+        .getInstances()
+        .get(0)
+        .getInstanceId();
+    Region region = Region.of(TestEnvironment.getCurrent()
+        .getInfo()
+        .getAuroraRegion());
+
+    RdsClient client = RdsClient.builder()
```
sergiyvamz, line 174: It's better to move this logic to aurora utility class.

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,273 @@
...
+    final Connection conn = DriverManager.getConnection(url, props);
+    assertTrue(conn.isValid(5));
+    Statement stmt = conn.createStatement();
+    stmt.executeQuery(query);
+    ResultSet rs = stmt.getResultSet();
+
+    Date date = null;
+    while (rs.next()) {
+      if (rs.getString(5) != null) {
+        date = format.parse(rs.getString(5));
```
sergiyvamz, line 203: We shouldn't rely on parsing timestamp. Parsing is locale dependent and the test might fail if database (or session) default settings changes.

Is it possible to call rs.getTime() or rs.getTimestamp()?
Is it possible to get metadata rs.getMetadata() and then check column type?

jasonlamz: The first test already checks all the column types of the query, including the timestamp. Should this test be removed then?

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
jasonlamz: It's not

sergiyvamz: I mean I don't see a branch of the code that handle the case with mariaDb. I see AuroraMysql and AuroraPG.

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,275 @@
...
+    } else {
+      AuroraMysqlDialect dialect = new AuroraMysqlDialect();
+      query = dialect.getTopologyQuery();
+      format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSSSSS");
+    }
+
+    final Connection conn = DriverManager.getConnection(url, props);
+    assertTrue(conn.isValid(5));
+    Statement stmt = conn.createStatement();
+    stmt.executeQuery(query);
```
karenc-bq, line 192: Can you use try with resources here? https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

karenc-bq: Same comment for the other locations.

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,275 @@
...
+  @ExtendWith(TestDriverProvider.class)
+  public void testAuroraTypes(TestDriver testDriver) throws SQLException {
+    LOGGER.info(testDriver.toString());
+    List<String> expectedTypes;
+
+    final Properties props = ConnectionStringHelper.getDefaultPropertiesWithNoPlugins();
+    DriverHelper.setConnectTimeout(testDriver, props, 10, TimeUnit.SECONDS);
+    DriverHelper.setSocketTimeout(testDriver, props, 10, TimeUnit.SECONDS);
+
+    String url =
```
karenc-bq, line 82: Can you please add the `final` keyword to all your constant local variables?
Also you can just call `ConnectionStringHelper.getWrapperUrl()` here.

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,275 @@
...
+                .getDatabaseInfo()
+                .getInstances()
+                .get(0)
+                .getPort(),
+            TestEnvironment.getCurrent().getInfo().getDatabaseInfo().getDefaultDbName());
+
+    if (TestEnvironment.getCurrent().getCurrentDriver() == TestDriver.MARIADB
+        && TestEnvironment.getCurrent().getInfo().getRequest().getDatabaseEngine() == DatabaseEngine.MYSQL) {
+      props.setProperty("permitMysqlScheme", "1");
+    }
```
karenc-bq, line 102: `getWrapperUrl` should already be appending `permitMysqlScheme` to the wrapper url, why is this if-clause required?

wrapper/src/test/java/integration/container/tests/TopologyQueryTests.java
```diff
@@ -0,0 +1,275 @@
...
+
+    LOGGER.finest("Connecting to " + url);
+
+    if (TestEnvironment.getCurrent().getCurrentDriver() == TestDriver.PG) {
+      AuroraPgDialect dialect = new AuroraPgDialect();
```
karenc-bq, line 107: I think you could move this if-else to the beginning of the test after `List<String> expectedTypes;`, before establishing a connection, since you don't need any connections here and you are just setting up for the test.

PR #919, opened 2024-03-08
 chore: automate aurora db engine updates in GH Actions

jasonlamz:
### Summary

Automate Aurora integration tests to use both latest and LTS versions

### Description

- Modify integration tests to use LTS and latest versions
- Allow user specified system property to override versions
- Use SDK to get db engine versions


.github/workflows/run-integration-tests.yml
```diff
@@ -11,8 +11,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  all-integration-tests:
-    name: 'Run Aurora integration tests'
+  all-integration-tests-latest:
```
sergiyvamz, line 14: Integration may run for hours and sometimes it reach the max allowed run time of 6h. Increasing  (doubling) amount of work for this github action will definitely increase the chance of such timeout out event. I'd suggest to split this script into two separate github actions.

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -471,15 +473,38 @@ private static String getAuroraDbEngine(TestEnvironmentRequest request) {
...
     }
+    return findAuroraDbEngineVersion(env, engineName, systemPropertyVersion);
+  }
+
+  private static String findAuroraDbEngineVersion(
+      TestEnvironment env,
+      String engineName,
+      String systemPropertyVersion
+  ) {
+    if (systemPropertyVersion == null || systemPropertyVersion.equals("lts")) {
```
sergiyvamz, line 501: Can we use switch-case syntax for readability?

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -923,4 +927,26 @@ public void addAuroraAwsIamUser(
...
+        DescribeDbEngineVersionsRequest.builder().engine(engine).build()
+    );
+    for (DBEngineVersion version : versions.dbEngineVersions()) {
+      res.add(version.engineVersion());
+    }
+    return res;
+  }
+
+  public String getLatestVersion(String engine) {
+    return Collections.max(getEngineVersions(engine));
```
sergiyvamz, line 943: Why "max"? Maybe collection.stream().sorted().last() ?

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -923,4 +927,26 @@ public void addAuroraAwsIamUser(
...
+  }
+
+  public String getLTSVersion(String engine) {
+    final DescribeDbEngineVersionsResponse versions = rdsClient.describeDBEngineVersions(
+        DescribeDbEngineVersionsRequest.builder().defaultOnly(true).engine(engine).build()
+    );
+    return versions.dbEngineVersions().get(0).engineVersion();
```
sergiyvamz, line 950: Can we check if collection isn't empty?

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -471,15 +473,38 @@ private static String getAuroraDbEngine(TestEnvironmentRequest request) {
     }
   }
 
-  private static String getAuroraDbEngineVersion(TestEnvironmentRequest request) {
+  private static String getAuroraDbEngineVersion(
+      TestEnvironment env,
+      TestEnvironmentRequest request) {
+    String engineName;
+    String systemPropertyVersion;
     switch (request.getDatabaseEngine()) {
```
sergiyvamz, line 481: We need support of TAZ clusters either

wrapper/src/test/java/integration/host/TestEnvironmentConfiguration.java
```diff
@@ -78,6 +78,10 @@ public class TestEnvironmentConfiguration {
   public String auroraClusterName = System.getenv("AURORA_CLUSTER_NAME"); // "cluster-mysql"
   public String auroraClusterDomain =
       System.getenv("AURORA_CLUSTER_DOMAIN"); // "XYZ.us-west-2.rds.amazonaws.com"
+  public String auroraMySqlDbEngineVersion =
```
sergiyvamz, line 81: Can we add a comment with a possible/expected values?

.github/workflows/run-integration-tests.yml
karenc-bq: Adding to what @sergiyvamz said, another way to run multiple jobs in parallel without duplicating code is to use matrix https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -471,15 +473,41 @@ private static String getAuroraDbEngine(TestEnvironmentRequest request) {
     }
   }
 
-  private static String getAuroraDbEngineVersion(TestEnvironmentRequest request) {
+  private static String getAuroraDbEngineVersion(
+      TestEnvironment env,
```
sergiyvamz, line 477: "env" contains all that the method needs. I think a single parameter would be enough.

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -471,15 +473,41 @@ private static String getAuroraDbEngine(TestEnvironmentRequest request) {
...
+  }
+
+  private static String findAuroraDbEngineVersion(
+      TestEnvironment env,
+      String engineName,
+      String systemPropertyVersion) {
+    if (systemPropertyVersion == null) {
+      return env.auroraUtil.getLTSVersion(engineName);
+    }
+    switch (systemPropertyVersion) {
```
sergiyvamz, line 503: systemPropertyVersion.toLower()

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -923,4 +928,33 @@ public void addAuroraAwsIamUser(
...
+      res.add(version.engineVersion());
+    }
+    return res;
+  }
+
+  public String getLatestVersion(String engine) {
+    return getEngineVersions(engine)
+        .stream()
+        .sorted()
+        .reduce((first, last) -> last)
```
sergiyvamz, line 947: Can we use sorted() to sort collection and then return first() or last()?
https://www.baeldung.com/java-8-sort-lambda

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -390,9 +389,12 @@ private static void createAuroraDbCluster(TestEnvironment env, int numOfInstance
 
       try {
         String engine = getAuroraDbEngine(env.info.getRequest());
-        String engineVersion = getAuroraDbEngineVersion(env.info.getRequest());
+        String engineVersion = getAuroraDbEngineVersion(env);
```
sergiyvamz, line 392: getAuroraDbEngineVersion() may return null. We need to check it.

sergiyvamz (approved):

PR #927, opened 2024-03-12
chore: gradle upgrade wrapper

jasonlamz:
### Summary

Upgrade Gradle version to 7.6.4

### Description

- Add check to install `findutils` for graalvm
- Removed failing lines in Gradle build


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.4-bin.zip
```
sergiyvamz, line 3: Why do we choose version 7.6.4 and not 8.6?

wrapper/src/test/java/integration/util/ContainerHelper.java
```diff
@@ -222,6 +222,9 @@ public T withFixedExposedPort(int hostPort, int containerPort) {
                     builder
                         .from(testContainerImageName)
                         .run("mkdir", "app")
+                        .run(testContainerImageName.contains("graalvm")
+                            ? "microdnf install findutils"
```
sergiyvamz, line 226: Why do we need this findutils?

gradle/wrapper/gradle-wrapper.properties
jasonlamz: There are a lot of changes when upgrading from version 7 to 8. Should it be 8.6 instead?
https://docs.gradle.org/current/userguide/upgrading_version_7.html#changes_8.0

wrapper/src/test/java/integration/util/ContainerHelper.java
jasonlamz: The GraalVM tests were failing with a `xargs is not available` error. Adding this package fixes that.

Issue #678, opened 2023-10-11
Unable to connect to Aurora MySQL reader with blue/green deployed cluster


markusrydh:
### Describe the bug

We are using this driver together with MariaDB driver to connect to an AWS Aurora MySQL 5.7 cluster using readWriteSplitting with an internal HikariCP connection pool. It works fine if the cluster is not deployed with blue/green deployment, but if it was deployed using blue/green deployment then connections to the reader fail due to unknown host. The host name that the driver tries to connect to contains "green" or "blue", but no such host name can be resolved.

Note that the actual blue/green deployment took place a long time ago, and no switch is performed related to the time of this issue.

### Expected Behavior

We expect the readWriteSplitting plugin to use hostnames for the reader replicas that can actually be resolved, so that the internal connection pools for read-only connections are populated.

### What plugins are used? What other connection properties were set?

Plugins: readWriteSplitting,failover,efm

### Current Behavior

The driver seems to use data from the @@aurora_server_id to generate fqdn to connect to, but since the instances seem to retain server-ids from the blue/green deployment the fqdn is not valid and it fails to connect.

```
2023-10-11 09:05:04.643  INFO 1 --- [ dbThreadPool-1] com.zaxxer.hikari.HikariDataSource       : hikari-db-readonly-pool - Starting...

2023-10-11 09:05:05.742 ERROR 1 --- [ dbThreadPool-1] com.zaxxer.hikari.pool.HikariPool        : hikari-db-readonly-pool - Exception during pool initialization.

java.sql.SQLNonTransientConnectionException: Socket fail to connect to host:address=(host=production-main-1-green-zymnmc.cxl<redacted>.rds.amazonaws.com)(port=3306)(type=primary). production-main-1-green-zymnmc.cxl<redacted>.rds.amazonaws.com
	at org.mariadb.jdbc.client.impl.ConnectionHelper.connectSocket(ConnectionHelper.java:137)
	at org.mariadb.jdbc.client.impl.StandardClient.<init>(StandardClient.java:99)
	at org.mariadb.jdbc.Driver.connect(Driver.java:70)
	at org.mariadb.jdbc.Driver.connect(Driver.java:101)
	at org.mariadb.jdbc.Driver.connect(Driver.java:27)
	at com.zaxxer.hikari.util.DriverDataSource.getConnection(DriverDataSource.java:138)
	at com.zaxxer.hikari.pool.PoolBase.newConnection(PoolBase.java:364)
[...]

PR #948, opened 2024-04-01
green node endpoint replacement

sergiyvamz:
### Summary

green node endpoint replacement

### Description

Allows the driver to to replace `Green` DNS endpoints when they get unavailable after Blue/Green switchover is completed.

### Additional Reviewers

@karenc-bq 
@davecramer 


docs/using-the-jdbc-driver/using-plugins/UsingTheFailoverPlugin.md
```diff
@@ -104,3 +104,8 @@ public class HikariCPSQLException implements SQLExceptionOverride {
     }
 }
 ```
+
+### Blue/Green Deployments
+Although AWS Advanced JDBC Driver is not compatible with [AWS Blue/Green Deployments](https://docs.aws.amazon.com/whitepapers/latest/overview-deployment-options/bluegreen-deployments.html) and officially doesn't support it, the Failover Plugin has been validated for clusters with Blue/Green Deployments. While general basic connectivity to `Blue` and `Green` clusters
+are always in place, some failover cases are not fully supported. The team is aware of these cases and working on them to bring
+AWS Blue/Green Deployments support at full scale. Meanwhile, the users may consider using configuration parameter [`enableGreenNodeReplacement`](../UsingTheJdbcDriver.md) that allows the driver to replace `Green` DNS endpoints when they get unavailable after Blue/Green switchover is completed.
```
davecramer, line 111: s/when they get unavailable'/'when they become unavailable'

wrapper/src/main/java/software/amazon/jdbc/DriverConnectionProvider.java
```diff
@@ -117,12 +122,69 @@ public Connection connect(
...
+      final String fixedHost = this.rdsUtils.removeGreenInstancePrefix(hostSpec.getHost());
+      final HostSpec connectionHostSpec = new HostSpecBuilder(hostSpec.getHostAvailabilityStrategy())
+          .copyFrom(hostSpec)
+          .host(fixedHost)
+          .build();
+
+      final ConnectInfo fixedConnectInfo = targetDriverDialect.prepareConnectInfo(protocol, connectionHostSpec, copy);
+
+      LOGGER.finest(() -> "Connecting (with fixed host) to " + fixedConnectInfo.url
+          + PropertyUtils.logProperties(PropertyUtils.maskProperties(fixedConnectInfo.props), "\nwith properties: \n"));
```
davecramer, line 184: This logger message won't make much sense outside of the context of this code.
Something like "Connecting to" + fixedConnectionInfo.url + " after correcting the hostname from" + oldHostName

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -156,6 +156,16 @@ public class PropertyDefinition {
               "true", "false"
           });
 
+  public static final AwsWrapperProperty ENABLE_GREEN_NODE_REPLACEMENT =
+      new AwsWrapperProperty(
+          "enableGreenNodeReplacement",
+          "false",
+          "Enables to replace a green node host names with original host name when green host DNS doesn't exist.",
```
davecramer, line 163: Enables replacing a green node host name with the original hostname after a blue/green switchover and the green name no longer resolves

davecramer (approved):

wrapper/src/main/java/software/amazon/jdbc/DataSourceConnectionProvider.java
```diff
@@ -157,6 +163,73 @@ public Connection connect(
...
+      final @NonNull HostSpec hostSpec,
+      final @NonNull Properties props)
+      throws SQLException {
+
+    try {
+      return ds.getConnection();
+
+    } catch (Throwable throwable) {
+
+      if (!PropertyDefinition.ENABLE_GREEN_NODE_REPLACEMENT.getBoolean(props)) {
```
karenc-bq, line 179:
```suggestion
      return ds.getConnection();
    } catch (Throwable throwable) {
      if (!PropertyDefinition.ENABLE_GREEN_NODE_REPLACEMENT.getBoolean(props)) {
```

wrapper/src/main/java/software/amazon/jdbc/DataSourceConnectionProvider.java
```diff
@@ -157,6 +163,73 @@ public Connection connect(
...
+        loopThrowable = loopThrowable.getCause();
+      }
+
+      if (unknownHostException == null) {
+        throw throwable;
+      }
+
+      if (!this.rdsUtils.isRdsDns(hostSpec.getHost()) || !this.rdsUtils.isGreenInstance(hostSpec.getHost())) {
+        throw throwable;
+      }
```
karenc-bq, line 200: Does it make sense to check the dns first and fail early before going through potentially 100 exceptions?

wrapper/src/main/java/software/amazon/jdbc/DataSourceConnectionProvider.java
```diff
@@ -157,6 +163,73 @@ public Connection connect(
...
+      } catch (UnknownHostException tmp) {
+        // do nothing
+      }
+
+      if (resolvedAddress != null) {
+        // Green node DNS exists
+        throw throwable;
+      }
+
+      // Green node DNS doesn't exist. Try to replace it with corresponding node name and connect again.
```
karenc-bq, line 215:
```suggestion
      // Green node DNS doesn't exist. Try to replace it with the corresponding node name and connect again.
```

wrapper/src/main/java/software/amazon/jdbc/DataSourceConnectionProvider.java
```diff
@@ -157,6 +163,73 @@ public Connection connect(
...
+      // check DNS for such green host name
+      InetAddress resolvedAddress = null;
+      try {
+        resolvedAddress = InetAddress.getByName(hostSpec.getHost());
+      } catch (UnknownHostException tmp) {
+        // do nothing
+      }
+
+      if (resolvedAddress != null) {
+        // Green node DNS exists
```
karenc-bq, line 211:
```suggestion
        // Green host DNS exists
```

wrapper/src/main/java/software/amazon/jdbc/DriverConnectionProvider.java
```diff
@@ -117,12 +122,71 @@ public Connection connect(
...
         + PropertyUtils.logProperties(PropertyUtils.maskProperties(connectInfo.props), "\nwith properties: \n"));
 
-    Connection conn = this.driver.connect(connectInfo.url, connectInfo.props);
+    Connection conn;
+    try {
+      conn = this.driver.connect(connectInfo.url, connectInfo.props);
+
+    } catch (Throwable throwable) {
+
+      if (!PropertyDefinition.ENABLE_GREEN_NODE_REPLACEMENT.getBoolean(props)) {
```
karenc-bq, line 137: What do you think of moving the duplicated code into a util class or helper method?
```java
  public static HostSpec replaceHost(Throwable throwable, RdsUtils rdsUtils, HostSpec hostSpec, Properties props) throws Throwable {
    if (!PropertyDefinition.ENABLE_GREEN_NODE_REPLACEMENT.getBoolean(props)) {
      throw throwable;
    }

    UnknownHostException unknownHostException = null;
    int maxDepth = 100;
    Throwable loopThrowable = throwable;
    while (--maxDepth > 0 && loopThrowable != null) {
      if (loopThrowable instanceof UnknownHostException) {
        unknownHostException = (UnknownHostException) loopThrowable;
        break;
      }
      loopThrowable = loopThrowable.getCause();
    }

    if (unknownHostException == null) {
      throw throwable;
    }

    if (!rdsUtils.isRdsDns(hostSpec.getHost()) || !this.rdsUtils.isGreenInstance(hostSpec.getHost())) {
      throw throwable;
    }

    // check DNS for such green host name
    InetAddress resolvedAddress = null;
    try {
      resolvedAddress = InetAddress.getByName(hostSpec.getHost());
    } catch (UnknownHostException tmp) {
      // do nothing
    }

    if (resolvedAddress != null) {
      // Green host DNS exists
      throw throwable;
    }

    // Green host DNS doesn't exist. Try to replace it with the corresponding hostname and connect again.

    final String fixedHost = rdsUtils.removeGreenInstancePrefix(hostSpec.getHost());
    return new HostSpecBuilder(hostSpec.getHostAvailabilityStrategy())
        .copyFrom(hostSpec)
        .host(fixedHost)
        .build();
  }
```

wrapper/src/main/java/software/amazon/jdbc/DataSourceConnectionProvider.java
sergiyvamz: I'm not sure. Checking DNS upfront takes time. Checking when exception thrown is handling error. Maybe we will make such change later as part of optimization.

wrapper/src/main/java/software/amazon/jdbc/DriverConnectionProvider.java
sergiyvamz: We can relocate this code to a helper class. The main question whether any other places that require such functionality. I don't see any other place in the driver codebase that may require such checks.

davecramer:
this fixes issue #678, correct ?

PR #1005, opened 2024-05-22
fix RW splitting tests

sergiyvamz:
### Summary

fix RW splitting tests

### Description

- Enable RW splitting tests. Various fixes for tests making them more consistent and stable.
- Refactoring cluster health check with an option to reboot test cluster. Check cluster endpoint and cluster RO endpoint to resolve according to a latest cluster topology.
- Rework various timeouts in tests
- Fixes for failover process taking into account RDS MultiAz clusters behaviour and Aurora clusters.
- Fixes for host list providers

### Additional Reviewers

@crystall-bitquill 
@karenc-bq 
@aaronchung-bitquill 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsPgDialect.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/dialect/RdsPgDialect.java
@@ -20,7 +20,7 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Logger;
 
@@ -34,7 +34,8 @@ public class RdsPgDialect extends PgDialect {
 
   private static final Logger LOGGER = Logger.getLogger(RdsPgDialect.class.getName());
 
-  private static final List<String> dialectUpdateCandidates = Collections.singletonList(
+  private static final List<String> dialectUpdateCandidates = Arrays.asList(
+      DialectCodes.RDS_MULTI_AZ_PG_CLUSTER,
       DialectCodes.AURORA_PG);
 
   private static final String extensionsSql = "SELECT (setting LIKE '%rds_tools%') AS rds_tools, "
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
@@ -326,7 +326,7 @@ public class MonitorImpl implements Monitor {
     }
 
     LOGGER.finest(() -> Messages.get(
-        "MonitorImpl.startMonitoringThread",
+        "MonitorImpl.stopMonitoringThread",
         new Object[]{this.hostSpec.getHost()}));
   }
 
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
@@ -326,15 +326,18 @@ public class ReadWriteSplittingPlugin extends AbstractConnectionPlugin
           switchToReaderConnection(hosts);
         } catch (final SQLException e) {
           if (!isConnectionUsable(currentConnection)) {
-            logAndThrowException(Messages.get("ReadWriteSplittingPlugin.errorSwitchingToReader"),
-                SqlState.CONNECTION_UNABLE_TO_CONNECT, e);
+            logAndThrowException(
+                Messages.get("ReadWriteSplittingPlugin.errorSwitchingToReader", new Object[] { e.getMessage() }),
+                SqlState.CONNECTION_UNABLE_TO_CONNECT,
+                e);
             return;
           }
 
-          // Failed to switch to a reader; use writer as a fallback
-          LOGGER.warning(() -> Messages.get(
+          // Failed to switch to a reader. {0}. The current writer will be used as a fallback: ''{1}''
+          LOGGER.fine(() -> Messages.get(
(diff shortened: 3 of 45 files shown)


wrapper/src/main/java/software/amazon/jdbc/dialect/RdsPgDialect.java
```diff
@@ -34,7 +34,8 @@ public class RdsPgDialect extends PgDialect {
 
   private static final Logger LOGGER = Logger.getLogger(RdsPgDialect.class.getName());
 
-  private static final List<String> dialectUpdateCandidates = Collections.singletonList(
+  private static final List<String> dialectUpdateCandidates = Arrays.asList(
+      DialectCodes.RDS_MULTI_AZ_PG_CLUSTER,
```
aaronchung-bitquill, line 38: minor nit: will this list ever change? If not then can we use something like Collections.unmodifiableList?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
```diff
@@ -326,7 +326,7 @@ public void run() {
     }
 
     LOGGER.finest(() -> Messages.get(
-        "MonitorImpl.startMonitoringThread",
+        "MonitorImpl.stopMonitoringThread",
```
joyc-bq, line 329: is this message name correct? On main I see it's called `MonitorImpl.stopped`

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -370,7 +373,7 @@ private void logAndThrowException(final String logMessage, final SqlState sqlSta
   private void logAndThrowException(
       final String logMessage, final SqlState sqlState, final Throwable cause)
       throws SQLException {
-    LOGGER.severe(logMessage);
+    LOGGER.fine(logMessage);
```
aaronchung-bitquill, line 376: Is the intent to eliminate potential double logging, as this exception is likely to be caught and logged upstream?

wrapper/src/test/java/integration/DriverHelper.java
```diff
@@ -202,10 +202,12 @@ public static String getHostnameSql(DatabaseEngine databaseEngine) {
     }
   }
 
+  // The method should be used on connections with a target drivers ONLY !!!
   public static void setConnectTimeout(Properties props, long timeout, TimeUnit timeUnit) {
     setConnectTimeout(TestEnvironment.getCurrent().getCurrentDriver(), props, timeout, timeUnit);
   }
 
+  // The method should be used on connections with a target drivers ONLY !!!
```
joyc-bq, line 210:
```suggestion
  // This method should be used on connections with a target driver ONLY!
```
this is a nit, apply to both comments

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -326,15 +326,18 @@ void switchConnectionIfRequired(final boolean readOnly) throws SQLException {
...
+            logAndThrowException(
+                Messages.get("ReadWriteSplittingPlugin.errorSwitchingToReader", new Object[] { e.getMessage() }),
+                SqlState.CONNECTION_UNABLE_TO_CONNECT,
+                e);
             return;
           }
 
-          // Failed to switch to a reader; use writer as a fallback
-          LOGGER.warning(() -> Messages.get(
+          // Failed to switch to a reader. {0}. The current writer will be used as a fallback: ''{1}''
```
joyc-bq, line 336: this comment uses {0} and {1} which will not print anything
```suggestion
          // Failed to switch to a reader. The current writer will be used as a fallback.
```

aaronchung-bitquill (approved):
Since it looks like there are no major issues and just nits, I'll approve this and leave addressing the comments up to your discretion.

joyc-bq (approved):
approved with comments

wrapper/src/main/java/software/amazon/jdbc/dialect/RdsPgDialect.java
sergiyvamz: We can use `Collections.unmodifiableList` but this list is a private member and there's no chance that user, or a custom user plugin, can change it, intentionally or accidentally.

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java
sergiyvamz: I think it's a copy-paste kind of error. It should be a "thread stopped" message. There're two MonitorImpl class. One for efm plugin, the other is for efm2 plugin. You must be checked the class from efm package (that's correct).

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/e90e1aba7e34bb1068789469cd508c35a5cd7b17/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/MonitorImpl.java#L328

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
sergiyvamz: I think we might scare users by logging SEVERE message. Severe isn't a proper level of error. It's not a critical error, the plugin can handle it properly and continue with normal workflow. So, I think that FINE would be appropriate level here.

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
sergiyvamz: There're two parameters are passed to `Messages.get()`: error message and connection host name.

Issue #1075, opened 2024-08-02
Driver not failing over to readers


Milesy:
### Describe the bug

When doing a failover of a writer, readonly connections are not failing over to the available readers. The failover takes approx 20 seconds while it waits on the new writer node starting up. 

### Expected Behavior

I expect, the failover to happen within seconds, and my read-only transactions to be split to the reader nodes.

### What plugins are used? What other connection properties were set?

ConnectTimeConnectionPluginFactory,ExecutionTimeConnectionPluginFactory,IamAuthConnectionPluginFactory,AuroraInitialConnectionStrategyPluginFactory,AuroraConnectionTrackerPluginFactory,ReadWriteSplittingPluginFactory,FailoverConnectionPluginFactory,HostMonitoringConnectionPluginFactory

### Current Behavior

Using the driver with Postgres Aurora Serverless, I have configured failover, with internal connection pooling, with read-write splitting enabled.

I have also set the failover mode to STRICT_READER (can be seen in the logs), and my tests are all against read only connections (can also be seen in the logs)

> 2024-08-02 19:01:34.965 DEBUG [ddTraceId-:ddSpanId-] 5488 --- [   scheduling-1] s.a.j.p.f.FailoverConnectionPlugin       : [] failoverMode=STRICT_READER

> 2024-08-02 19:01:34.965 TRACE [ddTraceId-:ddSpanId-] 5488 --- [   scheduling-1] s.a.jdbc.plugin.DefaultConnectionPlugin  : [] Executing method: 'Connection.setReadOnly'

I have 1 WRITER and 3 READERSs.

> 2024-08-02 19:01:03.781 TRACE [ddTraceId-:ddSpanId-] 5488 --- [   scheduling-1] s.a.j.h.RdsHostListProvider              : [] [From cache] Topology: 
>    HostSpec[host=aurora-reader-node-1-pyxdlwfq.c4q03196xb4j.us-east-1.rds.amazonaws.com, port=6160, READER, AVAILABLE, weight=4523, 2024-08-02 19:00:34.0]
>    HostSpec[host=aurora-reader-node-3-pyxdlwfq.c4q03196xb4j.us-east-1.rds.amazonaws.com, port=6160, READER, AVAILABLE, weight=2537, 2024-08-02 19:00:34.0]
[...]

PR #1099, opened 2024-08-23
Allow optional failover in the connect pipeline

aaron-congo:
### Summary

- adds failover logic to the connect pipeline. If the initial connection fails, the driver will attempt to use failover to connect to a different instance. This behavior is disabled by default but can be enabled by setting the "enableConnectFailover" connection property
- related to issue-1075: https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1075


wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
```diff
@@ -156,7 +156,7 @@ public HikariPooledConnectionProvider(
   public boolean acceptsUrl(
       @NonNull String protocol, @NonNull HostSpec hostSpec, @NonNull Properties props) {
     final RdsUrlType urlType = rdsUtils.identifyRdsType(hostSpec.getHost());
-    return RdsUrlType.RDS_INSTANCE.equals(urlType);
+    return RdsUrlType.RDS_INSTANCE.equals(urlType) || hostSpec.getHost().contains(".proxied");
```
aaron-congo, line 159: I changed this temporarily so that the integrations can use the toxiproxy instance URLs and still use internal connection pools. We probably will have to come up with an alternative to this change.

sergiyvamz: Should we add a new parameter to HikariPooledConnectionProvider constructor similar to HikariPoolMapping? A function/lambda that returns true or false depending on host name. It can check if pool can accept that host name, if host name is a RDS instance name.

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/HikariPooledConnectionProvider.java
aaron-congo: Added now

karenc-bq (approved):

sergiyvamz (approved):

wrapper/src/test/java/integration/container/tests/HikariTests.java
```diff
@@ -252,7 +260,206 @@ public void testEFMFailover() throws SQLException {
...
+      ConnectionProviderManager.releaseResources();
+    }
+  }
+
+  /**
+   * Given an instance URL, extracts the substring of the URL that is common to all instance URLs. For example, given
+   * "instance-1.ABC.cluster-XYZ.us-west-2.rds.amazonaws.com.proxied", returns
+   * ".ABC.cluster-XYZ.us-west-2.rds.amazonaws.com.proxied"
+   */
+  private static String getInstanceUrlSubstring(String instanceUrl) {
```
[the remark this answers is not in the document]
aaron-congo: Removed now

PR #1104, opened 2024-08-28
feat: limitless connection plugin - synchronously fetch router info if cache is empty

aaronchung-bitquill:
### Summary

If a new connection is made and the router cache is empty, synchronously fetch router info before trying to select a router.. 
This behaviour is toggle-able with a property `limitlessWaitForRouterInfo` which is default to true. 
If set to `false` and router cache is empty, it will fallback to using the provided DB Shard Group endpoint URL


wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -40,6 +40,10 @@
 
 public class LimitlessConnectionPlugin extends AbstractConnectionPlugin {
   private static final Logger LOGGER = Logger.getLogger(LimitlessConnectionPlugin.class.getName());
+  protected static final AwsWrapperProperty WAIT_F0R_ROUTER_INFO = new AwsWrapperProperty(
+      "limitlessWaitForRouterInfo",
+      "true",
+      ".");
```
aaronchung-bitquill, line 46: forgot to add description here

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -195,13 +206,24 @@ private Connection retryConnection(final List<HostSpec> limitlessRouters, final
...
+
+  private synchronized List<HostSpec> synchronouslyGetLimitlessRouters(final Properties props) throws SQLException {
+    final List<HostSpec> currentLimitlessRouters = this.limitlessRouterService.getLimitlessRouters(
+        this.pluginService.getHostListProvider().getClusterId(), props);
+    if (!currentLimitlessRouters.isEmpty()) {
+      return currentLimitlessRouters;
+    }
+
+    LOGGER.fine(Messages.get("LimitlessConnectionPlugin.synchronouslyGetLimitlessRouters"));
+    final List<HostSpec> newLimitlessRouters = this.limitlessRouterService.forceGetLimitlessRouters(
```
sergiyvamz, line 225: It still might be empty list in some cases. Should we add a loop?

aaronchung-bitquill: Good point. I'll add a while loop as well as some properties for max tries and retry intervals.

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -195,13 +216,42 @@ private Connection retryConnection(final List<HostSpec> limitlessRouters, final
...
+    LOGGER.fine(Messages.get("LimitlessConnectionPlugin.synchronouslyGetLimitlessRouters"));
+    List<HostSpec> newLimitlessRouters = this.limitlessRouterService.forceGetLimitlessRouters(
+        this.pluginService.getHostListProvider().getClusterId(), props);
+
+    int retryCount = 0;
+    int maxRetries = GET_ROUTER_MAX_RETRIES.getInteger(props);
+    int retryIntervalMs = GET_ROUTER_RETRY_INTERVAL_MILLIS.getInteger(props);
+    while (newLimitlessRouters.isEmpty() && retryCount < maxRetries) {
+      try {
+        newLimitlessRouters = this.limitlessRouterService.forceGetLimitlessRouters(
```
sergiyvamz, line 244: If we get a good (non-empty) list of routers we're going to sleep and then return the list? It doesn't look optimal.

sergiyvamz (approved):

PR #1109, opened 2024-09-03
fix: limitless connection plugin - add synchornized lock for synchron…

aaronchung-bitquill:
…ous fetch of routers

### Summary
Add synchronized lock to the synchronous fetch of routers so `synchronized` will apply to all instances of the Limitless plugin. 

Before this fix, I did a test spinning up 100 callables creating connections. I noticed `limitlessRouterService#forceGetLimitlessRouters` being called 10+  times at start up, when only 1 is needed. This should remedy that.  

Also moved the `retryCounter` to `finally` block so that if an exception is encountered, there isnt an infinite loop.

### Description

Added a static synchronized lock to the LimitlessConnectionPlugin to be used for `synchronouslyGetLimitlessRouters()`.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
@@ -40,25 +40,25 @@ import software.amazon.jdbc.wrapper.HighestWeightHostSelector;
 
 public class LimitlessConnectionPlugin extends AbstractConnectionPlugin {
   private static final Logger LOGGER = Logger.getLogger(LimitlessConnectionPlugin.class.getName());
-  protected static final AwsWrapperProperty WAIT_F0R_ROUTER_INFO = new AwsWrapperProperty(
+  public static final AwsWrapperProperty WAIT_F0R_ROUTER_INFO = new AwsWrapperProperty(
       "limitlessWaitForTransactionRouterInfo",
       "true",
       "If the cache of transaction router info is empty and a new connection is made, this property toggles whether "
           + "the plugin will wait and synchronously fetch transaction router info before selecting a transaction "
           + "router to connect to, or to fall back to using the provided DB Shard Group endpoint URL.");
-  protected static final AwsWrapperProperty GET_ROUTER_RETRY_INTERVAL_MILLIS = new AwsWrapperProperty(
+  public static final AwsWrapperProperty GET_ROUTER_RETRY_INTERVAL_MILLIS = new AwsWrapperProperty(
       "limitlessGetTransactionRouterInfoRetryIntervalMs",
       "300",
       "Interval in millis between retries fetching Limitless Transaction Router information.");
-  protected static final AwsWrapperProperty GET_ROUTER_MAX_RETRIES = new AwsWrapperProperty(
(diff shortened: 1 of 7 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -224,21 +225,24 @@ private Connection retryConnection(final List<HostSpec> limitlessRouters, final
...
-      return newLimitlessRouters;
+      LOGGER.fine(Messages.get("LimitlessConnectionPlugin.synchronouslyGetLimitlessRouters"));
+      List<HostSpec> newLimitlessRouters = this.limitlessRouterService.forceGetLimitlessRouters(
+          this.pluginService.getHostListProvider().getClusterId(), props);
+      if (newLimitlessRouters != null && !newLimitlessRouters.isEmpty()) {
+        return newLimitlessRouters;
+      }
     }
 
+    List<HostSpec> newLimitlessRouters = Collections.emptyList();
```
ColinKYuen, line 245: This is already declared above L238. Does Java not have a compile error when this happens?

ColinKYuen: Ah, nevermind. Scoped differently. Hard to see on Github

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -224,21 +225,24 @@ private Connection retryConnection(final List<HostSpec> limitlessRouters, final
...
 
-  private synchronized List<HostSpec> synchronouslyGetLimitlessRouters(final Properties props)
+  private List<HostSpec> synchronouslyGetLimitlessRouters(final Properties props)
       throws SQLException {
-    final List<HostSpec> currentLimitlessRouters = this.limitlessRouterService.getLimitlessRouters(
-        this.pluginService.getHostListProvider().getClusterId(), props);
-    if (!currentLimitlessRouters.isEmpty()) {
-      return currentLimitlessRouters;
-    }
+    synchronized (synchronouslyGetLimitlessRoutersLock) {
```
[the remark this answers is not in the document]
aaronchung-bitquill: Fixed!

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -180,18 +181,30 @@ private void initLimitlessRouterMonitorService() {
...
 
-  private Connection retryConnection(final List<HostSpec> limitlessRouters, final Properties props, int inputRetryCount,
-      final SQLException originalException) throws SQLException {
+  private Connection retryConnectWithLeastLoadedRouters(final List<HostSpec> limitlessRouters,
+      final Properties props, final SQLException originalException) throws SQLException {
 
-    int retryCount = inputRetryCount;
+    List<HostSpec> currentRouters = limitlessRouters;
+    int retryCount = 0;
     final int maxRetries = MAX_RETRIES.getInteger(props) >= 0
```
sergiyvamz, line 189: I believe the following line would be enough. It already includes the logic to validate user input and check for default value.

maxRetries = MAX_RETRIES.getInteger(props);

sergiyvamz (approved):

PR #1122, opened 2024-09-18
Implement the CustomEndpointPlugin

aaron-congo:
### Summary

Implement the CustomEndpointPlugin

### Description

- when reviewing, please leave notes on the TODO items if you have input
- as custom endpoints are used, monitor threads are created to monitor custom endpoint information
- custom endpoint information is fetched using the RDS API, so users of the CustomEndpointPlugin need to have the AWS Java SDK on their classpath
- connections register their plugin services with a given monitor if they are connecting to that monitor's custom endpoint
- when the monitor detects changes in custom endpoint information, it updates the custom endpoint information for all registered plugins
- added new methods in PluginService to get/set custom endpoint information, and to fetch the current list of hosts minus the hosts that are not allowed by the custom endpoint plugin. The failover and read/write plugin use this subset of hosts when deciding which host to switch to


wrapper/src/main/java/software/amazon/jdbc/util/SlidingExpirationCache.java
```diff
@@ -118,10 +118,15 @@ protected void removeAndDispose(K key) {
...
-    }
+    // TODO: what if the disposal function is long-running? It will lock up the cache until its done.
+    cache.computeIfPresent(key, (k, cacheItem) -> {
+      if (cacheItem.shouldCleanup()) {
+        removeAndDispose(key);
+        return null;
+      }
+
+      return cacheItem;
+    });
```
aaron-congo, line 129: This change was introduced because the shouldCleanup check and the removal of the item from the cache need to be synchronized. However, my concern with this change is that some disposal functions may run for a long time and the cache will be locked during this time, for example the failover2 plugin attempts to clean up its threads for up to 30 seconds. I could implement an additional change outlined below but I'm not sure if there's a better way. The additional change moves the disposal call until after the item has been removed from the cache and the cache has been unlocked.
```suggestion
    final AtomicReference<V> itemRef = new AtomicReference<>();
    cache.computeIfPresent(key, (k, cacheItem) -> {
      if (cacheItem.shouldCleanup()) {
        cache.remove(key);
        itemRef.set(cacheItem.item);
        return null;
      }

      return cacheItem;
    });

    V item = itemRef.get();
    if (item != null && itemDisposalFunc != null) {
      itemDisposalFunc.dispose(item);
    }
```

sergiyvamz: That's a valid concern. I'd just simplify it a bit. I don't think we need weak pointers here. A regular list (ArrayList) would be just fine. Line 125 (cache.remove) is unnecessary since your lambda return null that equals to item deletion.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -110,6 +112,7 @@ public class ConnectionPluginChainBuilder {
           put(ConnectTimeConnectionPluginFactory.class, WEIGHT_RELATIVE_TO_PRIOR_PLUGIN);
           put(ExecutionTimeConnectionPluginFactory.class, WEIGHT_RELATIVE_TO_PRIOR_PLUGIN);
           put(DeveloperConnectionPluginFactory.class, WEIGHT_RELATIVE_TO_PRIOR_PLUGIN);
+          put(CustomEndpointPluginFactory.class, WEIGHT_RELATIVE_TO_PRIOR_PLUGIN);
```
sergiyvamz, line 115: I'd put it before RW splitting plugin.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointInfo.java
```diff
@@ -0,0 +1,137 @@
...
+      MemberListType memberListType) {
+    this.endpointIdentifier = endpointIdentifier;
+    this.clusterIdentifier = clusterIdentifier;
+    this.url = url;
+    this.customEndpointType = customEndpointType;
+    this.members = members;
+    this.memberListType = memberListType;
+  }
+
+  public static CustomEndpointInfo fromDBClusterEndpoint(DBClusterEndpoint responseEndpointInfo) {
```
sergiyvamz, line 53: static?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,229 @@
...
+    this.customEndpointHostSpec = customClusterHostSpec;
+    this.region = getRegion(customClusterHostSpec, props);
+    this.rdsClient = rdsClientFunc.apply(customClusterHostSpec, this.region);
+    this.cacheEntryExpirationNano = cacheEntryExpirationNano;
+
+    this.refreshRateNano = TimeUnit.MILLISECONDS.toNanos(CUSTOM_ENDPOINT_INFO_REFRESH_RATE.getLong(props));
+
+    this.endpointIdentifier = this.rdsUtils.getRdsClusterId(customClusterHostSpec.getHost());
+    if (StringUtils.isNullOrEmpty(this.endpointIdentifier)) {
+      throw new RuntimeException(
```
sergiyvamz, line 86: I'm not sure RuntimeException. Should we consider SQLException?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,229 @@
...
+                      this.endpointIdentifier,
+                      this.region.id(),
+                      endpoints.size(),
+                      endpointURLs
+                  }
+              ));
+
+          // TODO: should we throw an exception and then return? Or wait and then try again?
+          long elapsedTime = System.nanoTime() - start;
+          long sleepDuration = Math.min(0, this.refreshRateNano - elapsedTime);
```
sergiyvamz, line 195: I don't think we need such a precision. We can just sleep for specified time despite how much time API call takes.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,229 @@
...
+          TimeUnit.NANOSECONDS.sleep(sleepDuration);
+          continue;
+        }
+
+        CustomEndpointInfo endpointInfo = CustomEndpointInfo.fromDBClusterEndpoint(endpoints.get(0));
+        if (customEndpointInfoCache.get(this.endpointIdentifier).equals(endpointInfo)) {
```
sergiyvamz, line 201: Qodana has a point. It could be a NPE here if cache has no such key.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,229 @@
...
+          long elapsedTime = System.nanoTime() - start;
+          long sleepDuration = Math.min(0, this.refreshRateNano - elapsedTime);
+          TimeUnit.NANOSECONDS.sleep(sleepDuration);
+          continue;
+        }
+
+        // The custom endpoint info has changed, so we need to update the info in the registered plugin services.
+        customEndpointInfoCache.put(this.endpointIdentifier, endpointInfo, this.cacheEntryExpirationNano);
```
sergiyvamz, line 209: Another thread might add such item in the cache. Would it be damaging? Can we use computeIfAbsent()?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,229 @@
...
+        this.pluginServicesLock.readLock().lock();
+        try {
+          for (PluginService service : this.pluginServices) {
```
sergiyvamz, line 212: Each plugin service instance belongs to a wrapper connection. By keeping references to plugin services you prevent them from disposing by GC. You need to consider using weak pointers.

sergiyvamz: Also, you call plugin services from a monitoring thread while plugin service might handle some other action so PluginServiceImpl should be multi-thread safe that we never though about. 

Is there alternative for keeping plugin services and sending them an updates?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,166 @@
...
+      boolean isInitialConnection,
+      JdbcCallable<Connection, SQLException> connectFunc) throws SQLException {
+    final Connection conn = connectFunc.call();
+    if (!isInitialConnection || !this.rdsUtils.isRdsCustomClusterDns(hostSpec.getHost())) {
+      return conn;
+    }
+
+    // If we get here we are making an initial connection to a custom cluster URL
+    String customClusterHost = hostSpec.getHost();
+    CustomEndpointMonitor monitor = monitors.get(customClusterHost, MONITOR_EXPIRATION_NANO);
```
sergiyvamz, line 139: monitors.computeIfPresent()?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,166 @@
...
+    if (monitor != null) {
+      // A monitor is already running for this custom endpoint.
+      this.customEndpointMonitor = monitor;
+      monitor.registerPluginService(this.pluginService);
+      return conn;
+    }
+
+    this.customEndpointMonitor = monitors.computeIfAbsent(
```
sergiyvamz, line 147: Should we switch two blocks at lines 139-145 and 147-156?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,166 @@
...
+        (customEndpoint) -> new CustomEndpointMonitorImpl(
+            hostSpec,
+            props,
+            this.rdsClientFunc,
+            CUSTOM_ENDPOINT_INFO_EXPIRATION_NANO
+        ),
+        MONITOR_EXPIRATION_NANO
+    );
+
+    this.customEndpointMonitor.registerPluginService(this.pluginService);
```
sergiyvamz, line 158: Cn we get Custom Endpoint status and update plugin service (instead of registering plugin service at monitor)?

wrapper/src/main/java/software/amazon/jdbc/PluginService.java
```diff
@@ -62,8 +62,11 @@ EnumSet<NodeChangeOptions> setCurrentConnection(
       @Nullable ConnectionPlugin skipNotificationForThisPlugin)
       throws SQLException;
 
+  // TODO: should we rename to getAllHosts?
   List<HostSpec> getHosts();
```
sergiyvamz, line 66: I'd keep getHosts() as-is so we can avoid dozens of changes throughout the code. getHosts() should return a list of nodes excluding those that do not conform with white/disabled lists.

And, at the same time, I'd add another method getAllHosts() that returns all topology nodes including not allowed ones.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
aaron-congo: If we put it before the RW splitting plugin the monitor thread will actually be created later rather than sooner, because the monitoring thread is created after connecting when we are returning backwards up the connect pipeline.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Sure. I noticed we are throwing RuntimeExceptions instead of SQLExceptions in some other areas of the code too, for example in the EFM plugin when we have trouble getting the node ID. Its a bit unclear to me how we're deciding when to use RuntimeExceptions instead of SQLExceptions - any tips?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Sure. Do you want me to remove these lines below as well?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Forgot to address this, will update

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: I don't think any other thread will be placing info at this key, the logic in CustomEndpointPlugin should only be creating one thread per custom endpoint which is shared among all connections, which means there is only one thread updating info for a given custom endpoint. computeIfAbsent would allow us to place an initial custom endpoint info object in the cache but would not allow us to update the info later.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: When connections are being closed they deregister their plugin service in releaseResources which would remove the reference and allow the GC to dispose of the connection, however I should maybe change to weak references anyways in case the connection does not cleanly close. 

On main, the PluginServiceImpl class already has a lock variable (`connectionSwitchLock`, which is used in `setCurrentConnection`), doesn't that imply the plugin service is already being used in some multi-threaded capacity? In the case of this PR, the shared data is stored in the statuses CacheMap which uses a ConcurrentHashMap, so I believe it this scenario should be thread-safe (let me know if I've missed something).

Regardless, I'm open to other suggestions if you have them. The tricky thing here is that we need a mechanism that allows each custom endpoint connection to update their custom endpoint info regularly over time. The only other way I can think of doing that is having a method that plugins or the plugin service can call to grab custom endpoint info from the cache, similar to `refresh` and `forceRefresh` in the host list provider. But plugins, and the plugin service, aren't aware of what other plugins are in use so they cannot call the custom endpoint plugin directly.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: `monitors` is a SlidingExpirationCache, which does not expose a computeIfPresent method, but I just realized I can get rid of this block and clean things up, so we shouldn't need it

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: I think we can actually get rid of the 139-145 block

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: We could, which would allow connections to get initial custom endpoint info, but then there would be no way for the connections to update their custom endpoint info over time (at least not with the current implementation). The purpose of registering the plugin services is so that the custom endpoint monitor can update the plugin service with new custom endpoint info if the custom endpoint changes

wrapper/src/main/java/software/amazon/jdbc/PluginService.java
aaron-congo: Okay, so all places that currently call getHosts() on main will now get the subset of hosts? I'm not sure if we want to do that in all places or if some of them should be updated to call getAllHosts()

sergiyvamz: Any particular place where we don't want that subset of hosts to be used?

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
sergiyvamz: We usually start monitoring before opening a connection. But I agree with your point that it makes more sense to start it when connection is successfully established. Either way work.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
sergiyvamz: I'd still suggest to avoid registering plugin service at the monitor. Instead, the plugin can fetch custom endpoint info, check if there's any changes and update its local plugin service if needed.

aaron-congo: that is a better approach, but I'm not sure how to achieve it since the plugin service is not directly aware of which plugins are being used. With the host list providers, the plugin service has a direct reference to them so it can call `hostListProvider.refresh`. However, it can't call `customEndpointPlugin.refresh`. Any suggested solutions to this?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: I removed the plugin service list after further discussions with Sergiy

wrapper/src/main/java/software/amazon/jdbc/PluginService.java
aaron-congo: I reviewed the calls to getHosts() with Sergiy and we determined where to grab all the hosts vs just the subset

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: After chatting with Sergiy and changing the design, we are using a single plugin service which can be used to updated the shared plugin service info cache with custom endpoint info.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
sergiyvamz: The reason to checked `SQLException` is that a user app catch them and can handle them. Unchecked exceptions are mostly for some unpredictable/unhandled cases.

In this particular case I'd suggest to wrap logic inside `createMonitorIfAbsent` in try-catch. You can catch `RutimeException` and convert it to `SQLException` and raise.

wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
```diff
@@ -365,18 +384,37 @@ protected EnumSet<NodeChangeOptions> compare(
...
 
+  @Override
+  public List<HostSpec> getHosts() {
+    CustomEndpointInfo customEndpointInfo =
+        this.getInfo(this.initialConnectionHostSpec.getHost(), CustomEndpointInfo.class, true);
+    if (customEndpointInfo == null) {
+      return this.hosts;
+    }
+
+    if (STATIC_LIST.equals(customEndpointInfo.getMemberListType())) {
```
sergiyvamz, line 399: I'd prefer to keep cutsom endpoint related classes and types in the CustomEndpoint plugin. PluginServiceImpl should have no references to them.

wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
```diff
@@ -365,18 +384,37 @@ protected EnumSet<NodeChangeOptions> compare(
...
+      return this.hosts.stream()
+          .filter((hostSpec -> customEndpointInfo.getStaticMembers().contains(hostSpec.getHostId())))
```
sergiyvamz, line 401: Search isn't efficient since customEndpointInfo.getStaticMembers() is List<String>. It's better to use Set<String>.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  protected final AtomicBoolean stop = new AtomicBoolean(false);
+  protected final RdsUtils rdsUtils = new RdsUtils();
+  protected final RdsClient rdsClient;
+  protected final HostSpec customEndpointHostSpec;
+  protected final String endpointIdentifier;
+  protected final Region region;
+  protected final long refreshRateNano;
+
+  protected final PluginService pluginService;
+  // TODO: is it problematic for each monitor thread to have its own executor service?
```
sergiyvamz, line 63: It's not a problem.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  protected final ExecutorService monitorExecutor = Executors.newSingleThreadExecutor(runnableTarget -> {
+    final Thread monitoringThread = new Thread(runnableTarget);
+    monitoringThread.setDaemon(true);
+    if (!StringUtils.isNullOrEmpty(monitoringThread.getName())) {
+      monitoringThread.setName(monitoringThread.getName() + "-CustomEndpointMonitor");
```
sergiyvamz, line 68: Thread name will be shown in logs. Can we make that prefix shorter? "-cem" ?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+   */
+  @Override
+  public void run() {
+    LOGGER.fine(
+        Messages.get(
+            "CustomEndpointMonitorImpl.startingMonitor",
+            new Object[] { this.customEndpointHostSpec.getHost() }));
+
+    try {
+      while (!this.stop.get() && !Thread.currentThread().isInterrupted()) {
```
sergiyvamz, line 125: Can we rid of `Thread.currentThread().isInterrupted()` condition and replace it with `return`  in corresponding try-catch block?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+                      this.endpointIdentifier,
+                      this.region.id(),
+                      endpoints.size(),
+                      endpointURLs
+                  }
+              ));
+
+          TimeUnit.NANOSECONDS.sleep(this.refreshRateNano);
+          continue;
+        }
```
sergiyvamz, line 150: Should we report about internal issue if the list contains more than one element?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+
+        long elapsedTime = System.nanoTime() - start;
+        long sleepDuration = Math.min(0, this.refreshRateNano - elapsedTime);
+        TimeUnit.NANOSECONDS.sleep(sleepDuration);
+      }
+    } catch (InterruptedException e) {
+      LOGGER.info(Messages.get("CustomEndpointMonitorImpl.interrupted", new Object[]{ this.customEndpointHostSpec }));
+      Thread.currentThread().interrupt();
+    } catch (Exception e) {
+      LOGGER.log(Level.SEVERE,
```
sergiyvamz, line 180: This unhandled exception effectively stops monitoring thread. I believe we need to report the issue and continue with the main monitoring loop.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  public void close() {
+    LOGGER.fine(
+        Messages.get(
+            "CustomEndpointMonitorImpl.stoppingMonitor",
+            new Object[]{ this.customEndpointHostSpec.getHost() }));
+
+    this.stop.set(true);
+
+    try {
+      // TODO: the run loop takes 30s by default but could take more depending on the user setting. Should we keep 30s
```
sergiyvamz, line 207: I think we can wait less than 30s. Most of the time the thread sleeps that can be interrupted by `this.monitorExecutor.shutdownNow()`. Any interaction with RDS API takes seconds to complete. I'd wait just 5s.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+  static {
+    PropertyDefinition.registerPluginProperties(CustomEndpointPlugin.class);
+  }
+
+  protected final PluginService pluginService;
+  protected final Properties props;
+  protected final RdsUtils rdsUtils = new RdsUtils();
+  protected final BiFunction<HostSpec, Region, RdsClient> rdsClientFunc;
+
+  private final TelemetryCounter endpointConnectionsCounter;
```
sergiyvamz, line 103: Why is it private and not protected?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+  protected Connection connectInternal(
+      HostSpec hostSpec,
+      Properties props,
+      JdbcCallable<Connection, SQLException> connectFunc) throws SQLException {
+    if (!this.rdsUtils.isRdsCustomClusterDns(hostSpec.getHost())) {
+      return connectFunc.call();
+    }
+
+    this.customEndpointHostSpec = hostSpec;
+    this.endpointConnectionsCounter.inc();
```
sergiyvamz, line 198: I'm not sure what the value this counter can provide to a user.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+
+  /**
+   * Determines the AWS region from the given parameters. The region specified by the props will be used if it is
+   * provided. Otherwise, the region will be parsed from the specified host URL.
+   *
+   * @param hostSpec Host information for the connection being established.
+   * @param props    The connection properties for the connection being established.
+   * @return The AWS region of the connection being established.
+   */
+  protected Region getRegion(HostSpec hostSpec, Properties props) {
```
sergiyvamz, line 244: I've seen similar piece of code before. It seems we can put it in a helper class.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+          Messages.get(
+              "CustomEndpointPlugin.waitingForCustomEndpointInfo",
+              new Object[]{ this.customEndpointHostSpec.getHost(), this.waitOnCachedInfoDurationMs }));
+      long waitForEndpointInfoTimeoutNano =
+          System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(this.waitOnCachedInfoDurationMs);
+
+      while (!isInfoInCache && System.nanoTime() < waitForEndpointInfoTimeoutNano) {
+        cachedInfo =
+            this.pluginService.getInfo(this.customEndpointHostSpec.getHost(), CustomEndpointInfo.class, true);
+        isInfoInCache = cachedInfo != null;
```
sergiyvamz, line 296: any sleep() here?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+      }
+
+      if (isInfoInCache) {
+        LOGGER.fine(
```
sergiyvamz, line 300: Should we raise TimeoutException here?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+
+  /**
+   * If custom endpoint info does not exist for the current custom endpoint, waits a short time for the info to be
+   * made available by the custom endpoint monitor. This is necessary so that other plugins can rely on accurate custom
+   * endpoint info. Since custom endpoint monitors and information are shared, we should not have to wait often.
+   *
+   * @return true if the custom endpoint info is available, or false if we timed out while waiting for the info to
+   *     become available.
+   */
+  protected boolean waitForCustomEndpointInfo() {
```
sergiyvamz, line 277: What if a user doesn't want to wait? Can we not wait and proceed with a custom endpoint and let DNS resolve it for us?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,370 @@
...
+    return jdbcMethodFunc.call();
+  }
+
+  /**
+   * Closes all active custom endpoint monitors.
+   */
+  public static void closeMonitors() {
+    LOGGER.info(Messages.get("CustomEndpointPlugin.closeMonitors"));
+
+    for (CustomEndpointMonitor monitor : monitors.getEntries().values()) {
```
sergiyvamz, line 360: It's better to move it to `monitors.clear()`

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -668,6 +673,18 @@ protected void failoverWriter() throws SQLException {
 
       // successfully re-connected to a writer node
       final HostSpec writerHostSpec = getWriter(failoverResult.getTopology());
+      final List<HostSpec> allowedHosts = this.pluginService.getHosts();
+      if (!allowedHosts.contains(writerHostSpec)) {
+        // TODO: should we increment the writer failed counter here or not?
```
sergiyvamz, line 678: Yes we should increment it

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -502,6 +504,16 @@ protected void failoverWriter() throws SQLException {
           .findFirst()
           .orElse(null);
 
+      List<HostSpec> allowedHosts = this.pluginService.getHosts();
+      if (writerCandidate != null && !allowedHosts.contains(writerCandidate)) {
+        // TODO: should we increment the writer failed counter here or not?
```
sergiyvamz, line 509: Yes we should increment it

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -492,7 +494,7 @@ protected void failoverWriter() throws SQLException {
         throw new FailoverFailedSQLException(Messages.get("Failover.unableToConnectToWriter"));
       }
 
-      final List<HostSpec> updatedHosts = this.pluginService.getHosts();
+      final List<HostSpec> updatedHosts = this.pluginService.getAllHosts();
```
sergiyvamz, line 497: Does it make sense to use `getHosts()` here?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Okay, I had read somewhere that creating many executor services consumes more system resources and might lead to performance degradation, and I've noticed we've done this elsewhere, but it hasn't seemed to cause performance trouble so far.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        this.pluginService.setInfo(this.customEndpointHostSpec.getHost(), endpointInfo, true);
+        this.infoChangedCounter.inc();
+
+        long elapsedTime = System.nanoTime() - start;
+        long sleepDuration = Math.min(0, this.refreshRateNano - elapsedTime);
+        TimeUnit.NANOSECONDS.sleep(sleepDuration);
+      }
+    } catch (InterruptedException e) {
+      LOGGER.info(Messages.get("CustomEndpointMonitorImpl.interrupted", new Object[]{ this.customEndpointHostSpec }));
+      Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
aaron-congo: This catch block is outside of the while loop so it will just return after this line even without an explicit return statement. If I add a return statement my IDE highlights it as being unnecessary.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: as mentioned above, the return statement isn't necessary because the monitor will already stop if an interrupted exception occurs. I believe an interrupted exception is only thrown if it is blocked (waiting or sleeping) though, which is why I'm checking the condition here.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Its definitely an exceptional case, which is why I log a warning about this above this line. I would normally throw a SQLException if this happened in the main thread, but this is happening in a monitor thread, so I'm not sure how to report the exception to the main thread.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: That was just from a copy-paste, I'll change to protected

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: It would allow them to see how many connections are being made to custom endpoints specifically. If you think its not valuable I'll remove it.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: I will change it to throw a SQLException if we timed out and there is not info in the cache. I cannot make it a TimeoutException because this method is called from connect and execute and both places do not have TimeoutException in their method signature.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: The problem with not waiting is that, if the custom endpoint info isn't discovered yet, there will be no disallowed hosts. Which means that failover and the read write splitting plugin may connect to hosts outside of the custom endpoint. It seems like the user would not want this to happen.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: Yeah looks like monitor.clear already closes the monitors, I'll remove this for loop

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
aaron-congo: I think it makes sense to use all hosts, the only thing updatedHosts is used for is to identify the writer, so it makes sense to include all hosts in case the subset of hosts doesn't include the writer. Below this we check if the identified writer is in the list of allowed hosts and throw an exception if it isn't

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -170,7 +179,7 @@ public void notifyNodeListChanged(final Map<String, EnumSet<NodeChangeOptions>>
   }
 
   private HostSpec getWriter() {
-    for (final HostSpec host : this.pluginService.getHosts()) {
+    for (final HostSpec host : this.pluginService.getAllHosts()) {
```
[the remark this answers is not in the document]
aaron-congo: Earlier when we were walking through the code trying to decide where to call getHosts vs getAllHosts we decided to use getAllHosts here because we are trying to identify the writer. Additionally, the beginning of this method checks whether we are using a writer cluster URL and returns early (before calling getWriter) if we are not. So we will not get to this line if we are connecting to a custom endpoint URL.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: I moved the parsing logic outside the computeIfAbsent lambda now, so it is now a SQLException instead of RuntimeException

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Do we want to continue monitoring even if there is a problem? I adjusted the code to continue monitoring if an exception occurs (unless its an interrupted exception), but I'm not sure if it makes sense or not.

wrapper/src/main/java/software/amazon/jdbc/util/SlidingExpirationCache.java
aaron-congo: This fix was merged into main in another PR, ended up using an ArrayList

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
sergiyvamz: InterruptedException is cough by try-catch block and that leads to stop of the thread. I'm still not sure we need this condition.

wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
aaron-congo: As discussed, this class no longer has references to classes in the customendpoint package

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: InterruptedExceptions are only thrown if the thread is interrupted while performing a blocking operation like sleep. Otherwise, if the thread is not blocking, the interrupted flag will be set but no interruption is thrown. If I removed this condition, the thread may continue running normally even after being interrupted.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,336 @@
...
+            }
+          },
+          CACHE_CLEANUP_RATE_NANO);
+
+  private static final Set<String> subscribedMethods =
+      Collections.unmodifiableSet(new HashSet<String>() {
+        {
+          addAll(SubscribedMethodHelper.NETWORK_BOUND_METHODS);
+          add("connect");
+          add("forceConnect");
```
sergiyvamz, line 74: I don't think we need forceConnect(). It's usually used by various monitoring threads and failover.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -0,0 +1,336 @@
...
+
+  /**
+   * If custom endpoint info does not exist for the current custom endpoint, waits a short time for the info to be
+   * made available by the custom endpoint monitor. This is necessary so that other plugins can rely on accurate custom
+   * endpoint info. Since custom endpoint monitors and information are shared, we should not have to wait often.
+   */
+  protected void waitForCustomEndpointInfo(CustomEndpointMonitor monitor) throws SQLException {
+    boolean hasCustomEndpointInfo = monitor.hasCustomEndpointInfo();
+
+    if (!hasCustomEndpointInfo) {
```
sergiyvamz, line 256: So we always wait if cache is empty? Is it so essential to wait for monitoring thread fetch custom endpoint definition? Can we make it configurable (wait/not wait)?

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: The problem with not waiting is that, if the custom endpoint info isn't discovered yet, there will be no disallowed hosts. Which means that failover and the read write splitting plugin may connect to hosts outside of the custom endpoint. It seems like the user would not want this to happen if they are enabling this plugin. I believe the user will not have to wait long or often, on my machine it takes ~260ms for the API call to complete and the user will only need to wait when info isn't in the cache already (cache entries last 5 minutes but will be extended by active monitors). If you want I can make it configurable, but if the wait is disabled the custom endpoint plugin may allow connections to instances outside the endpoint.

sergiyvamz: I'd prefer to delegate this decision to a user so a new configuration parameter would be great. Also I see that after calling `waitForCustomEndpointInfo()` the driver continues with `connectFunc` to get a connection. That means that custom endpoint info isn't directly required at this point.

aaron-congo: It might be required, there are two problem scenarios I can think of:
1. user has enableConnectFailover set to true, they try to connect but it fails and failover is kicked off, failover connects to a host outside of the custom endpoint info because we did not wait for the info to be found.
2. user connects successfully and then executes setReadOnly or hits failover while executing some other method, in both cases we could connect to an instance outside of the custom endpoint if we did not wait for the info to be found

sergiyvamz (approved):

PR #1131, opened 2024-09-27
fix: use the cluster URL as the default cluster ID

aaron-congo:
### Summary

fix: use the cluster URL as the default cluster ID

### Description

The RdsHostListProvider contains logic to use the cluster URL as a default cluster ID if the user passed a cluster URL, but there was a bug where this was not working correctly. The code was passing a string containing the host and port of the cluster URL, but the regex we use to identify whether a string is a cluster URL does not expect a port, so the logic was not correctly recognizing when cluster URLs were passed.


sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -197,7 +197,7 @@ protected void init() throws SQLException {
           this.isPrimaryClusterId = clusterSuggestedResult.isPrimaryClusterId;
         } else {
           final String clusterRdsHostUrl =
-              rdsHelper.getRdsClusterHostUrl(this.initialHostSpec.getHostAndPort());
```
ivankhoosty, line 200: so this broke 2.3.8, 2.3.9 and 2.4.0 as per https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1138 , maybe you should add a test ?

aaron-congo: Test added, thanks

PR #1133, opened 2024-09-27
failover2 on connect

sergiyvamz:
### Summary

Allows failover2 plugin to perform a failover during opening a new connection.

### Additional Reviewers

@karenc-bq 
@aaron-congo 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
@@ -43,6 +43,7 @@ import software.amazon.jdbc.plugin.failover.TransactionStateUnknownSQLException;
 import software.amazon.jdbc.plugin.staledns.AuroraStaleDnsHelper;
 import software.amazon.jdbc.targetdriverdialect.TargetDriverDialect;
 import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.PropertyUtils;
 import software.amazon.jdbc.util.RdsUrlType;
 import software.amazon.jdbc.util.RdsUtils;
 import software.amazon.jdbc.util.SqlState;
@@ -64,6 +65,8 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
   private static final Logger LOGGER = Logger.getLogger(FailoverConnectionPlugin.class.getName());
   private static final String TELEMETRY_WRITER_FAILOVER = "failover to writer node";
   private static final String TELEMETRY_READER_FAILOVER = "failover to replica";
+
+  private static final String INTERNAL_CONNECT_PROPERTY_NAME = "76c06979-49c4-4c86-9600-a63605b83f50";
 
   public static final AwsWrapperProperty FAILOVER_TIMEOUT_MS =
       new AwsWrapperProperty(
@@ -87,10 +90,18 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
           "random",
           "The strategy that should be used to select a new reader host while opening a new connection.");
 
+  public static final AwsWrapperProperty ENABLE_CONNECT_FAILOVER =
+      new AwsWrapperProperty(
+          "enableConnectFailover", "false",
+          "Enable/disable cluster-aware failover if the initial connection to the database fails due to a "
+              + "network exception. Note that this may result in a connection to a different instance in the cluster "
+              + "than was specified by the URL.");
+
   private static final Set<String> subscribedMethods =
       Collections.unmodifiableSet(new HashSet<String>() {
         {
           addAll(SubscribedMethodHelper.NETWORK_BOUND_METHODS);
+          add("connect");
           add("initHostProvider");
         }
       });
@@ -317,7 +328,6 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
    * @throws SQLException if an error occurs
    */
   protected void failover(final HostSpec failedHost) throws SQLException {
-    this.pluginService.setAvailability(failedHost.asAliases(), HostAvailability.NOT_AVAILABLE);
 
     if (this.failoverMode == FailoverMode.STRICT_WRITER) {
       failoverWriter();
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -317,7 +328,6 @@ protected <E extends Exception> void dealWithIllegalStateException(
    * @throws SQLException if an error occurs
    */
   protected void failover(final HostSpec failedHost) throws SQLException {
-    this.pluginService.setAvailability(failedHost.asAliases(), HostAvailability.NOT_AVAILABLE);
```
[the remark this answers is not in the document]
sergiyvamz: That line prevents an unavailable host gets available after 5 min. PluginService keeps node availability in cache. This line of code refreshed that cache and practically makes that node unavailable forever.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -606,4 +621,70 @@ protected boolean canDirectExecute(final String methodName) {
...
+      final JdbcCallable<Connection, SQLException> connectFunc)
+      throws SQLException {
+    return connectInternal(driverProtocol, hostSpec, props, isInitialConnection, connectFunc, false);
+  }
+
+  private Connection connectInternal(String driverProtocol, HostSpec hostSpec, Properties props,
+      boolean isInitialConnection, JdbcCallable<Connection, SQLException> connectFunc, boolean isForceConnect)
+      throws SQLException {
+
+    if (props.containsKey(INTERNAL_CONNECT_PROPERTY_NAME)) {
```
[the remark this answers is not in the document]
sergiyvamz: `forceConnect()` is for direct connection but it's not used in `failover2`. So `forceConnect()` makes not much sense to this case. Also, `failover2` uses regular `connect()` because `MonitoringHostListProvider` already verified topology so no need to force direct connect. And with a regular `connect()`, the call comes to `failover2` plugin again and our intention is to skip it.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -606,4 +621,70 @@ protected boolean canDirectExecute(final String methodName) {
...
+        .orElse(null);
+
+    if (hostSpecWithAvailability == null
+        || hostSpecWithAvailability.getAvailability() != HostAvailability.NOT_AVAILABLE) {
+
+      try {
+        conn = this.staleDnsHelper.getVerifiedConnection(isInitialConnection, this.hostListProviderService,
+            driverProtocol, hostSpec, props, connectFunc);
+      } catch (final SQLException e) {
+        if (isForceConnect || !this.shouldExceptionTriggerConnectionSwitch(e)) {
```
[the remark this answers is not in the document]
sergiyvamz: Fixed. Thank you!

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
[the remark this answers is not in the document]
aaron-congo: I see. Using the INTERNAL_CONNECT_PROPERTY_NAME property as a flag seems a bit hacky/strange to me but I can't think of an alternative at the moment.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -606,4 +621,69 @@ protected boolean canDirectExecute(final String methodName) {
...
+  public Connection connect(
+      final String driverProtocol,
+      final HostSpec hostSpec,
+      final Properties props,
+      final boolean isInitialConnection,
+      final JdbcCallable<Connection, SQLException> connectFunc)
+      throws SQLException {
+
+    if (!ENABLE_CONNECT_FAILOVER.getBoolean(props)) {
+      return connectFunc.call();
```
aaron-congo, line 635: Shouldn't we be connecting using getVerifiedConnection in this case?
```suggestion
      return this.staleDnsHelper.getVerifiedConnection(isInitialConnection, this.hostListProviderService,
            driverProtocol, hostSpec, props, connectFunc);
```

sergiyvamz: getVerifiedConnection() call already been made of a caller of connect() that is the same failover2 plugin. I don't think we need to execute it again.

aaron-congo: sorry are you saying that getVerifiedConnection has been previously called by another plugin or previously called by failover2?

sergiyvamz: failover2 `connect()` is called, then `getVerifiedConnection()` is called.
https://github.com/aws/aws-advanced-jdbc-wrapper/blob/b84d66c8eb6361e7b036e085ff27f02c1e91025b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java#L654

After that, if connection isn't established, a failover starts.

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/b84d66c8eb6361e7b036e085ff27f02c1e91025b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java#L664

And eventually it calls `pluginService.connect()`

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/b84d66c8eb6361e7b036e085ff27f02c1e91025b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java#L507

that will call failover2 `connect()` again. This call can be passed through to the next plugin since it's already been handled earlier.

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
aaron-congo: If I understand correctly this is explaining why we should use connectFunc.call instead of getVerifiedConnection when the props contain the INTERNAL_CONNECT_PROPERTY_NAME property, but this particular line is a different scenario: it is hit when connect failover is disabled. When connect failover is disabled we will never enter the failover logic below, so as the code currently is written getVerifiedConnection will never be called in this scenario

PR #1142, opened 2024-10-03
fix: SlidingExpirationCache and SlidingExpirationCacheWithCleanupThread

aaron-congo:
### Summary

fix: SlidingExpirationCache and SlidingExpirationCacheWithCleanupThread

### Description

Contains fixes for SlidingExpirationCache and SlidingExpirationCacheWithCleanupThread:
- in SlidingExpirationCache, synchronized shouldCleanup check and subsequent removal from cache. This prevents the scenario where a cache item is grabbed after shouldCleanup() returns true but before its removed from the cache.
- in SlidingExpirationCacheWithCleanupThread, some cleanup thread variables were static, which made it so that only one cache could create a cleanup thread. By removing the static keyword we ensure each cache has its own cleanup thread.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/SlidingExpirationCache.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/SlidingExpirationCache.java
@@ -16,7 +16,9 @@
 
 package software.amazon.jdbc.util;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/SlidingExpirationCache.java
```diff
@@ -118,9 +120,22 @@ protected void removeAndDispose(K key) {
   }
 
   protected void removeIfExpired(K key) {
-    final CacheItem cacheItem = cache.get(key);
-    if (cacheItem == null || cacheItem.shouldCleanup()) {
-      removeAndDispose(key);
+    // A list is used to store the cached item for later disposal since lambdas require references to outer variables
+    // to be final. This allows us to dispose of the item after it has been removed and the cache has been unlocked,
+    // which is important because the disposal function may be long-running.
+    final List<V> itemList = new ArrayList<>(1);
```
sergiyvamz, line 126: Why do you use List<> instead of a simple variable of type V?

aaron-congo: If I try to do that I get "Variable used in lambda expression should be final or effectively final". The solutions my IDE proposes are other making it an atomic reference or making it a list. This relates to [this discussion](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1122#discussion_r1767714929) on the custom endpoint PR.

sergiyvamz (approved):

PR #1147, opened 2024-10-08
fix - update limitless DB Shard Group DNS pattern

aaronchung-bitquill:
### Summary

fix - update limitless DB Shard Group DNS pattern

### Description

Change the limitless DNS Pattern from "limitless" to "dbshardgrp"

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
@@ -75,7 +75,7 @@ public class RdsUtils {
   private static final Pattern AURORA_DNS_PATTERN =
       Pattern.compile(
           "^(?<instance>.+)\\."
-              + "(?<dns>proxy-|cluster-|cluster-ro-|cluster-custom-|limitless-)?"
+              + "(?<dns>proxy-|cluster-|cluster-ro-|cluster-custom-|shardgrp-)?"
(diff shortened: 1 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
```diff
@@ -284,7 +284,7 @@ public String getRdsClusterHostUrl(final String host) {
     }
     final Matcher limitlessMatcher = AURORA_LIMITLESS_CLUSTER_PATTERN.matcher(host);
```
sergiyvamz, line 285: I'm not sure we need to include limitless cluster (that has quite unique feature) here. Let's discuss.

wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
```diff
@@ -90,14 +90,14 @@ public class RdsUtils {
   private static final Pattern AURORA_LIMITLESS_CLUSTER_PATTERN =
       Pattern.compile(
           "(?<instance>.+)\\."
-              + "(?<dns>limitless-)+"
+              + "(?<dns>shardgrp-)+"
               + "(?<domain>[a-zA-Z0-9]+\\.(?<region>[a-zA-Z0-9\\-]+)"
               + "\\.rds\\.(amazonaws\\.com(\\.cn)?|sc2s\\.sgov\\.gov|c2s\\.ic\\.gov))",
```
sergiyvamz, line 95: it should ends with "$"

sergiyvamz (approved):

PR #1152, opened 2024-10-16
fix: limitless plugin - add null checks

aaronchung-bitquill:

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
@@ -161,7 +161,7 @@ public class LimitlessConnectionPlugin extends AbstractConnectionPlugin {
         try {
           conn = connectFunc.call();
         } catch (final SQLException e) {
-          return retryConnectWithLeastLoadedRouters(limitlessRouters, props, conn, hostSpec);
+          return retryConnectWithLeastLoadedRouters(limitlessRouters, props, null, hostSpec);
         }
       }
       return conn;
@@ -174,7 +174,7 @@ public class LimitlessConnectionPlugin extends AbstractConnectionPlugin {
           HostRole.WRITER, RoundRobinHostSelector.STRATEGY_ROUND_ROBIN);
       LOGGER.fine(Messages.get(
           "LimitlessConnectionPlugin.selectedHost",
-          new Object[] {selectedHostSpec.getHost()}));
+          new Object[] {selectedHostSpec != null ? selectedHostSpec.getHost() : "null"}));
     } catch (SQLException e) {
       LOGGER.warning(Messages.get("LimitlessConnectionPlugin.errorSelectingRouter", new Object[] {e.getMessage()}));
       if (conn == null  || conn.isClosed()) {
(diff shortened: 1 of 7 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -161,7 +161,7 @@ private Connection connectInternalWithDialect(
         try {
           conn = connectFunc.call();
         } catch (final SQLException e) {
-          return retryConnectWithLeastLoadedRouters(limitlessRouters, props, conn, hostSpec);
+          return retryConnectWithLeastLoadedRouters(limitlessRouters, props, conn, hostSpec, connectFunc);
```
sergiyvamz, line 164: I'd explicitly pass null instead of conn. Just for readability.

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -238,8 +240,16 @@ private void initLimitlessRouterMonitorService() {
...
-  private Connection retryConnectWithLeastLoadedRouters(final List<HostSpec> limitlessRouters, final Properties props,
-      final Connection conn, final HostSpec hostSpec) throws SQLException {
+  private Connection retryConnectWithLeastLoadedRouters(
+      final List<HostSpec> limitlessRouters,
+      final Properties props,
+      final Connection inputConn,
+      final HostSpec hostSpec,
+      final JdbcCallable<Connection, SQLException> connectFunc) throws SQLException {
+    Connection conn = inputConn;
+    if (conn == null || conn.isClosed()) {
```
sergiyvamz, line 250: That'll try to establish a connection to the same node as the driver has just tried and failed. How this call be different?

sergiyvamz:
We also need a fix for monitor constructor. https://github.com/aws/aws-advanced-jdbc-wrapper/blob/7c5f391fec95f755d4a1b739e3d9763590b7859e/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java#L68

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
aaronchung-bitquill: Removed this and added more checks to throw exceptions when trying to use the connections.

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessConnectionPlugin.java
```diff
@@ -186,10 +186,12 @@ private Connection connectInternalWithDialect(
     try {
       return pluginService.connect(selectedHostSpec, props);
     } catch (SQLException e) {
-      LOGGER.fine(Messages.get(
-          "LimitlessConnectionPlugin.failedToConnectToHost",
-          new Object[] {selectedHostSpec.getHost()}));
-      selectedHostSpec.setAvailability(HostAvailability.NOT_AVAILABLE);
+      if (selectedHostSpec != null) {
```
sergiyvamz, line 189: I think it's always not null

sergiyvamz (approved):

PR #1169, opened 2024-10-28
fix: commit transaction after topology query (fixes #1168)

### LINKED ISSUE #1168: Unexpected Transaction State Error on Repeated setReadOnly Calls in AWS Aurora Wrapper
### Describe the bug

In an environment using the AWS wrapper driver, we've encountered an intermittent error where calling `setReadOnly` at the start or end of a transaction triggers the Aurora wrapper's topology, resulting in the error: "Cannot change transaction read-only property in the middle of a transaction."

### Expected Behavior

If not strictly necessary, we recommend that the `setReadOnly` function should avoid executing the topology call, or that the topology function be executed only after the connection is fully established.

failover,efm

### Current Behavior

Based on our direct analysis of the methods, we’ve inferred the following:

- Topology queries are intermittently executed at the end of a transaction.
- These queries are executed within the AWS-provided JDBC wrapper driver.
- The `setReadOnly` method in JDBC cannot be called during a transaction.

Given these points, we hypothesize that the error occurs in the following sequence:

1. Transaction ends.
2. `TransactionState` changes to `IDLE`.
3. Under certain conditions, a topology query is executed.
4. `TransactionState` changes back to `OPEN`.
5. `setReadOnly` method is called.
6. Error occurs.

### Reproduction Steps

When the `setReadOnly` function is executed repeatedly, the error described above occasionally occurs.

### Possible Solution

I’d like to propose that the Aurora wrapper driver don't trigger the topology call by default when `setReadOnly` is invoked.

Referencing the code in question: [SubscribedMethodHelper.java, line 59](https://github.com/aws/aws-advanced-jdbc-wrapper/blob/e390ac3065c12740eb610a22d8896b1f58604b0f/wrapper/src/main/java/software/amazon/jdbc/util/SubscribedMethodHelper.java#L59).

I'd be happy to create a PR to help fix this issue :)

### The AWS Advanced JDBC Driver version used

2.2.3

### JDK version used

correctto-17 17.0.9

macOS 15.0.1 (24A348)

finalchild:
### Summary

This PR tries to fix #1168, where `queryForTopology` automatically fired by FailOverPlugin opens a transaction and never closes, resulting in an unexpected change in the connection state.

### Description

If the connection doesn't auto-commit and wasn't in a transaction already, `queryForTopology` leaves a new transaction open. We have to either commit or roll back the transaction.

Note that this PR is a suggestion merely based on a speculation. I haven't tested that this works, nor do I fully understand the meaning of `this.hostListProviderService.isInTransaction()`. Please review and close/edit as you see fit.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/HostListProviderService.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/HostListProviderService.java
@@ -38,4 +38,6 @@ public interface HostListProviderService {
   Dialect getDialect();
 
   HostSpecBuilder getHostSpecBuilder();
+
+  boolean isInTransaction();
 }
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
@@ -360,12 +360,17 @@ public class RdsHostListProvider implements DynamicHostListProvider {
           new Object[] {e.getMessage()}));
     }
 
+    boolean shouldCommit = !conn.getAutoCommit() && !this.hostListProviderService.isInTransaction();
+
     try (final Statement stmt = conn.createStatement();
          final ResultSet resultSet = stmt.executeQuery(this.topologyQuery)) {
       return processQueryResults(resultSet);
     } catch (final SQLSyntaxErrorException e) {
       throw new SQLException(Messages.get("RdsHostListProvider.invalidQuery"), e);
     } finally {
+      if (shouldCommit) {
+        conn.commit();
+      }
       if (networkTimeout == 0 && !conn.isClosed()) {
         conn.setNetworkTimeout(networkTimeoutExecutor, networkTimeout);
       }


aaron-congo:
Hi @finalchild, I believe the fix for this has been merged already (see [here](https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1168#issuecomment-2477238166)). Are you okay with me closing this PR?

finalchild:
@aaron-congo Hi. I haven't got much time to investigate this issue deeper, but I'm not completely convinced that the released fix prevents the whole class of issue from occurring.

What if one calls PreparedStatement.getParameterMetaData while not in a transaction and auto-commit is false? The wrapper would try to update the topology, executing a query, and leave a transaction open. Later, when some state-sensitive methods like setReadOnly is called, an error will occur because a transaction is open, while 

I believe the only perfect solution is either closing the transaction later (like my patch), or temporarily turn off setAutoCommit in the first place.

I found this article which explains the behavior of setAutoCommit, which looks pretty harmless to save and recover later if there is no concurrency issues.
https://weinan.io/2017/05/21/jdbc-part5.html

sergiyvamz:
Hi @finalchild, thank you for taking the time to put this together and for the thoughtful follow-up discussion.

We've taken another look at this in the context of the current codebase, and we'd like to close this PR. The code this PR targets no longer exists in its original form: `queryForTopology` has been moved out of `RdsHostListProvider` into a `TopologyUtils` class, and the `HostListProviderService.isInTransaction()` method this PR introduced is no longer part of the design (the PR now also has merge conflicts against `main`). More importantly, topology refreshes are now handled by a dedicated background topology monitor that uses its own monitoring connection, configured via separate `monitoring-` properties and separate from the connection your application uses. Since the topology query never executes on the user's connection, it can't leave that connection's transaction state altered — which covers the `getParameterMetaData`/`setReadOnly` case you raised where autocommit is off.

We really appreciate the contribution and the discussion around it. If you do hit a case where a topology query still affects your application connection's transaction state on a current driver version, please open a new issue with reproduction details and we'll be glad to investigate.

PR #1174, opened 2024-11-01
fix: limitless - reduce extraneous conns made on first connection + s…

aaronchung-bitquill:
…hift connection logic from plugin to service class

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
@@ -16,7 +16,6 @@
 
 package software.amazon.jdbc.plugin.limitless;
 
-import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
@@ -27,11 +26,15 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.jdbc.AwsWrapperProperty;
+import software.amazon.jdbc.HostRole;
 import software.amazon.jdbc.HostSpec;
 import software.amazon.jdbc.PluginService;
+import software.amazon.jdbc.RoundRobinHostSelector;
+import software.amazon.jdbc.hostavailability.HostAvailability;
 import software.amazon.jdbc.util.Messages;
 import software.amazon.jdbc.util.SlidingExpirationCacheWithCleanupThread;
 import software.amazon.jdbc.util.Utils;
+import software.amazon.jdbc.wrapper.HighestWeightHostSelector;
 
 public class LimitlessRouterServiceImpl implements LimitlessRouterService {
   private static final Logger LOGGER =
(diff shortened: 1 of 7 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
```diff
@@ -96,37 +99,189 @@ public LimitlessRouterServiceImpl(
...
+      LOGGER.fine(Messages.get(
+          "LimitlessRouterServiceImpl.selectedHost",
+          new Object[] {selectedHostSpec != null ? selectedHostSpec.getHost() : "null"}));
+    } catch (SQLException e) {
+      retryConnectWithLeastLoadedRouters(context);
+      return;
+    }
+
+    try {
+      if (selectedHostSpec == null) {
```
sergiyvamz, line 150: This code should be out of try-final. At line 148?

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
```diff
@@ -96,37 +99,189 @@ public LimitlessRouterServiceImpl(
...
-  public List<HostSpec> forceGetLimitlessRoutersWithConn(
-      final Connection connection, final int hostPort, final Properties props) throws SQLException {
+  private void retryConnectWithLeastLoadedRouters(
+      final LimitlessConnectionContext context) throws SQLException {
+
+    int retryCount = 0;
+    final int maxRetries = LimitlessConnectionPlugin.MAX_RETRIES.getInteger(context.getProps());
+
+    while (retryCount++ < maxRetries) {
+      if (context.getLimitlessRouters()
```
sergiyvamz, line 179: We need a null check for getLimitlessRouters()

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
```diff
@@ -96,37 +99,189 @@ public LimitlessRouterServiceImpl(
...
+    int retryIntervalMs = LimitlessConnectionPlugin.GET_ROUTER_RETRY_INTERVAL_MILLIS.getInteger(context.getProps());
+    do {
+      try {
+        synchronouslyGetLimitlessRouters(context);
+        if (!Utils.isNullOrEmpty(context.getLimitlessRouters())) {
+          return;
+        }
+        Thread.sleep(retryIntervalMs);
+      } catch (final InterruptedException e) {
+        Thread.currentThread().interrupt();
```
sergiyvamz, line 249: return
?

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
```diff
@@ -96,37 +99,189 @@ public LimitlessRouterServiceImpl(
...
       }
 
+      if (context.getConnection() == null || context.getConnection().isClosed()) {
+        context.setConnection(context.getConnectFunc().call());
+      }
       final List<HostSpec> newLimitlessRouters =
-          this.queryHelper.queryForLimitlessRouters(connection, hostPort);
+          this.queryHelper.queryForLimitlessRouters(context.getConnection(), context.getHostSpec().getPort());
 
+      context.setLimitlessRouters(newLimitlessRouters);
```
sergiyvamz, line 281: if (!Utils.isNullOrEmpty(newLimitlessRouters)) {
...
}

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
sergiyvamz, line 179: Null check

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
aaronchung-bitquill: Added a line to throw an exception in case the it is either null or empty. 
Reason for this is so that `#synchronouslyGetLimitlessRoutersWithRetry` can catch it and retry. 
Otherwise if `context.limitlessRouters` is not null or empty for some reason, it will not retry.

sergiyvamz (approved):

PR #1175, opened 2024-11-01
fix: ensure custom endpoint monitor obeys refresh rate

aaron-congo:
#### Description

 - Math.min was mistakenly used where Math.max should have been used, resulting in the custom endpoint monitor never sleeping. This PR corrects the logic so that the monitor obeys sleeps according to the refresh rate
 - also includes some other minor fixes and cleans up the custom endpoint test

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
@@ -37,7 +37,6 @@ import software.amazon.jdbc.HostSpec;
 import software.amazon.jdbc.PluginService;
 import software.amazon.jdbc.util.CacheMap;
 import software.amazon.jdbc.util.Messages;
-import software.amazon.jdbc.util.RdsUtils;
 import software.amazon.jdbc.util.StringUtils;
 import software.amazon.jdbc.util.telemetry.TelemetryCounter;
 import software.amazon.jdbc.util.telemetry.TelemetryFactory;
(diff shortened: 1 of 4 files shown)


wrapper/src/test/java/integration/container/tests/CustomEndpointTest.java
```diff
@@ -80,139 +78,123 @@
...
-        client.deleteDBClusterEndpoint((builder) -> builder.dbClusterEndpointIdentifier(endpointId));
-      } catch (DbClusterEndpointNotFoundException e) {
-        // Custom endpoint already does not exist - do nothing.
-      }
-    }
-
-    waitUntilEndpointsDeleted(client);
-  }
-
-  private static void waitUntilEndpointsDeleted(RdsClient client) {
```
aaron-congo, line 126: This used to be called after deleting the test custom endpoint after all tests in the file had been completed. I removed it to save some time. Each custom endpoint will have a unique ID, so even though other tests could start running while the custom endpoint is still being deleted, no other tests should be require it to be fully deleted before running. Let me know if you think I should add it back

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -188,13 +186,18 @@ public void run() {
         }
       }
     } catch (InterruptedException e) {
-      LOGGER.info(Messages.get("CustomEndpointMonitorImpl.interrupted", new Object[]{ this.customEndpointHostSpec }));
+      LOGGER.info(
```
sergiyvamz, line 189: nit: Log level fine is probably better in this case.

PR #1182, opened 2024-11-07
fix: ensure abort interrupts running queries

### LINKED ISSUE #1173: Expected Behavior of Abort with the AWS JDBC Wrapper
I'm developing an application that connects to PostgreSQL via JDBC using Spring Boot and HikariCP.

When attempting to stop the application (sending SIGINT) while a long-running query is active, it seems that HikariCP waits for the query to complete before aborting the connection. This behavior appears to be related to the AWS JDBC Wrapper's implementation:
(https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main/wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java#L186)

In contrast, when using only the standard PostgreSQL driver, the connection is forcibly terminated even if a long-running query is in progress.

Question:
Is there a configuration or method to avoid this behavior?
According to the java.sql.Connection#abort specification, the connection should be forcibly terminated.

Thank you.

aaron-congo:
### Description
- fixes #1173 
- abort was not interrupting queries because the plugin manager was locked during query execution and abort was not included in ASYNCHRONOUS_METHODS. This PR fixes this problem by allowing abort to be executed asynchronously
- after allowing asynchronous abort, the exception triggered by abort was triggering failover. To avoid this, closedExplicitly is set when abort is called and checked in shouldExceptionTriggerConnectionSwitch

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/AsynchronousMethodsHelper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/AsynchronousMethodsHelper.java
@@ -22,6 +22,7 @@ import java.util.List;
 public class AsynchronousMethodsHelper {
   public static final List<String> ASYNCHRONOUS_METHODS = Arrays.asList(
       "Statement.cancel",
-      "PreparedStatement.cancel"
+      "PreparedStatement.cancel",
+      "Connection.abort"
   );
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/#semantic-versioning-200).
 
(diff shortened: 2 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/AsynchronousMethodsHelper.java
```diff
@@ -22,6 +22,7 @@
 public class AsynchronousMethodsHelper {
   public static final List<String> ASYNCHRONOUS_METHODS = Arrays.asList(
       "Statement.cancel",
-      "PreparedStatement.cancel"
+      "PreparedStatement.cancel",
+      "Connection.abort"
```
aaron-congo, line 26: Should Connection.close be added here as well?

sergiyvamz: No, I'd keep "abort" in the list, not "close".

sergiyvamz (approved):

Issue #1181, opened 2024-11-07
ConfigurationProfileBuilder withAwsCredentialsProviderHandler is not working


ivankhoosty:
### Describe the bug

I wanted to set an (existing) awsCredentialsProvider bean for the driver to use, rather than fallback to DefaultCredentialsProvider, and it didn't work. Looking at the code, ConfigurationProfileBuilder withAwsCredentialsProviderHandler never calls AwsCredentialsManager.setCustomHandler to actually use the provided AwsCredentialsProviderHandler.

### Expected Behavior

ConfigurationProfileBuilder withAwsCredentialsProviderHandler should make sure that AwsCredentialsProviderHandler with awsCredentialsProvider is used by the driver in aws sdk calls.

### What plugins are used? What other connection properties were set?

ConnectTimeConnectionPluginFactory, ExecutionTimeConnectionPluginFactory, AuroraInitialConnectionStrategyPluginFactory, AuroraConnectionTrackerPluginFactory, FailoverConnectionPluginFactory, IamAuthConnectionPluginFactory

### Current Behavior

 ConfigurationProfileBuilder.get().withAwsCredentialsProviderHandler((hostSpec, props) -> awsCredentialsProvider) isn't doing anything, so it's necessary to call: AwsCredentialsManager.setCustomHandler((hostSpec, props) -> awsCredentialsProvider);

### Reproduction Steps

ConfigurationProfileBuilder.get().withAwsCredentialsProviderHandler((hostSpec, props) -> awsCredentialsProvider)

then observe that this awsCredentialsProvider is never used in aws sdk calls.

### The AWS Advanced JDBC Driver version used

2.5.2

### JDK version used

21.0.1

### Operating System and version

unix

aaron-congo:
Hi @ivankhoosty ,

We have just merged in a fix. Could you kindly checkout our latest [snapshot](https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#using-a-snapshot-of-the-driver) build and let us know if the issue still persists?

Thank you!

aaron-congo:
Hi @ivankhoosty,

The fix is now available in the recently released v2.5.3 of the driver. Since the fix is now released I will be closing this ticket. Please feel free to re-open or create a new ticket if you encounter any issues.

Thank you!

ivankhoosty:
[...]

PR #1183, opened 2024-11-07
fix: use the AwsCredentialsProviderHandler from the ConfigurationProfile when it is defined

aaron-congo:
### Description

- Addresses #1181
- This PR fixes a bug where passing an AwsCredentialsProviderHandler to a ConfigurationProfile had no effect because it was never used anywhere.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
@@ -49,6 +49,7 @@ import software.amazon.jdbc.JdbcCallable;
 import software.amazon.jdbc.PluginService;
 import software.amazon.jdbc.PropertyDefinition;
 import software.amazon.jdbc.authentication.AwsCredentialsManager;
+import software.amazon.jdbc.profile.ConfigurationProfile;
 import software.amazon.jdbc.util.Messages;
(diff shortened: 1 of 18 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -175,7 +176,12 @@ public AwsSecretsManagerConnectionPlugin(final PluginService pluginService, fina
 
     this.secretKey = Pair.of(secretId, region);
 
+    ConfigurationProfile profile = this.pluginService.getConfigurationProfile();
```
aaron-congo, line 179: Does this block need to be added to the OktaAuth or FederatedAuth plugins? I noticed they use credentials providers, but they do not use AwsCredentialsManager, instead they use a CredentialsProviderFactory. The only implementation o f CredentialsProviderFactory always uses AnonymousCredentialsProvider. Should these plugins be using the credentials provider defined by either the ConfigurationProfile or the AwsCredentialsManager?

aaron-congo: talked with Karen and Aaron and they said Okta and Federated Auth should not be using the custom handler

sergiyvamz (approved):

PR #1192, opened 2024-11-15
test: do not set instance class identifier when creating a limitless cluster

aaron-congo:
### Description

The instance class we usually set for integration test clusters is not compatible with the latest PG version, which is limitless. This PR avoids setting the instance class if the cluster version includes the substring "limitless", so that we will not get an error for those integration test runs.


karenc-bq (approved):

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -276,16 +276,21 @@ public String createAuroraCluster() throws InterruptedException {
...
+          .dbClusterIdentifier(dbIdentifier)
+          .dbInstanceIdentifier(instanceName)
+          .engine(dbEngine)
+          .engineVersion(dbEngineVersion)
+          .publiclyAccessible(true)
+          .tags(testRunnerTag);
+
+      // The latest PG version is limitless, which is not compatible with dbInstanceClass (db.r5.large), so we will
+      // leave the instance class unspecified in this scenario.
+      if (!dbEngineVersion.contains("limitless")) {
```
sergiyvamz, line 289: Can we create an internal boolean flag isEngineLimitless and use it here? We can set this flag in createCluster()

aaron-congo:
After further investigation, it seems you cannot create a limitless cluster via the API (see docs [here](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/limitless-create-cluster.html)). To get the integration tests working again, I'll open up a separate PR that selects the latest non-limitless version to run integration tests against. We can decide if we want to test against a limitless cluster separately.

PR #1197, opened 2024-11-21
test: exclude limitless versions when selecting latest version for testing

aaron-congo:
### Description

- our test logic executes a DescribeDbEngineVersions request to select the latest Aurora version to run integration tests against. In the past this has selected regular Aurora versions, but recently the latest version returned by the API is a limitless version. This PR fixes the problem by filtering out limitless versions. If we decide we want to test against limitless later, we should add a separate task/workflow.

### DIFF
--- a/wrapper/src/test/java/integration/util/AuroraTestUtility.java
+++ b/wrapper/src/test/java/integration/util/AuroraTestUtility.java
@@ -1415,6 +1415,7 @@ public class AuroraTestUtility {
 
   public String getLatestVersion(String engine) {
     return getEngineVersions(engine).stream()
+        .filter(version -> !version.contains("limitless"))
         .sorted(Comparator.reverseOrder())
         .findFirst()
         .orElse(null);


wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -1415,6 +1415,7 @@ public List<String> getEngineVersions(String engine) {
 
   public String getLatestVersion(String engine) {
     return getEngineVersions(engine).stream()
+        .filter(version -> !version.contains("limitless"))
```
aaron-congo, line 1418: Do we want to filter out other versions that are suffixed in case another non-regular version is added in the future? Or would we prefer to deal with that later? If we decide on the former I will change to the code below. Note that currently the only version that contains a "-", and the only non-regular version, is "16.4-limitless"
```suggestion
        .filter(version -> !version.contains("limitless"))
```

karenc-bq (approved):

wrapper/src/test/java/integration/util/AuroraTestUtility.java
sergiyvamz: I'd prefer a strict condition with "-limitless"

sergiyvamz (approved):

PR #1202, opened 2024-11-26
fix: add writer custom endpoint type

karenc-bq:
### Summary

The AWS Java SDK may return WRITER:
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/rds/model/CreateDbClusterEndpointResponse.html#customEndpointType()

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointRoleType.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointRoleType.java
@@ -17,10 +17,10 @@
 package software.amazon.jdbc.plugin.customendpoint;
 
 /**
- * Enum representing the possible roles of instances specified by a custom endpoint. Note that, currently, it is not
- * possible to create a WRITER custom endpoint.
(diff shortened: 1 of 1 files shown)


aaron-congo (commented):
The SDK page you linked does seem to indicate that the WRITER type is possible, but [this page](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Endpoints.Custom.Considerations.html) seems to contradict that:
"Each custom endpoint has an associated type that determines which DB instances are eligible to be associated with that endpoint. Currently, the type can be READER or ANY."

Do we want to add this anyways?

karenc-bq:
> The SDK page you linked does seem to indicate that the WRITER type is possible, but [this page](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Endpoints.Custom.Considerations.html) seems to contradict that: "Each custom endpoint has an associated type that determines which DB instances are eligible to be associated with that endpoint. Currently, the type can be READER or ANY."
> 
> Do we want to add this anyways?

I am leaning towards us matching with whatever the API returns so we are not caught off guard when WRITER type endpoints are supported.

aaron-congo (approved):

PR #1221, opened 2024-12-13
fix: avoid setting ignoreNewTopologyRequestsEndTimeNano on initial connection

aaron-congo:
### Description

- fixes a bug where the failover2 plugin occasionally incorrectly threw a FailoverFailedSQLException. Before these changes, there was a chance that ignoreNewTopologyRequestsEndTimeNano would be set to a positive value on initial connection when the writer is first being verified. If failover occurred after initial connection but before this duration was over, the failover plugin's request for new topology would be ignored and the old cached topology returned instead. Since the old topology still indicated the old writer as the current writer, the failover plugin was incorrectly trying to connect to the old failed writer. This connection attempt would fail and the plugin would throw a FailoverFailedSQLException. 
  - to address this problem, we now initialize ignoreNewTopologyRequestsEndTimeNano to a value of -1. The first time that the writer is verified (during initial connection), we set it to a value of 0. Any subsequent events where the writer is verified indicate successful failover, so ignoreNewTopologyRequestsEndTimeNano is updated to a positive value.
- also fixed a bug where ignoreNewTopologyRequestsEndTimeNano was being set to a positive value before the topology was updated. This could result in topology requests returning stale incorrect topology if the request came in after ignoreNewTopologyRequestsEndTimeNano was set but before the topology was updated.
- also fixed a bug in failover 1 where ReconnectToWriterHandler was incorrectly failing when a single instance cluster failover occurred
- also added and cleaned up logs for the failover2 plugin to make future debugging easier


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -448,14 +462,22 @@ protected List<HostSpec> openAnyConnectionAndUpdateTopology() {
         try {
           if (!StringUtils.isNullOrEmpty(this.getWriterNodeId(this.monitoringConnection.get()))) {
             this.isVerifiedWriterConnection = true;
+            wasWriterVerified = true;
```
sergiyvamz, line 465: It seems that this.isVerifiedWriterConnection does the same

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -483,8 +483,14 @@ protected void failoverWriter() throws SQLException {
           .findFirst()
           .orElse(null);
 
+      if (writerCandidate == null) {
+        this.failoverWriterFailedCounter.inc();
+        LOGGER.severe(Utils.logTopology(updatedHosts, Messages.get("Failover.noWriterHost")));
+        throw new FailoverFailedSQLException(Messages.get("Failover.unableToConnectToWriter"));
```
sergiyvamz, line 489: Should we use this message instead? 
Messages.get("Failover.noWriterHost")

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
aaron-congo: Sure, I updated the other FailoverFailedSQLException messages to match the logs I added as well so that they are more specific.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
aaron-congo: the difference is that `isVerifiedWriterConnection` could be set to true by this thread or by the node monitor threads. `wasWriterVerified` is only set to true if this thread verifies the writer. I renamed it to `writerVerifiedByThisThread` to be clearer.

PR #1227, opened 2024-12-18
fix: various reader failover fixes

aaron-congo:
### Description
- failover1 fixes:
  - use query to verify host role instead of topology, since server-side topology may not be updated yet.
  - include the original writer as a candidate when failoverMode=strict-reader, since it is likely now a reader. The host role will be verified after connecting.
  - in the ClusterAwareReaderFailoverHandler we pass a variable indicating whether failoverMode=strict-reader, but before these changes the variable was incorrectly passed before failoverMode had been initialized so it was getting the wrong value.
- failover2 fixes:
  - fixed a bug where reader failover was failing because the plugin assumed that the topology returned by forceRefreshHostList was not stale. In practice, the topology may not be stale, but it likely is because we do not wait for it to get updated before starting reader failover. To solve this, reader failover now tries to connect to all nodes in the topology. If failoverMode=strict-reader, the plugin will query the server to verify the host's role.
  - also fixed a bug where the original writer was only attempted once, which was causing failover failures for 2-instance clusters with failoverMode=strict-reader because it takes some time for the original writer to recover. All nodes will now be attempted multiple times until we hit the failover timeout.
- telemetry fixes (failover1 and failover2):
  - fixed a bug where telemetryContext.setSuccess and telemetryContext.setException were not called when failover succeeded
  - fixed a bug where failoverReaderFailedCounter was increased twice instead of once when failover failed

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
@@ -578,22 +578,6 @@ public class FailoverConnectionPlugin extends AbstractConnectionPlugin {
     } else {
       failoverReader(failedHost);
     }
-
-    if (isInTransaction || this.pluginService.isInTransaction()) {
-      if (this.pluginManagerService != null) {
-        this.pluginManagerService.setInTransaction(false);
-      }
-      // "Transaction resolution unknown. Please re-configure session state if required and try
-      // restarting transaction."
-      final String errorMessage = Messages.get("Failover.transactionResolutionUnknownError");
-      LOGGER.info(errorMessage);
-      throw new TransactionStateUnknownSQLException();
-    } else {
-      // "The active SQL connection has changed due to a connection failure. Please re-configure
-      // session state if required. "
-      LOGGER.severe(() -> Messages.get("Failover.connectionChangedError"));
-      throw new FailoverSuccessSQLException();
-    }
   }
 
   protected void failoverReader(final HostSpec failedHostSpec) throws SQLException {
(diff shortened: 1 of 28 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -364,12 +366,10 @@ private void initSettings() {
...
   }
 
-  protected void initFailoverMode() {
+  protected void initFailoverMode(final String initialUrl) {
     if (this.rdsUrlType == null) {
       this.failoverMode = FailoverMode.fromValue(FAILOVER_MODE.getString(this.properties));
-      final HostSpec initialHostSpec = this.hostListProviderService.getInitialConnectionHostSpec();
-      this.rdsUrlType = this.rdsHelper.identifyRdsType(initialHostSpec.getHost());
-
+      this.rdsUrlType = this.rdsHelper.identifyRdsType(ConnectionUrlParser.parseHost(initialUrl));
```
sergiyvamz, line 372: Why are we changing it? I think initial host works well. Initial host may have a list of urls. Does ConnectionUrlParser.parseHost() supports it?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -342,85 +345,110 @@ protected void failoverReader() throws SQLException {
...
+                    new ArrayList<>(remainingHosts),
+                    HostRole.READER,
+                    this.failoverReaderHostSelectorStrategySetting);
+          } catch (UnsupportedOperationException | SQLException ex) {
+            // can't use selected strategy to get a reader host
+            LOGGER.finest("Error: " + ex.getMessage());
+            break;
+          }
+
+          if (readerCandidate == null) {
```
sergiyvamz, line 392: Why not to continue with other possible hosts in remainingHosts?

sergiyvamz (approved):

PR #1238, opened 2025-01-08
improve forceConnect pipeline

sergiyvamz:
### Summary

improve forceConnect pipeline:
- review implemented forceConnect() for existing plugins
- keep forceConnect() for plugins related to db authentication; remove forceConnect() for other plugins
- allows to skip a particular plugin while calling connect() or forceConnect() pipeline

### Additional Reviewers

@karenc-bq 
@aaronchung-bitquill 
@aaron-congo 


wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
```diff
@@ -45,7 +46,7 @@
 import software.amazon.jdbc.util.telemetry.TelemetryFactory;
 import software.amazon.jdbc.util.telemetry.TelemetryGauge;
 
-public class IamAuthConnectionPlugin extends AbstractConnectionPlugin {
+public class IamAuthConnectionPlugin extends AbstractConnectionPlugin implements AuthenticationConnectionPlugin {
```
aaronchung-bitquill, line 49: I think the FederatedAuthPlugin should also implement AuthenticationConnectionPlugin

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -259,19 +272,28 @@ protected <T, E extends Exception> PluginChainJdbcCallable<T, E> makePluginChain
...
-          pluginSubscribedMethods.contains(ALL_METHODS)
-              || pluginSubscribedMethods.contains(methodName);
+          (!useForceConnectPipeline || (plugin instanceof AuthenticationConnectionPlugin))
+          && (pluginSubscribedMethods.contains(ALL_METHODS) || pluginSubscribedMethods.contains(methodName));
 
       if (isSubscribed) {
         if (pluginChainFunc == null) {
-          pluginChainFunc = (pipelineFunc, jdbcFunc) ->
+          // This case is for DefaultConnectionPlugin that always terminates the list of plugins.
+          // Default plugin couldn't be skipped.
```
aaronchung-bitquill, line 281:
```suggestion
          // Default plugin can't be skipped.
```

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -259,19 +272,28 @@ protected <T, E extends Exception> PluginChainJdbcCallable<T, E> makePluginChain
       final Set<String> pluginSubscribedMethods = plugin.getSubscribedMethods();
       final String pluginName = pluginNameByClass.getOrDefault(plugin.getClass(), plugin.getClass().getSimpleName());
       final boolean isSubscribed =
-          pluginSubscribedMethods.contains(ALL_METHODS)
-              || pluginSubscribedMethods.contains(methodName);
+          (!useForceConnectPipeline || (plugin instanceof AuthenticationConnectionPlugin))
```
aaron-congo, line 275: if I understand correctly, only AuthenticationConnectionPlugins will potentially be subscribed to forceConnect. This means that for all non-auth plugins we can:
- remove the forceConnect method
- remove "forceConnect" from the subscribed methods if it is explicitly included
- move the implementation of connectInternal() connect() and then get rid of connectInternal()

It looks like this PR only removes forceConnect from AuroraInitialConnectionPlugin, can you update the PR with the above suggestions for non-auth plugins?

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -54,7 +55,7 @@
  * This connection plugin will always be the last plugin in the connection plugin chain, and will
  * invoke the JDBC method passed down the chain.
  */
-public final class DefaultConnectionPlugin implements ConnectionPlugin {
+public final class DefaultConnectionPlugin implements ConnectionPlugin, AuthenticationConnectionPlugin {
```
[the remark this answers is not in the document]
aaronchung-bitquill: Alternatively, we can rename `AuthenticationConnectionPlugin` to something more appropriate to what it is used for.

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -259,19 +272,28 @@ protected <T, E extends Exception> PluginChainJdbcCallable<T, E> makePluginChain
       final Set<String> pluginSubscribedMethods = plugin.getSubscribedMethods();
```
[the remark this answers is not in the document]
sergiyvamz: Such approach can work for the first call of makePluginChainFunc(). Entire plugin chain lambda is cached after and never re-created again. That means that plugin check should be incorporated into plugin chain lambda and should work for every case. That also means that pluginToSkip should be a parameter for plugin chain lambda (PluginChainJdbcCallable<>.call(...))

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -259,19 +272,28 @@ protected <T, E extends Exception> PluginChainJdbcCallable<T, E> makePluginChain
...
+            if (pluginToSkip == plugin) {
+              return finalPluginChainFunc.call(pipelineFunc, jdbcFunc, pluginToSkip);
+            } else {
+              return executeWithTelemetry(
+                  () -> pipelineFunc.call(
+                      plugin,
+                      () -> finalPluginChainFunc.call(pipelineFunc, jdbcFunc, pluginToSkip)),
                   pluginName);
+            }
+          };
```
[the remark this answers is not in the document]
sergiyvamz: pluginToSkip should be checked on every level of plugin chain lambda. If condition's met, then a particular plugin should be skipped and next level lambda should be called instead. Such logic is missed in your proposed code.

wrapper/src/main/java/software/amazon/jdbc/plugin/dev/DeveloperConnectionPlugin.java
```diff
@@ -24,13 +24,15 @@
...
+import software.amazon.jdbc.AuthenticationConnectionPlugin;
 import software.amazon.jdbc.HostSpec;
 import software.amazon.jdbc.JdbcCallable;
 import software.amazon.jdbc.plugin.AbstractConnectionPlugin;
 import software.amazon.jdbc.util.StringUtils;
 import software.amazon.jdbc.util.WrapperUtils;
 
-public class DeveloperConnectionPlugin extends AbstractConnectionPlugin implements ExceptionSimulator {
+public class DeveloperConnectionPlugin extends AbstractConnectionPlugin
+    implements AuthenticationConnectionPlugin, ExceptionSimulator {
```
aaron-congo, line 35: Similar to my previous comment, it seems a bit misleading to label this as an Authentication plugin when it doesn't actually provide auth functionality. Can we similarly either remove this and explicitly check for this plugin in ConnectionPluginManager#makePluginChainFunc or do something like rename AuthenticationConnectionPlugin to ForceConnectPlugin

aaron-congo:
`FastestResponseStrategyPlugin` and `BenchmarkPlugin` still have a forceConnect method/implementation but will be excluded from the forceConnect pipeline with these changes. If they should be in the forceConnect pipeline, can we check for them in ConnectionPluginManager#makePluginChainFunc similarly to DefaultConnectionPlugin? And if they should not be in the forceConnect pipeline, can we remove their forceConnect/connectInternal methods?

aaron-congo (approved):

PR #1246, opened 2025-01-15
fix: failover fixes and fix flaky failover tests

aaron-congo:
### Summary

- a bug was occurring where our code was incorrectly concluding that the new writer host after failover was not allowed. This was happening because we were comparing HostSpec details between the new writer host and the allowed hosts from the topology to determine whether the new writer was allowed. When failover occurs the role and availability of the writer changes, so the old code could incorrectly conclude that the new writer was not in the list of allowed hosts because of a mismatch in host roles or availability. This issue is solved by comparing host names instead of HostSpec details when determining if the new writer is in the list of allowed hosts.
- in reader failover for the failover 1 plugin, we submit an internal failover task that forms a list of hosts to attempt in order of priority. If failover mode is STRICT_READER we have to verify that the host is a reader once connected. Before these changes, this verification was done after the reader failover task was returned from the internal failover task. If the host returned from the internal failover task ended up being a writer we would submit a new internal failover task. The problem with this approach is that hosts that are beneath the writer in the priority list will no longer be attempted. Each time we submit an internal failover task it may return the writer and we would have to try again. These changes move the reader verification check to the ConnectionAttemptTask thread so that, if we are in STRICT_READER mode, the internal failover task only returns a result if it has already been verified as a reader.
- also contains the following test fixes:
  - our integration tests were experiencing various flaky failures. Most failures were related to toxiproxy endpoints being unexpectedly down. These changes update the testcontainers dependencies and add a check for the toxiproxy endpoints to makeSureAllInstancesUp to hopefully aleviate these failures. Note that the testcontainers dependencies should all have the same version in the future.
  - another flaky failure that we saw was an NPE in the topology monitor when attempting to access the first column of the topology query result using mysql. We believe this may be occurring because the server may not properly return the topology query results if it is failing over. To guard against this we add a try-catch to the topology query result processing and checking the ResultSet column count before attempting to process the result. If it doesn't return the correct results we will try the topology again later.
  - the old code did not close and clear the topology monitors or topology node monitors. This could cause test failures between two failover tests because the old test would set the ignoreNewTopologyRequest for the monitor, then the new test would grab the same monitor and incorrectly ignore the topology request when its own failover process was initiated. This caused the new test to incorrectly rely on the old cached topology, so it would end up incorrectly concluding that the old writer was the new writer and then would fail to connect to it. This issue is fixed by clearing the topology monitors between test runs so that each test has to create its own topology monitors.
  - also changed some expected exceptions to be more specific, so that the test would not incorrectly pass when an unexpected SQLException subclass was thrown.


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsHostListProvider.java
```diff
@@ -85,17 +85,8 @@ public MonitoringRdsHostListProvider(
         CLUSTER_TOPOLOGY_HIGH_REFRESH_RATE_MS.getLong(this.properties));
   }
 
-  /**
-   * Clear topology cache for the current cluster.
-   */
-  public void clear() {
```
aaron-congo, line 91: intellij was giving me a warning that this implementation is the exact same as the superclass implementation so it can be removed

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsHostListProvider.java
```diff
@@ -85,17 +85,8 @@ public MonitoringRdsHostListProvider(
...
-    topologyCache.remove(this.clusterId);
-  }
-
   public static void clearCache() {
-    topologyCache.clear();
```
aaron-congo, line 96: intellij was giving me a warning that this implementation is the same as the superclass clearAll implementation

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -644,17 +647,34 @@ protected String getSuggestedWriterNodeId(final Connection connection) throws SQ
...
       final ResultSet resultSet,
       final String suggestedWriterNodeId) throws SQLException {
 
     final HashMap<String, HostSpec> hostMap = new HashMap<>();
 
+    int expectedColumnCount = 4;
+    int actualColumnCount = resultSet.getMetaData().getColumnCount();
+    if (actualColumnCount < expectedColumnCount) {
+      // We expect at least 4 columns. Note that the server may return 0 columns if failover has occurred.
+      LOGGER.fine(
```
sergiyvamz, line 660: finest() ?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -644,17 +647,34 @@ protected String getSuggestedWriterNodeId(final Connection connection) throws SQ
...
     // Data is result set is ordered by last updated time so the latest records go last.
     // When adding hosts to a map, the newer records replace the older ones.
     while (resultSet.next()) {
-      final HostSpec host = createHost(resultSet, suggestedWriterNodeId);
-      hostMap.put(host.getHost(), host);
+      try {
+        final HostSpec host = createHost(resultSet, suggestedWriterNodeId);
+        hostMap.put(host.getHost(), host);
+      } catch (Exception e) {
+        LOGGER.fine(
```
sergiyvamz, line 674: finest() ?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -403,11 +389,43 @@ public ReaderFailoverResult call() {
 
         final Connection conn = pluginService.forceConnect(this.newHost, copy);
         pluginService.setAvailability(this.newHost.asAliases(), HostAvailability.AVAILABLE);
+
+        if (enableFailoverStrictReader) {
```
sergiyvamz, line 393: I'd suggest to add enableFailoverStrictReader to a constructor parameters.

wrapper/src/main/java/software/amazon/jdbc/util/Utils.java
```diff
@@ -26,6 +26,16 @@ public static boolean isNullOrEmpty(final Collection<?> c) {
     return c == null || c.isEmpty();
   }
 
+  public static boolean containsHost(final List<HostSpec> hosts, String host) {
+    for (final HostSpec hostSpec : hosts) {
+      if (hostSpec.getHost().equals(host)) {
```
sergiyvamz, line 31: I believe that the port should be verified as well.

sergiyvamz (approved):

PR #1251, opened 2025-01-21
test: fix multi-az integration test failures

aaron-congo:
### Description

- our multi-az integration tests were failing because our tests rely on the cluster being fully up and stable, but the old writer takes a long time to stabilize after failover. Note that aurora clusters do not have this issue. This PR fixes the issue by simulating failover using toxiproxy when testing against a multi-az environment. Note that this means the multi-az failover tests will test the scenario where the old writer remains the new writer after failover, since we are not simulating the topology query.
- before these changes our multi-az tests were running against a constant version, regardless of whether the requested version was "default" or "latest". This PR also updates the multi-az createCluster request to use "default" or "latest" like the aurora tests do


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -653,14 +653,9 @@ protected String getSuggestedWriterNodeId(final Connection connection) throws SQ
 
     final HashMap<String, HostSpec> hostMap = new HashMap<>();
 
-    int expectedColumnCount = 4;
```
aaron-congo, line 656: Expecting a column count of 4 was causing failovers against multi-az clusters because the multi-az topology query expects a different number of columns (3)

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -1166,15 +1214,27 @@ public String hostToIP(String hostname) {
   }
 
   public String hostToIP(String hostname, boolean fail) {
-    try {
-      final InetAddress inet = InetAddress.getByName(hostname);
-      return inet.getHostAddress();
-    } catch (UnknownHostException e) {
-      if (fail) {
-        fail("The IP address of host " + hostname + " could not be determined");
+    int remainingTries = 5;
```
aaron-congo, line 1217: I was seeing occasional test failures where the host could not be determined, presumably this is an intermittent problem and it doesn't happen so hopefully the retry here will get rid of the issue

sergiyvamz (approved):

PR #1252, opened 2025-01-22
chore: refactor AuroraTestUtility

aaron-congo:
### Description

- removes most instance fields in AuroraTestUtility, in particular the cluster details. Some functions were relying on these details being passed in and others were relying on internal state set by previous functions, making it difficult to understand which details were in use
- addresses some small IDE warnings/suggestions in the code


wrapper/src/test/java/integration/container/tests/AutoscalingTests.java
```diff
@@ -127,8 +128,11 @@ public void test_pooledConnectionAutoScaling_setReadOnlyOnOldConnection()
       }
 
       final Connection newInstanceConn;
+      final String instanceClass =
+          auroraUtil.getDbInstanceClass(TestEnvironment.getCurrent().getInfo().getRequest());
       final TestInstanceInfo newInstance =
-          auroraUtil.createInstance("auto-scaling-instance");
+          auroraUtil.createInstance(instanceClass, "auto-scaling-instance");
```
aaron-congo, line 134: Before these changes, the new instance was added to the TestEnvironment in a convoluted way: an earlier call to AuroraTestUtility would pass in the TestEnvironment instance list, which the class would then assign to an internal instance variable. Then, in this call, the instance list would be updated. In this PR we removed the internal instance variable. Should I change the method signature to require the TestEnvironment instance list to be passed in so that we can update it inside the function instead of relying on callers to update it themselves? It would be convenient but it also would be a side effect of calling the function

wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
```diff
@@ -857,7 +857,7 @@ public void test_pooledConnection_differentUsers() throws SQLException {
 
       assertThrows(
           HikariPool.PoolInitializationException.class, () -> {
-            try (final Connection conn = DriverManager.getConnection(
+            try (final Connection ignored = DriverManager.getConnection(
```
aaron-congo, line 860: just fixing an intellij warning here

wrapper/src/test/java/integration/util/AuroraTestUtility.java
```diff
@@ -935,7 +953,7 @@ public void makeSureInstancesUp(List<TestInstanceInfo> instances, long timeoutSe
               host,
               instanceInfo.getPort(),
               dbName);
-          try (final Connection conn = DriverManager.getConnection(url, props)) {
+          try (final Connection ignored = DriverManager.getConnection(url, props)) {
```
aaron-congo, line 956: just getting rid of an intellij warning here

sergiyvamz (approved):

PR #1258, opened 2025-01-23
test: fix test_pooledConnection_differentUsers

aaron-congo:
### Description

test_pooledConnection_differentUsers was failing on multi-az mysql-latest with MariaDB. When trying to connect with the limited user the test was throwing the error "RSA public key is not available client side". Presumably this is because of some changes in the newest version of rds-mysql. The advice on this error message from stackoverflow was to set allowPublicKeyRetrieval to true.

### DIFF
--- a/wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
+++ b/wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
@@ -826,6 +826,10 @@ public class ReadWriteSplittingTests {
     String limitedUserNewDb = "limited_user_db";
     limitedUserProps.setProperty(PropertyDefinition.USER.name, limitedUserName);
     limitedUserProps.setProperty(PropertyDefinition.PASSWORD.name, limitedUserPassword);
+    // This property is required when using limited_user with the mariadb driver against multi-az mysql version 8.4,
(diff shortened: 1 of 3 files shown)


wrapper/src/test/java/integration/container/tests/ReadWriteSplittingTests.java
```diff
@@ -826,6 +826,10 @@ public void test_pooledConnection_differentUsers() throws SQLException {
     String limitedUserNewDb = "limited_user_db";
     limitedUserProps.setProperty(PropertyDefinition.USER.name, limitedUserName);
     limitedUserProps.setProperty(PropertyDefinition.PASSWORD.name, limitedUserPassword);
+    // This property is required when using limited_user with the mariadb driver against multi-az mysql version 8.4,
+    // or you will get the error "RSA public key is not available client side". The mariadb driver may not fully support
+    // all aspects of mysql 8.4's SSL mechanisms, which is why this property is only required for newer mysql versions.
+    limitedUserProps.setProperty("allowPublicKeyRetrieval", "true");
```
sergiyvamz, line 832: Does it make sense to more this property to ConnectionStringHelper.getDefaultProperties() and add it in case of MariaDb driver?

sergiyvamz (approved):

PR #1265, opened 2025-01-29
fix issue with custom domains during failover

sergiyvamz:
### Summary

Fix issue with custom domains during failover
- In case of connections are opened and with a custom domain they are not invalidated during failover
- Fix various issues with tracked connections and custom domains
- Improved verification of tracked connection; Already closed tracked connections are removed.
- Added new configuration parameters `verifyWriterForCustomHostPattern` and `verifyReaderForCustomHostPattern`

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@aaronchung-bitquill 


wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -110,12 +143,8 @@ public Connection connect(
...
     final RdsUrlType type = this.rdsUtils.identifyRdsType(hostSpec.getHost());
 
-    if (!type.isRdsCluster()) {
-      // It's not a cluster endpoint. Continue with a normal workflow.
-      return connectFunc.call();
-    }
-
-    if (type == RdsUrlType.RDS_WRITER_CLUSTER) {
+    if (type == RdsUrlType.RDS_WRITER_CLUSTER
+        || isInitialConnection && this.verifyOpenedConnectionType == VerifyOpenedConnectionType.WRITER) {
```
aaron-congo, line 147: I think we need an extra set of brackets, otherwise we will still verify when using a writer cluster and it is not an initial connection
```suggestion
        || (isInitialConnection && this.verifyOpenedConnectionType == VerifyOpenedConnectionType.WRITER)) {
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -124,7 +153,8 @@ public Connection connect(
       return writerCandidateConn;
     }
 
-    if (type == RdsUrlType.RDS_READER_CLUSTER) {
+    if (type == RdsUrlType.RDS_READER_CLUSTER
+        || isInitialConnection && this.verifyOpenedConnectionType == VerifyOpenedConnectionType.READER) {
```
aaron-congo, line 157:
```suggestion
        || (isInitialConnection && this.verifyOpenedConnectionType == VerifyOpenedConnectionType.READER)) {
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -67,16 +69,47 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
           "1000",
           "Time between each retry of opening a connection.");
 
+  public static final AwsWrapperProperty VERIFY_OPENED_CONNECTION_TYPE =
+      new AwsWrapperProperty(
+          "verifyOpenedConnectionType",
+          null,
+          "Force to verify an opened connection to be either a writer or a reader.");
```
aaron-congo, line 76:
```suggestion
          "Specifies whether the role for opened connection should be verified. Set to 'writer' to verify that the connection opened is to a writer, and 'reader' to verify that the connection opened is to a reader.");
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -67,16 +69,47 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
           "1000",
           "Time between each retry of opening a connection.");
 
+  public static final AwsWrapperProperty VERIFY_OPENED_CONNECTION_TYPE =
+      new AwsWrapperProperty(
+          "verifyOpenedConnectionType",
```
aaron-congo, line 74: Can you please add this property to UsingTheAuroraInitialConnectionStrategyPlugin.md?

wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
```diff
@@ -50,16 +52,17 @@ public class OpenedConnectionTracker {
...
             return invalidateThread;
           });
-  private static final ExecutorService abortConnectionExecutorService =
-      Executors.newCachedThreadPool(
-          r -> {
-            final Thread abortThread = new Thread(r);
-            abortThread.setDaemon(true);
-            return abortThread;
-          });
+  private static final Executor abortConnectionExecutor = new SynchronousExecutor();
```
aaron-congo, line 55: Just double checking, are tasks submitted to the executor daemon threads?

wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
```diff
@@ -50,16 +52,17 @@ public class OpenedConnectionTracker {
...
 
   private static final Logger LOGGER = Logger.getLogger(OpenedConnectionTracker.class.getName());
   private static final RdsUtils rdsUtils = new RdsUtils();
+
+  private static final Set<String> safeCheckIfClosed = new HashSet<>(Arrays.asList(
```
aaron-congo, line 60:
```suggestion
  private static final Set<String> safeToCheckClosedClasses = new HashSet<>(Arrays.asList(
```

wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
```diff
@@ -100,28 +107,27 @@ public void invalidateAllConnections(final HostSpec hostSpec) {
...
 
     try {
-      final Optional<String> instanceEndpoint = Arrays.stream(node)
-          .filter(x -> rdsUtils.isRdsInstance(rdsUtils.removePort(x)))
-          .findFirst();
-      if (!instanceEndpoint.isPresent()) {
-        return;
+      for (String key : keys) {
+        try {
+          final Queue<WeakReference<Connection>> connectionQueue = openedConnections.get(key);
```
aaron-congo, line 118: I believe connectionQueue could be null, which we handle in logConnectionQueue but not invalidateConnections

wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
```diff
@@ -134,8 +140,11 @@ public void invalidateCurrentConnection(final HostSpec hostSpec, final Connectio
     }
 
     final Queue<WeakReference<Connection>> connectionQueue = openedConnections.get(host);
-    logConnectionQueue(host, connectionQueue);
-    connectionQueue.removeIf(connectionWeakReference -> Objects.equals(connectionWeakReference.get(), connection));
+    if (connectionQueue != null) {
+      logConnectionQueue(host, connectionQueue);
+      connectionQueue.removeIf(connectionWeakReference -> connectionWeakReference != null
+          && Objects.equals(connectionWeakReference.get(), connection));
```
aaron-congo, line 146: Should we also remove if connectionWeakReference is null? 
```suggestion
      connectionQueue.removeIf(connectionWeakReference -> connectionWeakReference == null || (connectionWeakReference != null
          && Objects.equals(connectionWeakReference.get(), connection)));
```

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -213,6 +218,19 @@ public <T, E extends Exception> T execute(
...
       final Object[] jdbcMethodArgs)
       throws E {
+
+    try {
+      if (this.enableFailoverSetting
+          && !METHOD_ABORT.equals(methodName)
+          && !METHOD_CLOSE.equals(methodName)
+          && this.pluginService.getCurrentConnection() != null
+          && this.pluginService.getCurrentConnection().isClosed()) {
+        this.pickNewConnection();
```
aaron-congo, line 228: I'm wondering if the condition for picking a new connection here should be stricter. For example, if the user is calling `isClosed` themselves, I don't think it makes sense to fail over

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -106,23 +115,42 @@ public <T, E extends Exception> T execute(final Class<T> resultClass, final Clas
     this.rememberWriter();
 
     try {
+      if (!methodName.equals(METHOD_CLOSE) && methodName.equals(METHOD_ABORT)) {
+        long localRefreshHostListTillTimeNano = refreshHostListTillTimeNano.get();
+        boolean needRefreshHostLists = localRefreshHostListTillTimeNano != 0
+            && localRefreshHostListTillTimeNano < System.nanoTime();
+        if (!needRefreshHostLists && localRefreshHostListTillTimeNano != 0) {
+          refreshHostListTillTimeNano.compareAndSet(localRefreshHostListTillTimeNano, 0);
+        }
```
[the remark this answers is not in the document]
sergiyvamz: That's correct. localRefreshHostListTillTimeNano is set from hostListRefreshThresholdTimeNano and in most case the value will be the same. However hostListRefreshThresholdTimeNano is a static variable and another thread may change. compareAndSet() just ensure that the value of hostListRefreshThresholdTimeNano hasn't changed since we get a local value of it.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -106,23 +115,42 @@ public <T, E extends Exception> T execute(final Class<T> resultClass, final Clas
...
-      } else if (this.needUpdateCurrentWriter) {
-        this.checkWriterChanged();
+        tracker.removeConnectionTracking(currentHostSpec, this.pluginService.getCurrentConnection());
       }
       return result;
 
     } catch (final Exception e) {
       if (e instanceof FailoverSQLException) {
-        this.checkWriterChanged();
+        refreshHostListTillTimeNano.set(System.nanoTime() + TOPOLOGY_CHANGES_EXPECTED_TIME_MS);
```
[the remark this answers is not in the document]
sergiyvamz: According to compiler, there's not method that can raise FailoverSQLException. Since the method is generic the compiler couldn't verify that so a safe solution is to catch Exception and manually verify exception type we need.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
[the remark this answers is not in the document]
sergiyvamz: I reworked this piece of code to make it more readable.

wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
sergiyvamz: Aborting connection is processed as a task in invalidateConnectionsExecutorService (that runs in a daemon thread) but the task uses SynchronousExecutor to abort each connection.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -106,23 +115,42 @@ public <T, E extends Exception> T execute(final Class<T> resultClass, final Clas
     this.rememberWriter();
 
     try {
+      if (!methodName.equals(METHOD_CLOSE) && methodName.equals(METHOD_ABORT)) {
+        long localRefreshHostListTillTimeNano = refreshHostListTillTimeNano.get();
```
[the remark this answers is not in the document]
aaron-congo: Marking unresolved because it hasn't been changed, let me know if you don't think it is necessary. I think its clearer because it will then match the non-local variable name.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -43,17 +45,24 @@ public class AuroraConnectionTrackerPlugin extends AbstractConnectionPlugin impl
...
         {
           addAll(SubscribedMethodHelper.NETWORK_BOUND_METHODS);
+          add(METHOD_CLOSE);
+          add(METHOD_ABORT);
           add("connect");
           add("notifyNodeListChanged");
         }
       });
 
+  private static final AtomicLong hostListRefreshThresholdTimeNano = new AtomicLong(0);
```
aaron-congo, line 64: nit: could remove the "time" portion of the name to make it shorter/clearer
```suggestion
  private static final AtomicLong hostListRefreshThresholdNano = new AtomicLong(0);
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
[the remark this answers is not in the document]
aaron-congo: I realize that hostListRefreshThresholdTimeNano may change since it is static but I don't understand why we set it to 0 in the compareAndSet call if it hasn't changed in the last few lines of code. Why does that mean we should clear the value?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -213,6 +218,19 @@ public <T, E extends Exception> T execute(
...
       final Object[] jdbcMethodArgs)
       throws E {
+
+    try {
+      if (this.enableFailoverSetting
+          && !this.canDirectExecute(methodName)
+          && !this.closedExplicitly.get()
+          && this.pluginService.getCurrentConnection() != null
+          && this.pluginService.getCurrentConnection().isClosed()) {
+        this.pickNewConnection();
```
aaron-congo, line 228: Should we log here seeing as this is a bit of a special scenario? Also, should we add a configuration parameter to enable/disable this? Not sure if some customers might prefer to error out if they end up executing against a closed connection.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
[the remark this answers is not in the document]
sergiyvamz: We reset `hostListRefreshThresholdTimeNano` to zero and stop further refreshes of topology when we reached a time specified in this variable. I added comments to the code. I hope it explains what and why.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
sergiyvamz: > I realize that hostListRefreshThresholdTimeNano may change since it is static but I don't understand why we set it to 0 in the compareAndSet call if it hasn't changed in the last few lines of code. Why does that mean we should clear the value?

I added comments to the code. I hope it adds clarity. The code sets a time value to hostListRefreshThresholdTimeNano when it's necessary to keep refreshing host list. When the specified time is reached, we reset hostListRefreshThresholdTimeNano to zero and deactivates further host list refreshes.

sergiyvamz: > Should we log here seeing as this is a bit of a special scenario? Also, should we add a configuration parameter to enable/disable this? Not sure if some customers might prefer to error out if they end up executing against a closed connection.

Failover code has necessary logging so I believe it provides enough details for troubleshooting. I don't think that user has a choice here. Without failover in this case, user (most probably) gets an error executing JDBC call on closed connection. With this code, user may get a chance to failover and proceed with JDBC call successfully.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
sergiyvamz: I'd prefer to keep "time" in the name. It clearly state the variable holds a nano time, not interval time.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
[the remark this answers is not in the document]
aaron-congo: Oh I misunderstood. I thought we wanted to refresh once after this time was hit. It looks like we are instead refreshing before this time is hit.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
aaron-congo: Right, I was thinking that for customer logs it might be helpful to know that failover was kicked off because of executing against a closed connection which may be different than executing against a network-failed connection. Up to you though.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -106,23 +115,50 @@ public <T, E extends Exception> T execute(final Class<T> resultClass, final Clas
     this.rememberWriter();
 
     try {
+      if (!methodName.equals(METHOD_CLOSE) && !methodName.equals(METHOD_ABORT)) {
+        long localHostListRefreshThresholdTimeNano = hostListRefreshThresholdTimeNano.get();
+        boolean needRefreshHostLists = false;
+        if (localHostListRefreshThresholdTimeNano > 0) {
+          if (localHostListRefreshThresholdTimeNano < System.nanoTime()) {
```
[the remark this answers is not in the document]
sergiyvamz: Good catch!

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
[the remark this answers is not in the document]
sergiyvamz: Good name suggestion.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -43,17 +45,24 @@ public class AuroraConnectionTrackerPlugin extends AbstractConnectionPlugin impl
 
   private static final Logger LOGGER = Logger.getLogger(AuroraConnectionTrackerPlugin.class.getName());
 
+  // Check topology changes 3 min after last failover
+  private static final long TOPOLOGY_CHANGES_EXPECTED_TIME_MS = TimeUnit.MINUTES.toNanos(3);
```
karenc-bq, line 49: Need to rename this to `TOPOLOGY_CHANGES_EXPECTED_TIME_NS` if we are using nanoseconds

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -67,16 +69,47 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
...
+      return nameToValue.get(value.toLowerCase());
+    }
+  }
+
   private final PluginService pluginService;
   private HostListProviderService hostListProviderService;
   private final RdsUtils rdsUtils = new RdsUtils();
 
+  private VerifyOpenedConnectionType verifyOpenedConnectionType = null;
+
```
karenc-bq, line 103:
```suggestion
```

karenc-bq (approved):

PR #1276, opened 2025-02-05
[WIP] Add interfaces for refactoring work

aaron-congo:


wrapper/src/main/java/software/amazon/jdbc/util/StorageService.java
```diff
@@ -0,0 +1,89 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.util;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.plugin.customendpoint.CustomEndpointInfo;
+
+public interface StorageService {
```
sergiyvamz, line 22: boolean exists(String itemCategory, Object key);

wrapper/src/main/java/software/amazon/jdbc/util/events/EventPublisher.java
```diff
@@ -0,0 +1,42 @@
...
+   * @param eventClasses the classes of events that the subscriber wants to unsubscribe from.
+   */
+  void unsubscribe(EventSubscriber subscriber, Class<?>... eventClasses);
+
+  /**
+   * Publish an event. All subscribers to the given event class will be notified of the event.
+   *
+   * @param event the event to publish.
+   */
+  void publish(Object event);
```
sergiyvamz, line 41: It's better to replace Object with a simple interface message interface (even if such interface is empty)

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorState.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.util.monitoring;
+
+public enum MonitorState {
```
sergiyvamz, line 19: CREATED?

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/Monitor.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package software.amazon.jdbc.util.monitoring;
+
+public interface Monitor {
+  void start();
+
+  void stop();
+
+  MonitorStatus getStatus();
```
sergiyvamz, line 24: There's an exception as part of MonitorStatus. Is it an unhandled exception that stops/breaks a monitor?

wrapper/src/main/java/software/amazon/jdbc/util/events/EventPublisher.java
aaron-congo: I was wondering about that. I didn't add it because I think subscribers will have to cast the event whether there's an interface or not, since our events will be pretty generic. But I can see how it might be useful so I can add it.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorState.java
aaron-congo: Does that indicate a monitor that has been instantiated but never run? With the current design the monitor service doesn't provide a method to submit a monitor without running it, we just have the "runIfAbsent" method which would automatically create and run the monitor. So I'm not sure if we'd ever have a monitor in this state.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorStatus.java
```diff
@@ -0,0 +1,54 @@
...
+package software.amazon.jdbc.util.monitoring;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class MonitorStatus {
+  private MonitorState state;
+  private long lastUsedTimeNs;
+  private @Nullable Throwable exception;
+
+  public MonitorStatus(MonitorState state, long lastUsedTimeNs) {
```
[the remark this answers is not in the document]
aaron-congo: I was also wondering about that. If we make it immutable the monitor will have to create a new MonitorStatus object each time it needs to update its state, which could be quite frequent since it will be regularly updating the lastUsedTime field. I was also wondering if we should get rid of the MonitorStatus object and just have its fields directly in the Monitor class. If we did that we would have to add getters to the Monitor interface (getMonitorState, getLastUsedTime, getException).

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/Monitor.java
aaron-congo: That was my intention yes, it would be a reference to an exception that caused a monitor to unexpectedly stop.

wrapper/src/main/java/software/amazon/jdbc/util/events/EventPublisher.java
```diff
@@ -0,0 +1,42 @@
...
+package software.amazon.jdbc.util.events;
+
+public interface EventPublisher {
+  /**
+   * Register the given subscriber for the given event classes.
+   *
+   * @param subscriber   the subscriber to be notified when the given event classes occur.
+   * @param eventClasses the classes of event that the subscriber should be notified of.
+   */
+  void subscribe(EventSubscriber subscriber, Class<?>... eventClasses);
```
sergiyvamz, line 26: Class<? extends Event>

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/Monitor.java
sergiyvamz: That means it's a handled exception (since it's been caught and stored). That contradicts the original goal. It's possible that monitor isn't able to catch it so monitor service needs to handle them.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorState.java
sergiyvamz: Theoretically it's possible to create an instance, add it to the list of monitors and let other threads to observe it. In this case the monitor is created but isn't yet started. Otherwise your code has to guarantee that such monitors won't be available for other observers.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorStatus.java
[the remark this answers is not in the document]
sergiyvamz: I'd start from immutable version of MonitorStatus. If it becomes a bottleneck then we can refine our design.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorState.java
aaron-congo: Currently the API doesn't have a method that creates a monitor but does not run it. The only way to submit a monitor is to call the runIfAbsent method which will automatically start the monitor when it is created. If we discover we need a method that submits a monitor but does not start then yes we should add this state.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/Monitor.java
aaron-congo: Oh, I was thinking the monitor would use a try-catch for its entire run loop that would catch the generic Exception type (eg it would catch any unhandled exceptions) and then it would set its exception status. So the monitor itself would be responsible for indicating it has hit an unexpected error, and the monitor service's watcher thread would check the monitor's status to see if it has hit an error. Do we want to pursue a different approach?

wrapper/src/main/java/software/amazon/jdbc/util/events/EventPublisher.java
aaron-congo: When I make this change I get the warning "Possible heap pollution from parameterized vararg type", so maybe I'll make this a SetClass<? extends Event> instead of a vararg

PR #1286, opened 2025-02-14
[WIP] Implement refactoring changes

aaron-congo:
Note: will be merged into the refactoring feature branch, not the main branch


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -262,19 +268,19 @@ public FetchTopologyResult getTopology(final Connection conn, final boolean forc
       final List<HostSpec> hosts = queryForTopology(conn);
 
       if (!Utils.isNullOrEmpty(hosts)) {
-        topologyCache.put(this.clusterId, hosts, this.refreshRateNano);
+        storageService.set(ItemCategory.TOPOLOGY, this.clusterId, new Topology(hosts));
```
aaron-congo, line 271: The get() method requires that you pass a Class specifying what the expected value is. This is needed inside the method to check if it matches what is in the cache and cast it for the caller. The limitation with this is that the caller cannot specify a generic class. For example, there is no way for the caller to specify specifically that they want a List<HostSpec>. They can specify that they want a List.class but that still does not let us properly check if the List is a List<HostSpec>. The solutions to this are:
1. Use a non-generic class to wrap the generic object, which is what I've done here
2. Remove the expected class as a parameter to get(), and return an Object without casting or checking an expected type. In this case the caller would have to check if the returned Object was the correct type and cast it, which is annoying
3. Google Guava has a [TypeToken](https://guava.dev/releases/21.0/api/docs/com/google/common/reflect/TypeToken.html) dependency which may provide a way to specify and preserve the generic type info, but this requires adding it as a dependency which may not be acceptable

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -492,15 +508,15 @@ protected String getHostEndpoint(final String nodeName) {
    * @return list of hosts that represents topology. If there's no topology in the cache or the
    *     cached topology is outdated, it returns null.
    */
-  public @Nullable List<HostSpec> getCachedTopology() {
-    return topologyCache.get(this.clusterId);
+  public @Nullable List<HostSpec> getStoredTopology() {
+    Topology topology = storageService.get(ItemCategory.TOPOLOGY, this.clusterId, Topology.class);
```
aaron-congo, line 512: The get() method requires that you pass a Class specifying what the expected value is. This is needed inside the method to check if it matches what is in the cache and cast it for the caller. The limitation with this is that the caller cannot specify a generic class. For example, there is no way for the caller to specify specifically that they want a List<HostSpec>. They can specify that they want a List.class but that still does not let us properly check if the List is a List<HostSpec>. The solutions to this are:
1. Use a non-generic class to wrap the generic object, which is what I've done here
2. Remove the expected class as a parameter to get(), and return an Object without casting or checking an expected type. In this case the caller would have to check if the returned Object was the correct type and cast it, which is annoying
3. Google Guava has a [TypeToken](https://guava.dev/releases/21.0/api/docs/com/google/common/reflect/TypeToken.html) dependency which may provide a way to specify and preserve the generic type info, but this requires adding it as a dependency which may not be acceptable

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -241,14 +247,14 @@ public FetchTopologyResult getTopology(final Connection conn, final boolean forc
       this.clusterIdChanged(oldClusterId);
     }
 
-    final List<HostSpec> cachedHosts = topologyCache.get(this.clusterId);
+    final List<HostSpec> storedHosts = getStoredTopology();
```
sergiyvamz, line 250: this.getStoredTopology() - it's increase readability

wrapper/src/main/java/software/amazon/jdbc/util/storage/StorageServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+      String itemCategory,
+      Class<V> itemClass,
+      boolean isRenewableExpiration,
+      long timeToLiveNanos,
+      long cleanupIntervalNanos,
+      @Nullable ShouldDisposeFunc<V> shouldDisposeFunc,
+      @Nullable ItemDisposalFunc<V> itemDisposalFunc) {
+    caches.computeIfAbsent(
+        itemCategory,
+        category -> new ExpirationCache<>(
```
sergiyvamz, line 46: So the design is to have a separate cache (that sits in another cache) for each item category?

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -99,10 +101,17 @@ public ConnectionWrapper(
             effectiveConnectionProvider,
             this,
             telemetryFactory);
+    final StorageService storageService = new StorageServiceImpl();
```
sergiyvamz, line 104: Do we to initialize a new copy of StorageService? As I see it's provided to the constructor.

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -113,6 +120,15 @@ public static void register() throws SQLException {
     final Driver driver = new Driver();
     DriverManager.registerDriver(driver);
     registeredDriver = driver;
+
+    storageService.registerItemCategoryIfAbsent(
```
sergiyvamz, line 124: Should topology be registered here or by HostList provider?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
sergiyvamz, line 512: ItemCategory.TOPOLOGY and Topology.class play quite similar role. Can we simplify the design?

wrapper/src/main/java/software/amazon/jdbc/util/storage/StorageServiceImpl.java
aaron-congo: Each item category has its own cache so that each item category can have its own settings, since different item categories will have different needs (expiration time, cleanup time, shouldDisposeFunc, itemDisposalFunc). The caches are stored in a map that serves as a cache directory so we can quickly find the given cache for an item category.

wrapper/src/main/java/software/amazon/jdbc/Driver.java
aaron-congo: I would prefer to register all default item categories here so that we know they have all been registered and don't need to worry about if they've been registered later. If we do not register here then it is less clear whether an item category has been registered or not, so we will need to call the registerIfAbsent method more often, which would be cumbersome since it requires 7 parameters. Alternatively, we could lazily register default item categories automatically so that no one has to worry whether the default categories have been registered or not.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
aaron-congo: We could remove the item category parameter and keep the item class parameter. This would simplify the API but would also make the storage service less clear/flexible. The item category parameter allows us to identify what a value represents, if we remove it this will be less clear. For example, if you are storing an Integer in the StorageService it is not clear what that Integer represents if there is no item category - it could represent the CPU utilization of an instance, a node response time, or anything else. We also would not be able to store multiple Integers at the same key, for example a node response time and a CPU utilization could not both be keyed to an instance URL. This may not be the best example because it would probably be best to map the instance URL to an InstanceInfo object, but it still illustrates my point

sergiyvamz: Having all types of data enlisted in ItemCategory enum might be a serious constraint for user custom plugins. Enums are not extendable solution. Can we use data class type only?

wrapper/src/main/java/software/amazon/jdbc/util/storage/StorageServiceImpl.java
sergiyvamz: Map of caches might not be a savvy solution. Each cache has its own clean-up thread so I can assume that it would be as many clean-up threads as data types. Most probably each cache has just one or two entries. For example, topology cache will have a single entry. If user app works with two different clusters (that is a rare case) then topology cache will have two entries... Can we pack all entries into a single cache (i.e. instead of map of caches use a single cache)?

wrapper/src/main/java/software/amazon/jdbc/Driver.java
sergiyvamz: Hmm not sure I see these changes. Can you double check?

aaron-congo: Forgot to push, updated now

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
aaron-congo: ItemCategory does not hold enum values, it holds String constants for the reason you mentioned: so that user custom plugins can still use their own categories via their own strings. We can consider using data class type only but it has some downsides which I have mentioned above

wrapper/src/main/java/software/amazon/jdbc/util/storage/StorageServiceImpl.java
aaron-congo: Good point, but it sounds like the main issue is the cleanup mechanism and number of threads instead of the internal structure. I'd like to keep the data grouped together because it allows us to cleanly store the data together with the settings for how to handle that data in one spot (ExpiringCache). I've updated the code now so that there is only one cleanup thread for all items, which is created and run in StorageServiceImpl.

wrapper/src/main/java/software/amazon/jdbc/util/storage/StorageServiceImpl.java
```diff
@@ -0,0 +1,232 @@
...
+  protected static final ScheduledExecutorService cleanupExecutor = Executors.newSingleThreadScheduledExecutor((r -> {
+    final Thread thread = new Thread(r);
+    thread.setDaemon(true);
+    return thread;
+  }));
+
+  static {
+    Map<String, Supplier<ExpirationCache<Object, ?>>> suppliers = new HashMap<>();
+    suppliers.put(ItemCategory.TOPOLOGY, () -> new ExpirationCacheBuilder<>(Topology.class).build());
+    suppliers.put(ItemCategory.CUSTOM_ENDPOINT, () -> new ExpirationCacheBuilder<>(CustomEndpointInfo.class).build());
```
sergiyvamz, line 53: This type of data might be not used by user, in a particular driver configuration. I think it should be initialized by corresponding plugin

aaron-congo: This line does not actually instantiate any caches, it just defines a lambda that can be used to create the custom endpoint info cache if it is needed. The lambda is only ever called the first time someone tries to store something in the custom endpoint cache, at which point it will be created. The benefit of this is that our code doesn't need to worry about whether the cache exists or not yet. Instead, it will be automatically created, but only if it is actually needed.

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -99,10 +103,19 @@ public ConnectionWrapper(
             effectiveConnectionProvider,
             this,
             telemetryFactory);
+    ServiceContainer serviceContainer = new ServiceContainerImpl(storageService, pluginManager, telemetryFactory);
```
aaron-congo, line 106: I don't love how we partially initialize the ServiceContainer here and then complete the initialization with the setters below. I wonder if we should consider a proper service locator to avoid this.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorService.java
```diff
@@ -24,52 +24,64 @@
...
+
+  /**
+   * Process a monitor error. The monitor service will respond to the error based on the monitor error responses defined
+   * when the monitor type was registered.
+   *
+   * @param monitor   the monitor that encountered the unexpected exception.
+   * @param key       the key for the monitor.
+   * @param exception the unexpected exception that occurred.
+   */
+  void processMonitorError(Monitor monitor, Object key, Exception exception);
```
aaron-congo, line 68: I added this method on the suggestion that monitors be able to notify the monitor service when they encounter an unexpected exception. The monitor service can then respond to this based on pre-configured settings for that monitor type, for example it may try and re-create the monitor. I think this is a good idea, but it requires the monitor to pass in its own key here, so that the monitor service can update its cache with the re-created monitor. It seems bad that the monitor needs to keep track of its own key, but I'm not sure how to fix this.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorServiceImpl.java
```diff
@@ -0,0 +1,232 @@
...
+      Class<T> monitorClass,
+      long timeToLiveNanos,
+      long inactiveTimeoutNanos,
+      Set<MonitorErrorResponse> errorResponses,
+      @Nullable ShouldDisposeFunc<T> shouldDisposeFunc) {
+    monitorCaches.computeIfAbsent(
+        monitorClass,
+        mc -> {
+          monitorSettingsByType.putIfAbsent(monitorClass, new MonitorSettings(inactiveTimeoutNanos, errorResponses));
+          ShouldDisposeFunc<MonitorItem> wrappedShouldDisposeFunc = shouldDisposeFunc == null ? null
```
aaron-congo, line 144: I created the MonitorItem inner class so that monitors can be stored together with the suppliers used to create them. This allows us to re-create the monitor if it encounters an error or gets stuck. However, it makes things messy in this function. The caller passes in a shouldDisposeFunc for the monitor itself (eg ClusterTopologyMonitorImpl), but since we are actually storing MonitorItem's in the cache, we need to pass ShouldDisposeFunc\<MonitorItem>. For now I resolved this by wrapping the passed in lambda, but this seems messy and I am getting a warning when I try to cast monitorItem.getMonitor() to T.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorService.java
```diff
@@ -24,52 +24,64 @@
...
-   * @param errorResponses    a Set defining actions to take if the monitor is in an error state.
-   * @param expirationTimeNs  how long a monitor should be stored before expiring, in nanoseconds. If the monitor is
-   *                          expired and shouldDisposeFunc returns `true`, the monitor will be stopped.
-   * @param shouldDisposeFunc a function defining whether an item should be stopped if expired. If `null` is passed, the
-   *                          monitor will always be stopped if the monitor is expired.
+   * @param monitorClass         the class of the monitor, eg `CustomEndpointMonitorImpl.class`.
+   * @param timeToLiveNanos      how long a monitor should be stored before being considered expired, in nanoseconds. If
+   *                             the monitor is expired and shouldDisposeFunc returns `true`, the monitor will be
+   *                             stopped.
+   * @param inactiveTimeoutNanos a duration in nanoseconds defining the maximum amount of time that a monitor should
```
aaron-congo, line 35: I like the idea of this argument which allows us to detect when monitors are frozen or stuck. However, I don't like that the difference between timeToLiveNanos and inactiveTimeoutNanos is hard to understand, even if we document it. To clarify, timeToLiveNanos represents a monitor's expiration, which is renewed if runIfAbsent is called but the monitor already exists. If the monitor has not been referenced via runIfAbsent in a long time it will be considered expired. inactiveTimeoutNanos represents a heartbeat threshold. If we haven't had a heartbeat within this threshold the monitor will be considered stuck.

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -67,6 +69,8 @@ public class AwsWrapperDataSource implements DataSource, Referenceable, Serializ
   private static final String SERVER_NAME = "serverName";
   private static final String SERVER_PORT = "serverPort";
 
+  private static final StorageService storageService = new StorageServiceImpl();
```
sergiyvamz, line 72: We have another instance of the service in Driver class. We should use a single one.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/AbstractMonitor.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.util.monitoring;
+
+import java.util.logging.Logger;
+import software.amazon.jdbc.util.Messages;
+
+public abstract class AbstractMonitor implements Monitor, Runnable {
```
sergiyvamz, line 22: Does it make sense to add heartbeat functionality to this class? It'd be nice just to inherit it.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/Monitor.java
```diff
@@ -17,9 +17,11 @@
 package software.amazon.jdbc.util.monitoring;
 
 public interface Monitor {
-  void start();
+  void execute();
```
sergiyvamz, line 20: I'd personally prefer `start()` and `stop()`. It's cleaner for me and it shows that monitor runs for a long time. `execute()` sounds like a quick run and get result.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorService.java
sergiyvamz: You're passing monitor reference to monitor service. I believe that's sufficient. you don't need a key parameter. Monitor service can find a monitor and all related info by monitor reference.

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
sergiyvamz: There's nothing wrong about partially initialize the object. By design this object is supposed to hold service references. Some of them could be missed, i.e. not yet initialized. I think we 're fine. We could improve it later.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorService.java
aaron-congo: Yeah, the monitor service will have to iterate over its entries to find it but I guess that doesn't matter much since there shouldn't be a large number of monitors of any given type

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
aaron-congo: The Driver and AwsWrapperDataSource are independent classes so there is not a way to pass the storage service from one to the other. The Driver class needs the storage service reference so that it can clear it when Driver.clearCaches is called. I think the only way we could make sure its a single instance is if we make it a singleton, do we want to do that?

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/AbstractMonitor.java
aaron-congo: That would be nice, I'm not sure how to add that though. Each monitor type will have its own custom run loop that it needs to define, and the last updated timestamp needs to be updated somewhere in that custom loop. Not sure how to extract the heartbeat piece to the abstract monitor while leaving the custom loop implementation to the underlying monitor class.

aaron-congo:
Closed in favor of #1371

aaron-congo:
We will merge in #1371 instead, closing

Issue #1321, opened 2025-03-11
ClusterTopologyMonitorImpl leaks threads


vvitiesc:
### Describe the bug

`"Thread-75817-nm" #271317 daemon prio=5 os_prio=0 cpu=2817.92ms elapsed=1082.59s tid=0x00007fbb144d4000 nid=0x1ff7 sleeping [0x00007fb7e2463000]
   java.lang.Thread.State: TIMED_WAITING (sleeping)
        at java.lang.Thread.sleep(java.base@11.0.18/Native Method)
        at java.lang.Thread.sleep(java.base@11.0.18/Thread.java:334)
        at java.util.concurrent.TimeUnit.sleep(java.base@11.0.18/TimeUnit.java:446)
        at software.amazon.jdbc.hostlistprovider.monitoring.ClusterTopologyMonitorImpl$NodeMonitoringThread.run(ClusterTopologyMonitorImpl.java:842)

"Thread-75819-nm" #271429 daemon prio=5 os_prio=0 cpu=2663.22ms elapsed=998.01s tid=0x00007fbad4176800 nid=0x2079 sleeping [0x00007fb7c89ca000]
   java.lang.Thread.State: TIMED_WAITING (sleeping)
        at java.lang.Thread.sleep(java.base@11.0.18/Native Method)
        at java.lang.Thread.sleep(java.base@11.0.18/Thread.java:334)
        at java.util.concurrent.TimeUnit.sleep(java.base@11.0.18/TimeUnit.java:446)
        at software.amazon.jdbc.hostlistprovider.monitoring.ClusterTopologyMonitorImpl$NodeMonitoringThread.run(ClusterTopologyMonitorImpl.java:842)

"Thread-75820-nm" #271430 daemon prio=5 os_prio=0 cpu=2662.86ms elapsed=997.99s tid=0x00007fbad4177000 nid=0x207a sleeping [0x00007fb7c4182000]
   java.lang.Thread.State: TIMED_WAITING (sleeping)
        at java.lang.Thread.sleep(java.base@11.0.18/Native Method)
        at java.lang.Thread.sleep(java.base@11.0.18/Thread.java:334)
        at java.util.concurrent.TimeUnit.sleep(java.base@11.0.18/TimeUnit.java:446)
        at software.amazon.jdbc.hostlistprovider.monitoring.ClusterTopologyMonitorImpl$NodeMonitoringThread.run(ClusterTopologyMonitorImpl.java:842)

[...]

PR #1325, opened 2025-03-17
prevent threads leaking

sergiyvamz:
### Summary

Use ExecutorService to manage node monitoring threads
https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1321

### Additional Reviewers

@karenc-bq 


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -438,13 +441,58 @@ public void run() {
...
+        if (!this.nodeExecutorService.isShutdown()) {
+          this.nodeExecutorService.shutdown();
+        }
+
+        try {
+          if (!this.nodeExecutorService.awaitTermination(30, TimeUnit.SECONDS)) {
+            this.nodeExecutorService.shutdownNow();
+          }
+        } catch (InterruptedException e) {
+          // do nothing
```
karenc-bq, line 463: Would it make sense to call `this.nodeExecutorService.shutdownNow();` here if the awaitTemrmination request is interrupted

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
sergiyvamz: There's another final block in run() method that can catch InterruptedException and handle it.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -427,6 +429,7 @@ public void run() {
 
     } finally {
       this.stop.set(true);
+      this.shutdownNodeExecutorService();
 
       final Connection conn = this.monitoringConnection.get();
```
ivankhoosty, line 434: @sergiyvamz would it be better to use getAndSet ? in a few places you are splitting get and set on atomic in this class...

Issue #1337, opened 2025-03-28
Property names from software.amazon.jdbc.PropertyDefinition in camelCase are not useful for configuration with java spring


eugen-eugen:
### **Use Case**
In my use case I use the aws jdbc wrapper with java spring boot and must configure it with os environments variables.
For example for **wrapperPlugins** I want to set:

SPRING_DATASOURCE_HIKARI_DATA-SOURCE-PROPERTIES_WRAPPERPLUGINS=iam

The java spring converts it into jdbc driver property 
wrapper**p**lugins=iam

### **Problem**
The property is then not in camel case, but in lowcase. The aws wrapper expects this property name in camel case, however, and then doesn't recognize the property.

### **Feature Request**
The wrapper must not be case sensitive while it recognizes property names.
  

### Use Case

In my use case I use the aws jdbc wrapper with java spring boot and must configure it with os environments variables.
For example for wrapperPlugins I want to set:

SPRING_DATASOURCE_HIKARI_DATA-SOURCE-PROPERTIES_WRAPPERPLUGINS=iam

The java spring converts it into jdbc driver property
wrapperplugins=iam

### Proposed Solution

The wrapper must not be case sensitive while it recognizes property names. In order to not have beaking changes this feature must be configurable with a further parameter, which name, however must be low case, for example: 
`property-case-sensitive (true (default) / false) `

### Acknowledgements

- [x] I may be able to implement this feature request
- [x] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

2.31

### JDK version used

17

### Operating System and version

macOS

sergiyvamz:
Hi @eugen-eugen 

Thanks for reaching out and raising this issue. We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

sergiyvamz:
Hi @eugen-eugen 

Could you please try the latest snapshot build and set the `wrapperCaseSensitive=false` (or `wrappercasesensitive=false`) connection parameter to see if this resolves your issue?

eugen-eugen:
Hi @sergiyvamz ,
thanks a lot, with the current wrapper, built on main branch i can use the os environment variables for configuration, if i set wrapperCaseSensitive=false. It works as anticipated.
BR

sergiyvamz:
Great! I'll be closing the issue. Feel free to reopen it or create a new one if needed.

PR #1361, opened 2025-04-17
case sensitive parameter

sergiyvamz:
### Summary

Allow driver to recognize lower case parameters.
https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1337

### Additional Reviewers

@karenc-bq 


wrapper/src/main/java/software/amazon/jdbc/util/ConnectionUrlParser.java
```diff
@@ -188,6 +192,42 @@ public static void parsePropertiesFromUrl(final String url, final Properties pro
...
+    String caseSensitiveSettingStr = props.getProperty(PropertyDefinition.CASE_SENSITIVE.name);
+    if (caseSensitiveSettingStr == null) {
+      // try low case
+      caseSensitiveSettingStr = props.getProperty(PropertyDefinition.CASE_SENSITIVE.name.toLowerCase());
+    }
+    if (caseSensitiveSettingStr == null) {
+      caseSensitiveSettingStr = PropertyDefinition.CASE_SENSITIVE.defaultValue;
+    }
+
+    final boolean caseSensitive = Boolean.parseBoolean(caseSensitiveSettingStr);
```
karenc-bq, line 206: Instead of having this here, what do you think of overloading the `getBoolean` method in `AwsWrapperProperty` and move this logic there?
E.g.

```java
  public boolean getBoolean(final Properties properties) {
    return this.getBoolean(properties, false);
  }

  public boolean getBoolean(final Properties properties, final boolean checkLowerCase) {
    Object value = properties.get(name);
    if (checkLowerCase && value == null) {
      value = properties.get(name);
    }
    if (value instanceof Boolean) {
      return (Boolean) value;
    }
    return Boolean.parseBoolean(properties.getProperty(name, defaultValue));
  }
```

karenc-bq: Then we can just simplify this to 
```java
final boolean caseSensitive = PropertyDefinition.CASE_SENSITIVE.getBoolean(props, true);
```

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -179,6 +186,10 @@ public class PropertyDefinition {
     return PROPS_BY_NAME.get(name);
   }
 
+  public static @Nullable AwsWrapperProperty byNameIgnoreCase(final String name) {
+    return PROPS_BY_NAME_LOWERCASE.get(name.toLowerCase());
+  }
+
```
karenc-bq, line 192: Instead of a new function can we overload the original `byName`?
```java
  public static @Nullable AwsWrapperProperty byName(final String name) {
    return byName(name, false);
  }

  public static @Nullable AwsWrapperProperty byName(final String name, final boolean ignoreCase) {
    return ignoreCase ? PROPS_BY_NAME_LOWERCASE.get(name.toLowerCase()) : PROPS_BY_NAME.get(name);
  }
```
If we do this we can simplify the logic in ConnectionUrlParser

wrapper/src/main/java/software/amazon/jdbc/util/ConnectionUrlParser.java
```diff
@@ -188,6 +192,42 @@ public static void parsePropertiesFromUrl(final String url, final Properties pro
...
+          props.setProperty(entry.getKey().toString(), entry.getValue().toString());
+          continue;
+        }
+
+        awsWrapperProperty = PropertyDefinition.byNameIgnoreCase(entry.getKey().toString());
+        if (awsWrapperProperty != null) {
+          // use original (camel-case) property name
+          props.setProperty(awsWrapperProperty.name, entry.getValue().toString());
+          continue;
+        }
```
karenc-bq, line 225: I think we can simplify this to:

```suggestion
        final AwsWrapperProperty awsWrapperProperty = PropertyDefinition.byName(entry.getKey().toString(), true);
        if (awsWrapperProperty != null) {
          props.setProperty(awsWrapperProperty.name, entry.getValue().toString());
          continue;
        }
```

karenc-bq (approved):

PR #1371, opened 2025-04-24
Implement refactoring improvements and adapt topology and custom endpoint caches and monitors

aaron-congo:
### Description

Implements the following refactoring improvements:
1. StorageService - a centralized, easy to access class that stores various types of information that our driver uses, eg custom endpoint info and topology info.
2. MonitorService - a centralized, easy to access class that standardizes the submission and lifetime of the various monitors that our driver creates, eg custom endpoint monitors and topology monitors. Submitted monitors are stored under this service.
3. ConnectionService - replaces the forceConnect functionality, providing the ability to create auxiliary ConnectionWrapper objects that no longer share a PluginService with the connection that is requesting the auxiliary connection.
4. EventPublisher/EventSubscriber - a publish-subscribe pattern that allows the communication of generic events between interested driver components.

So far, the custom endpoint and topology caches and monitors have been adapted to use the StorageService and MonitorService. In the future, more caches and monitors will be adapted as well.


wrapper/src/main/java/software/amazon/jdbc/ConnectionPlugin.java
```diff
@@ -93,7 +94,10 @@ Connection connect(
    * @return a {@link Connection} to the requested host
    * @throws SQLException if there was an error establishing a {@link Connection} to the requested
    *                      host
+   *
+   * @deprecated Use {@link ConnectionService#createAuxiliaryConnection(HostSpec, Properties)} instead.
    */
+  @Deprecated
```
sergiyvamz, line 100: Hmm we have plugins (not monitors) that need this pipeline. InitialConnection is one of them.

wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
```diff
@@ -206,8 +209,14 @@ public HostSpec getInitialConnectionHostSpec() {
   }
 
   @Override
+  public ServiceContainer getServiceContainer() {
```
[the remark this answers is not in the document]
sergiyvamz: I'd create a new implementation of getInstance() with ServiceContainer as a parameter. Current getInstance() can be marked for deprecation. If any of our users have their own plugin implementations we should give them time to adopt.

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -53,6 +53,11 @@ public class PropertyDefinition {
       new AwsWrapperProperty(
           "wrapperPlugins", null, "Comma separated list of connection plugin codes");
 
+  public static final AwsWrapperProperty IS_AUXILIARY_CONNECTION =
```
sergiyvamz, line 56: Should it be a public configuration parameter?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsMultiAzHostListProvider.java
```diff
@@ -31,32 +30,40 @@ public class MonitoringRdsMultiAzHostListProvider extends MonitoringRdsHostListP
...
   protected ClusterTopologyMonitor initMonitor() {
-    return monitors.computeIfAbsent(this.clusterId,
-        (key) -> new MultiAzClusterTopologyMonitorImpl(
-            key, topologyCache, this.initialHostSpec, this.properties, this.pluginService,
-            this.hostListProviderService, this.clusterInstanceTemplate,
-            this.refreshRateNano, this.highRefreshRateNano, TOPOLOGY_CACHE_EXPIRATION_NANO,
+    return monitorService.runIfAbsent(MultiAzClusterTopologyMonitorImpl.class,
+        this.clusterId,
+        () -> new MultiAzClusterTopologyMonitorImpl(
+            this.serviceContainer,
```
sergiyvamz, line 56: Hmm I'm not sure monitor should use serviceContainer that belongs to a particular wrapper connection. Monitor service should provide all necessary services (like ConnectionService) to a monitor instance.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -170,8 +171,13 @@ public Connection connect(
       final boolean isInitialConnection,
       final JdbcCallable<Connection, SQLException> connectFunc)
       throws SQLException {
-
-    ConnectionProvider connProvider = this.connProviderManager.getConnectionProvider(driverProtocol, hostSpec, props);
+    ConnectionProvider connProvider;
+    if (PropertyDefinition.IS_AUXILIARY_CONNECTION.getBoolean(props))  {
```
sergiyvamz, line 175: We already have forceConnect pipeline for that.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitor.java
```diff
@@ -22,8 +22,9 @@
 import java.util.concurrent.TimeoutException;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.util.monitoring.Monitor;
 
-public interface ClusterTopologyMonitor extends AutoCloseable, Runnable {
+public interface ClusterTopologyMonitor extends Monitor, Runnable {
```
sergiyvamz, line 27: Do you need to extend from Runnable?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
```diff
@@ -210,22 +204,20 @@ public Connection connect(
...
   protected CustomEndpointMonitor createMonitorIfAbsent(Properties props) {
-    return monitors.computeIfAbsent(
-        this.customEndpointHostSpec.getHost(),
-        (customEndpoint) -> new CustomEndpointMonitorImpl(
-            this.pluginService,
+    return this.monitorService.runIfAbsent(
+        CustomEndpointMonitorImpl.class,
+        this.customEndpointHostSpec.getUrl(),
+        () -> new CustomEndpointMonitorImpl(
+            this.serviceContainer,
```
sergiyvamz, line 211: Similar comment

wrapper/src/main/java/software/amazon/jdbc/plugin/AuxiliaryPluginFactory.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Properties;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.util.connection.ConnectionService;
+
+/**
+ * A marker interface for plugin factories that create plugins that should be included in auxiliary connections created
+ * with {@link ConnectionService#createAuxiliaryConnection(HostSpec, Properties)}.
+ */
+public interface AuxiliaryPluginFactory {}
```
sergiyvamz, line 27: We need to review this solution. It's a functional duplicate of forceConnect().

wrapper/src/main/java/software/amazon/jdbc/util/events/PeriodicEventPublisher.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.TimeUnit;
+import software.amazon.jdbc.util.StringUtils;
+
+/**
+ * An event publisher that periodically sends out all messages received during the latest interval of time. Messages are
+ * recorded in a set so that messages of equivalent value are not duplicated in the same message batch.
+ */
+public class PeriodicEventPublisher implements EventPublisher {
+  protected static final long DEFAULT_MESSAGE_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(30);
+  protected final Map<Class<? extends Event>, Set<EventSubscriber>> subscribers = new ConcurrentHashMap<>();
```
sergiyvamz, line 33: Should we use weak references to EventSubscriber?

wrapper/src/main/java/software/amazon/jdbc/util/storage/ExternallyManagedCache.java
```diff
@@ -0,0 +1,283 @@
...
+   * @return the value stored at the given key, or null if the key does not exist or the value is expired.
+   */
+  public @Nullable V get(@NonNull K key) {
+    CacheItem cacheItem = this.cache.get(key);
+    if (cacheItem == null) {
+      return null;
+    }
+
+    if (cacheItem.isExpired()) {
+      return null;
```
sergiyvamz, line 84: Should we remove cacheItem from the cache?

wrapper/src/main/java/software/amazon/jdbc/ConnectionPlugin.java
aaron-congo: I checked on main and AuroraInitialConnectionStrategyPlugin does not implement forceConnect. In this PR, all of the plugins that implemented forceConnect have been changed to use ConnectionService#createAuxiliaryConnection, which is similar to forceConnect but slightly different (see my comment below for a more detailed explanation)

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
aaron-congo: I was wondering about that as well, but this parameter is accessed in multiple packages (util.connection and plugin). I can add a sentence in the description to clarify that it should not be used by users, and we can keep it out of our documents, but is there a better solution?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsMultiAzHostListProvider.java
aaron-congo: I was wondering about this as well. We can switch to passing the individual services like you are suggesting but that makes the already-long constructor list 4 parameters longer and I know the intent of the service container was to reduce the amount of constructor parameters around the driver. Both options have pros and cons but I'm fine with either option so let me know what you prefer.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
aaron-congo: This is my understanding from our discussions about ConnectionService: the problem with forceConnect is that it results in shared plugin services between connections, when it would be cleaner if each connection were separate and had their own plugin services. The main difference between forceConnect and createAuxiliaryConnection is that connections created with createAuxiliaryConnection do not share plugin services. Let me know if I am missing something. Also, see [this todo](https://github.com/aws/aws-advanced-jdbc-wrapper/blob/storage-service/wrapper/src/main/java/software/amazon/jdbc/util/connection/ConnectionServiceImpl.java#L65), I'm not sure how to make sure the new connection does not share a connection provider or target driver dialect.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitor.java
aaron-congo: I think I can remove Runnable now, good catch

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointPlugin.java
aaron-congo: See my comment [here](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1371#discussion_r2064397863), we are using the service container to reduce constructor parameters but I was wondering if there were cases when we want to use the longer constructor parameters for more isolation. Sounds like you prefer monitors to have the longer constructors so that they are more isolated?

wrapper/src/main/java/software/amazon/jdbc/util/events/PeriodicEventPublisher.java
aaron-congo: Hmm good question. What is the problem we would be trying to avoid? I know weak references help with garbage collection but am unsure which scenario would cause a garbage collection problem in this case

wrapper/src/main/java/software/amazon/jdbc/ConnectionPlugin.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import software.amazon.jdbc.util.connection.ConnectionService;
```
sergiyvamz, line 26: don't need it?

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginManager.java
```diff
@@ -118,7 +118,7 @@ public class ConnectionPluginManager implements CanReleaseResources, Wrapper {
   public ConnectionPluginManager(
       final @NonNull ConnectionProvider defaultConnProvider,
       final @Nullable ConnectionProvider effectiveConnProvider,
-      final @NonNull ConnectionWrapper connectionWrapper,
+      final @Nullable ConnectionWrapper connectionWrapper,
```
sergiyvamz, line 121: I think it make sense to find all references of connectionWrapper and add null checks there. For debug build we can throw an exception, and for production build we just do nothing. Not a first priority.

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+  }
+
+  @Override
+  @Deprecated
+  public void setAllowedAndBlockedHosts(AllowedAndBlockedHosts allowedAndBlockedHosts) {
+    this.storageService.set(this.initialConnectionHostSpec.getHost(), allowedAndBlockedHosts);
+  }
+
+  @Override
+  public boolean acceptsStrategy(HostRole role, String strategy) throws SQLException {
```
sergiyvamz, line 222: Supporting strategies (acceptStrategy and getHostSpecByStrategy) is interesting. It's not too hard to support them but we kinda not expecting forceConnect pipeline to use them. Can we just temporary let them raise an exception?

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+      }
+    }
+
+    if (!changes.isEmpty()) {
+      this.pluginManager.notifyNodeListChanged(changes);
+    }
+  }
+
+  @Override
+  public boolean isInTransaction() {
```
sergiyvamz, line 478: I don't think that connections in monitoring threads will use transactions. We can simplify implementation here, like do nothing.

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+    return this.getHostListProvider() instanceof StaticHostListProvider;
+  }
+
+  @Override
+  public void setHostListProvider(final HostListProvider hostListProvider) {
+    this.hostListProvider = hostListProvider;
+  }
+
+  @Override
+  public Connection connect(final HostSpec hostSpec, final Properties props) throws SQLException {
```
sergiyvamz, line 605: I believe we need to raise an exception here. We don't want to support this pipeline.

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+    return this.connect(hostSpec, props, null);
+  }
+
+  @Override
+  public Connection connect(
```
sergiyvamz, line 610: The same, raise an exception.

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void releaseResources() {
+    LOGGER.fine(() -> Messages.get("PluginServiceImpl.releaseResources"));
+
+    try {
+      if (this.currentConnection != null && !this.currentConnection.isClosed()) {
```
sergiyvamz, line 651: We can't close current connection here. The whole idea of a "current connection" makes no sense for monitoring service and monitors.

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
```diff
@@ -0,0 +1,795 @@
...
+    return this.pluginManager.getTelemetryFactory();
+  }
+
+  public String getTargetName() {
+    return this.pluginManager.getDefaultConnProvider().getTargetName();
+  }
+
+  @Override
+  public @NonNull SessionStateService getSessionStateService() {
+    return this.sessionStateService;
```
sergiyvamz, line 780: Not sure about supporting session state. We don't have a current connection here. We don't need to take care after connection state.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -265,8 +256,13 @@ protected List<HostSpec> waitTillTopologyGetsUpdated(final long timeoutMs) throw
...
   }
 
+  private List<HostSpec> getStoredHosts() {
+    Topology topology = storageService.get(Topology.class, this.clusterId);
+    return topology == null ? null : topology.getHosts();
+  }
+
   @Override
-  public void close() throws Exception {
+  public void stop() {
```
sergiyvamz, line 265: Should we relocate it to AbstractMonitor?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -503,7 +510,7 @@ protected List<HostSpec> openAnyConnectionAndUpdateTopology() {
 
       // open a new connection
       try {
-        conn = this.pluginService.forceConnect(this.initialHostSpec, this.monitoringProperties);
+        conn = this.connectionService.createAuxiliaryConnection(this.initialHostSpec, this.monitoringProperties);
```
sergiyvamz, line 513: Any chance we rename the method to a simpler one? Like openConnection() or just open()

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsHostListProvider.java
```diff
@@ -89,41 +77,46 @@ public static void clearCache() {
...
   }
 
-  protected ClusterTopologyMonitor initMonitor() {
-    return monitors.computeIfAbsent(this.clusterId,
-        (key) -> new ClusterTopologyMonitorImpl(
-            key, topologyCache, this.initialHostSpec, this.properties, this.pluginService,
-            this.hostListProviderService, this.clusterInstanceTemplate,
-            this.refreshRateNano, this.highRefreshRateNano, TOPOLOGY_CACHE_EXPIRATION_NANO,
+  protected ClusterTopologyMonitor initMonitor() throws SQLException {
+    return monitorService.runIfAbsent(
```
sergiyvamz, line 86: It's getting longer but it makes sense.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsHostListProvider.java
```diff
@@ -89,41 +77,46 @@ public static void clearCache() {
...
             this.writerTopologyQuery,
-            this.nodeIdQuery),
-        MONITOR_EXPIRATION_NANO);
+            this.nodeIdQuery));
   }
 
   @Override
   protected List<HostSpec> queryForTopology(final Connection conn) throws SQLException {
-    ClusterTopologyMonitor monitor = monitors.get(this.clusterId, MONITOR_EXPIRATION_NANO);
+    ClusterTopologyMonitor monitor = monitorService.get(ClusterTopologyMonitorImpl.class, this.clusterId);
```
sergiyvamz, line 115: Should we use ClusterTopologyMonitorImpl.class or ClusterTopologyMonitor.class?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPluginFactory.java
```diff
@@ -25,6 +25,6 @@ public class FailoverConnectionPluginFactory implements ConnectionPluginFactory
 
   @Override
   public ConnectionPlugin getInstance(final PluginService pluginService, final Properties props) {
```
sergiyvamz, line 27: Do we want to deprecate this method and add a new one that accepts service container and properties?

wrapper/src/main/java/software/amazon/jdbc/util/connection/ConnectionServiceImpl.java
```diff
@@ -0,0 +1,87 @@
...
+      Dialect dbDialect,
+      Properties props) throws SQLException {
+    this.targetDriverProtocol = targetDriverProtocol;
+
+    ServiceContainer serviceContainer = new ServiceContainerImpl(storageService, monitorService, telemetryFactory);
+    this.pluginManager = new ConnectionPluginManager(
+        connectionProvider,
+        null,
+        null,
+        // TODO: is it okay to share the telemetry factory?
```
sergiyvamz, line 57: It should be ok to share it.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/AbstractMonitor.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Forms the suffix for the monitor thread name by abbreviating the concrete class name. For example, a
+   * {@link CustomEndpointMonitorImpl} will have a suffix of "cemi".
+   *
+   * @return the suffix for the monitor thread name.
+   */
+  private String getMonitorNameSuffix() {
+    return this.getClass().getSimpleName().replaceAll("[a-z]", "").toLowerCase();
```
sergiyvamz, line 95: Cool :)

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/MonitoringRdsHostListProvider.java
aaron-congo: I think it makes sense to use the concrete class because we might have other concrete implementations of ClusterTopologyMonitor in the future

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPluginFactory.java
aaron-congo: I have a question about this, see my comment [here](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1371#discussion_r2064368233)

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
aaron-congo: I'm fine with it, my only concern is that it might not be clear whether this method creates a full ConnectionWrapper, a pure target driver connection, or a "helper" connection created with forceConnect.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
aaron-congo: I've now refactored things so that there is a default stop() implementation in AbstractMonitor that concrete implementations can override if needed

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
aaron-congo: Wouldn't current connection just refer to the current monitoring connection in use? Or is this variable not intended to be used at all in monitoring connections? Should currentConnection from the class?

wrapper/src/main/java/software/amazon/jdbc/MonitorPluginService.java
aaron-congo: If we don't expect this to be used and don't want monitoring threads to call it I think we should throw an exception or log a warning and return (I guess return false?). For now I have the unexpected methods in this class throw exceptions so that I can see if they are being called in our integration tests, but we can decide whether we want them to log warnings instead.

wrapper/src/main/java/software/amazon/jdbc/util/CoreServicesContainer.java
```diff
@@ -0,0 +1,58 @@
...
+import software.amazon.jdbc.util.monitoring.CoreMonitorServiceImpl;
+import software.amazon.jdbc.util.storage.StorageService;
+import software.amazon.jdbc.util.storage.StorageServiceImpl;
+
+/**
+ * A singleton container object used to instantiate core services. This class should be used instead of directly
+ * instantiating core services so that only one instance of each service is instantiated.
+ */
+public class CoreServicesContainer {
+  private static final CoreServicesContainer INSTANCE = new CoreServicesContainer();
```
sergiyvamz, line 31: The instance will initialize immediately when the class is loaded. Is it desirable behaviour? Technically it's a singleton but with no delayed/lazy initialization.

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -105,6 +108,18 @@ public class Driver implements java.sql.Driver {
...
 
+  private final StorageService storageService;
+  private final CoreMonitorService monitorService;
+
+  public Driver() {
+    this(CoreServicesContainer.getInstance());
+  }
+
+  public Driver(CoreServicesContainer coreServicesContainer) {
+    this.storageService = coreServicesContainer.getStorageService();
```
sergiyvamz, line 119: We probably need to replace all references of storageService and monitorService to resolving them on fly.

wrapper/src/main/java/software/amazon/jdbc/Driver.java
```diff
@@ -216,14 +231,16 @@ public Connection connect(final String url, final Properties info) throws SQLExc
         effectiveConnectionProvider = configurationProfile.getConnectionProvider();
       }
 
+      ServiceContainer serviceContainer = new ServiceContainerImpl(storageService, monitorService, telemetryFactory);
```
sergiyvamz, line 234: Can we use a factory approach instead of direct instantiated a class? That's mostly for testing purposes.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
```diff
@@ -0,0 +1,768 @@
...
+    for (final HostSpec hostSpec : hosts) {
+      if (hostSpec.getRole() == HostRole.WRITER) {
+        return hostSpec;
+      }
+    }
+    return null;
+  }
+
+  @Override
+  @Deprecated
```
sergiyvamz, line 234: What's the point of marking unreleased method as deprecated?

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -34,92 +31,85 @@
...
+      CoreMonitorService monitorService,
+      StorageService storageService,
+      TelemetryFactory telemetryFactory,
       HostSpec customEndpointHostSpec,
       String endpointIdentifier,
       Region region,
       long refreshRateNano,
       BiFunction<HostSpec, Region, RdsClient> rdsClientFunc) {
-    this.pluginService = pluginService;
+    super(30);
```
sergiyvamz, line 87: Probably it's better to use a constant with human readable name rather than cryptic "30". It's a nit anyway.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -276,9 +276,9 @@ public List<HostSpec> getReaderHostsByPriority(final List<HostSpec> hosts) {
     hostsByPriority.addAll(downHostList);
 
     final int numOfReaders = activeReaders.size() + downHostList.size();
-    if (writerHost != null && (numOfReaders == 0
-          || this.pluginService.getDialect().getFailoverRestrictions()
-                .contains(FailoverRestriction.ENABLE_WRITER_IN_TASK_B))) {
+    final boolean enableWriterInTaskB =
```
sergiyvamz, line 279: Just a note.

By resolving value of this expression you disable code optimization. The following expression
may be optimized by compiler so if the first part of expression is true the remaining part is not calculated.

numOfReaders == 0 || enableWriterInTaskB

If you calculate this expression earlier then make the code less performant in some cases.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
aaron-congo: This method is deprecated in the regular plugin service (PluginServiceImpl) and shouldn't be used anywhere. Is it better to leave out the deprecated tag since its technically an unreleased method for this particular plugin service, or to keep the deprecated tag since it is deprecated in the regular plugin service and it should not be used? Regardless, I will change the implementation to throw an UnsupportedOperationException like the other methods we don't want called

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
aaron-congo: Good catch, this was uninentional

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: should be fixed now

sergiyvamz (approved):

Issue #1401, opened 2025-05-19
Set names for all internal threads


dnovitski:
### Describe the feature

Set descriptive names for each thread created by AWS JDBC wrapper, such as by these thread pools:

<img width="1535" alt="Image" src="https://github.com/user-attachments/assets/aeda9f40-3da5-45b6-9f8e-0d7458ea0718" />

### Use Case

While profiling I noticed generic thread names such as these:

<img width="752" alt="Image" src="https://github.com/user-attachments/assets/ff62cd4b-cdf8-4294-a6ff-7793ee40119f" />

Initially it was not clear what these were, but after debugging we found these threads are created by AWS JDBC wrapper. As there is no name set specifically, Java generates generic thread names.

It would be helpful and better practice if these threads can have a more descriptive name.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

2.3.9

### JDK version used

openjdk version "21" 2023-09-19 OpenJDK Runtime Environment Zulu21.28+85-CA (build 21+35) OpenJDK 64-Bit Server VM Zulu21.28+85-CA (build 21+35, mixed mode, sharing)

### Operating System and version

MacOS 15.3.2

hsuamz:
Thank you for reaching out @dnovitski.

Will discuss this with the team. The enhancement makes sense to me.

PR #1410, opened 2025-05-20
Set names for all internal threads

dnovitski:
### Summary

Implements issue https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1401 to give all threads a name

### Description

- Replaced Executors.newXXX with new ExecutorFactory util which takes thread name argument 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
@@ -31,7 +31,6 @@ import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -55,6 +54,7 @@ import software.amazon.jdbc.util.PropertyUtils;
 import software.amazon.jdbc.util.RdsUtils;
 import software.amazon.jdbc.util.StringUtils;
 import software.amazon.jdbc.util.SynchronousExecutor;
+import software.amazon.jdbc.util.ExecutorFactory;
 import software.amazon.jdbc.util.Utils;
 
 public class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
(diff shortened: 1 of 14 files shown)


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -109,14 +109,7 @@ public class ClusterTopologyMonitorImpl implements ClusterTopologyMonitor {
...
 
-  protected final ExecutorService monitorExecutor = Executors.newSingleThreadExecutor(runnableTarget -> {
-    final Thread monitoringThread = new Thread(runnableTarget);
-    monitoringThread.setDaemon(true);
-    if (!StringUtils.isNullOrEmpty(monitoringThread.getName())) {
-      monitoringThread.setName(monitoringThread.getName() + "-m");
-    }
-    return monitoringThread;
-  });
+  protected final ExecutorService monitorExecutor = ExecutorFactory.newSingleThreadExecutor("ClusterTopologyMonitorImpl#monitorExecutor");
```
sergiyvamz, line 112: We'd like to use daemon threads for all monitoring purposes.

sergiyvamz: If user application uses more than 1 DB cluster then there'll be more than 1 monitoring topology monitors. Monitoring thread should take it into account and avoid using the same name for different threads.

wrapper/src/main/java/software/amazon/jdbc/util/ExecutorFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public static ExecutorService newCachedThreadPool(String threadName) {
+    return Executors.newCachedThreadPool(getThreadFactory(threadName));
+  }
+
+  public static ExecutorService newFixedThreadPool(int nThreads, String threadName) {
+    return Executors.newFixedThreadPool(nThreads, getThreadFactory(threadName));
+  }
+
+  private static ThreadFactory getThreadFactory(String threadName) {
```
sergiyvamz, line 36: There're few places where similar monitoring threads are opened against different nodes. It seems that those threads will have the same name. Any chance to avoid using the same name for different threads?

dnovitski: Good point. 

I've added a static thread factory map so threads with the same name will use the same thread factory which has an atomic counter for each new thread created to differentiate them

dnovitski:
Example screenshot from Visual VM

Screenshot 2025-05-21 at 13 15 09

dnovitski:
Checkstyle issues should be fixed now

wrapper/src/main/java/software/amazon/jdbc/util/ExecutorFactory.java
```diff
@@ -0,0 +1,54 @@
...
+    return THREAD_FACTORY_MAP.computeIfAbsent(threadName, ExecutorFactory::createThreadFactory);
+  }
+
+  private static ThreadFactory createThreadFactory(String threadName) {
+    AtomicLong threadCounter = new AtomicLong();
+    return runnable -> {
+      String formattedThreadName = String.format("%s %s-%d", "AWS JDBC wrapper",
```
karenc-bq, line 47:
```suggestion
      String formattedThreadName = String.format("%s %s-%d", "AWS JDBC Driver",
```

wrapper/src/main/java/software/amazon/jdbc/util/ExecutorFactory.java
sergiyvamz: Most of driver logs include thread names and the only concern here is to make them less readable. Is there a chance to come up to a shorter thread names? Potentially we can completely rid of class name. Corresponding class name and method could mentioned in the documentation.

dnovitski: Fair point as well. I've removed the `classname#` part of the thread name and shortened it while still keeping some descriptive name for it.

PR #1422, opened 2025-05-28
fix: add expiration time for cached reader connections

karenc-bq:
### Summary

If connection pool isn't used, the read/write splitting plugin caches the reader connection forever, all read operations will be redirected to a single instance when `setReadOnly(true)` is called. This may impede performance.
This PR provides a connection parameter allowing user applications to redirect read operations to other instances after certain expiration time.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 import java.util.logging.Level;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.jdbc.AwsWrapperProperty;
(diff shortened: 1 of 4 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -428,7 +441,10 @@ private void switchToReaderConnection(final List<HostSpec> hosts)
               new Object[] {this.readerHostSpec.getUrl()}));
         }
 
-        this.readerConnection.close();
+        Connection conn = this.readerConnection.get(true);
+        if (conn != null) {
+          conn.close();
```
sergiyvamz, line 446: try-catch ?

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -506,10 +522,22 @@ private void getNewReaderConnection() throws SQLException {
     switchCurrentConnectionTo(this.readerConnection, this.readerHostSpec);
   }
 
+  private boolean isConnectionUsable(final CacheItem<Connection> cachedConnection) throws SQLException {
```
sergiyvamz, line 525: Can you combine it with the following method?

karenc-bq: I don't think so without using templates, which doesn't really improve the performance.

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -381,6 +388,12 @@ private void switchToWriterConnection(
         new Object[] {writerHost.getUrl()}));
   }
 
+  private void switchCurrentConnectionTo(
```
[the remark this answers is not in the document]
karenc-bq: To reduce the number of changes required throughout the whole file.

PR #1441, opened 2025-06-13
fix: minor Blue/Green F

aaron-congo:
Minor Blue/Green fixes

- adds null/empty checks
- adds some local variables to reuse retrieved values that have been verified to be non-empty/non-null
- fixes a minor bug where the wrong key was being used to add some B/G logging info

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
@@ -174,7 +174,10 @@ public class BlueGreenConnectionPlugin extends AbstractConnectionPlugin {
 
         if (conn == null) {
 
-          this.bgStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          BlueGreenStatus latestStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          if (latestStatus != null) {
+            this.bgStatus = latestStatus;
+          }
 
           routing = this.bgStatus.getConnectRouting().stream()
               .filter(r -> r.isMatch(hostSpec, hostRole))
@@ -276,7 +279,10 @@ public class BlueGreenConnectionPlugin extends AbstractConnectionPlugin {
 
         if (!result.isPresent()) {
 
-          this.bgStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          BlueGreenStatus latestStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          if (latestStatus != null) {
+            this.bgStatus = latestStatus;
+          }
 
           routing = this.bgStatus.getExecuteRouting().stream()
               .filter(r -> r.isMatch(currentHostSpec, hostRole))
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
@@ -190,14 +190,14 @@ public class BlueGreenStatusProvider {
   }
 
   protected void prepareStatus(
-      final BlueGreenRole role,
-      final BlueGreenInterimStatus interimStatus) {
+      final @NonNull BlueGreenRole role,
+      final @NonNull BlueGreenInterimStatus interimStatus) {
 
     this.processStatusLock.lock();
     try {
 
       // Detect changes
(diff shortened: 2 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
```diff
@@ -171,7 +171,10 @@ public Connection connect(
 
         if (conn == null) {
 
-          this.bgStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          BlueGreenStatus latestStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
```
sergiyvamz: That increases a chance to use a stale BG status. No BG status means that BGD doesn't exist so no routing should be applied.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -608,31 +617,43 @@ protected BlueGreenStatus getStatusOfPreparation() {
...
-              x.getValue(),
-              substituteHostSpecWithIp,
-              Collections.singletonList(hostSpec),
-              null));
-        });
+    for (Map.Entry<String, BlueGreenRole> entry : this.roleByHost.entrySet()) {
+      String host = entry.getKey();
+      BlueGreenRole role = entry.getValue();
+      Pair<HostSpec, HostSpec> nodePair = this.correspondingNodes.get(host);
+      if (role == BlueGreenRole.TARGET || nodePair == null) {
```
sergiyvamz: I'd suggest the following code as it's more readable:

`if (role != BlueGreenRole.SOURCE || nodePair == null) {`

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -826,13 +849,16 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
                   x.getValue(),
                   this.bgdId));
 
+              BlueGreenInterimStatus interimStatus = this.interimStatuses[x.getValue().getValue()];
+              if (interimStatus == null) {
```
sergiyvamz:
```
if (interimStatus != null) {
   connectRouting.add( ... )
}
```

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -854,10 +885,14 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
                   iamHosts,
                   isBlueHostInstance ? (iamHost) -> this.registerIamHost(greenHost, iamHost) : null));
 
+              BlueGreenInterimStatus interimStatus = this.interimStatuses[x.getValue().getValue()];
+              if (interimStatus == null) {
```
sergiyvamz:
```
if (interimStatus != null) {
   connectRouting.add( ... );
}
```

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
```diff
@@ -168,10 +168,12 @@ public Connection connect(
             isInitialConnection,
             connectFunc,
             this.pluginService);
-
         if (conn == null) {
 
           this.bgStatus = this.pluginService.getStatus(BlueGreenStatus.class, this.bgdId);
+          if (this.bgStatus == null) {
```
sergiyvamz, line 174: `this.endTimeNano` won't be set up in such case.

aaron-congo: I think it will, when the return statement is hit we will automatically enter the finally block which sets the endTimeNano if it hasn't been set yet. But let me know if I'm incorrect

sergiyvamz: Then the measured time includes an actual connect time. Original idea was to measure that routing time. If it's easy to fix I'd prefer just routing time.

aaron-congo: I see, okay I've updated the code to set the end time before connecting

PR #1451, opened 2025-06-25
chore: various warning fixes

sergiyvamz:
### Additional Reviewers

@karenc-bq 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/exceptions/ExceptionHandler.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/exceptions/ExceptionHandler.java
@@ -29,7 +29,7 @@ public interface ExceptionHandler {
    *
    * @deprecated Use similar method below that accepts throwable and target driver dialect.
    */
-  @Deprecated()
+  @Deprecated
   boolean isNetworkException(Throwable throwable);
 
   boolean isNetworkException(Throwable throwable, @Nullable TargetDriverDialect targetDriverDialect);


wrapper/build.gradle.kts
```diff
@@ -26,6 +25,12 @@ plugins {
     id("com.kncept.junit.reporter")
 }
 
+var useJacoco = (!project.hasProperty("jacocoEnabled") || project.property("jacocoEnabled").toString().toBoolean())
```
karenc-bq, line 28: Should this be
```
project.hasProperty("jacocoEnabled") && project.property("jacocoEnabled").toString().toBoolean()
```

wrapper/src/main/java/software/amazon/jdbc/exceptions/ExceptionHandler.java
```diff
@@ -24,6 +24,9 @@ public interface ExceptionHandler {
   /**
    * The method determines whether provided throwable is about any network issues.
    *
+   * @param throwable A throwable object to check.
+   * @return true, if a provided throwable object is network-related.
+   *
    * @deprecated Use similar method below that accepts throwable and target driver dialect.
    */
   @Deprecated()
```
karenc-bq:
```suggestion
  @Deprecated
```

wrapper/build.gradle.kts
sergiyvamz: I tried to make jacocoEnabled parameter optional and enable jacoco by default. That'd mimic the current logic if parameter isn't specified. You proposing an opposite logic where jacocoEnabled should be explicitly specified in order to enable it.

PR #1477, opened 2025-07-16
feat: add plugin for DSQL auth tokens

danielfrankcom:
### Summary

Adds a new `dsql` plugin which creates authentication tokens for [Amazon Aurora DSQL](https://aws.amazon.com/rds/aurora/dsql/).

### Description

Changes include the added `dsql` plugin, which is an extension of the existing `IamAuthConnectionPlugin` with a custom token generator. Both new files are unit tested.

A new dependency was added which customers will also need if using the plugin. Requirements will be documented in an upcoming change.

Credit to @karenc-bq for the initial plugin implementation.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginFactory.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginFactory.java
@@ -17,14 +17,18 @@
 package software.amazon.jdbc.plugin.iam;
 
 import java.util.Properties;
+import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.jdbc.ConnectionPlugin;
 import software.amazon.jdbc.ConnectionPluginFactory;
 import software.amazon.jdbc.PluginService;
 import software.amazon.jdbc.util.IamAuthUtils;
 
-public class IamAuthConnectionPluginFactory implements ConnectionPluginFactory {
+/**
+ * Provides {@link ConnectionPlugin} instances which can be used to connect to Amazon Aurora DSQL.
+ */
+public class DsqlIamConnectionPluginFactory implements ConnectionPluginFactory {
   @Override
-  public ConnectionPlugin getInstance(final PluginService pluginService, final Properties props) {
-    return new IamAuthConnectionPlugin(pluginService, IamAuthUtils.getRdsTokenUtility());
+  public ConnectionPlugin getInstance(@NonNull final PluginService pluginService, @NonNull final Properties props) {
+    return new IamAuthConnectionPlugin(pluginService, IamAuthUtils.getDsqlTokenUtility());
   }
 }
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlTokenUtility.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlTokenUtility.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.plugin.iam;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.dsql.DsqlUtilities;
+
+/**
+ * Represents an {@link IamTokenUtility} which provides auth tokens for connecting to Amazon Aurora DSQL.
+ */
+public class DsqlTokenUtility implements IamTokenUtility {
+
+  public DsqlTokenUtility() { }
+
+  @Override
+  public String generateAuthenticationToken(
+      @NonNull final AwsCredentialsProvider credentialsProvider,
+      @NonNull final Region region,
+      @NonNull final String hostname,
+      final int port,
+      @NonNull final String username) {
+    final DsqlUtilities utilities = DsqlUtilities.builder()
+        .credentialsProvider(credentialsProvider)
+        .region(region)
+        .build();
+
+    if (username.equals("admin")) {
+      return utilities.generateDbConnectAdminAuthToken((builder) ->
+          builder.hostname(hostname).region(region)
+      );
+    } else {
+      return utilities.generateDbConnectAuthToken((builder) ->
+          builder.hostname(hostname).region(region)
+      );
+    }
+  }
+}
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
@@ -86,11 +86,7 @@ public class IamAuthConnectionPlugin extends AbstractConnectionPlugin {
 
   private final IamTokenUtility iamTokenUtility;
 
-  public IamAuthConnectionPlugin(final @NonNull PluginService pluginService) {
-    this(pluginService, IamAuthUtils.getTokenUtility());
-  }
-
-  IamAuthConnectionPlugin(final @NonNull PluginService pluginService, IamTokenUtility utility) {
+  public IamAuthConnectionPlugin(final @NonNull PluginService pluginService, final IamTokenUtility utility) {
     this.iamTokenUtility = utility;
     this.pluginService = pluginService;
     this.telemetryFactory = pluginService.getTelemetryFactory();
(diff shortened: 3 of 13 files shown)


danielfrankcom:
Latest change is a simple rebase to update the branch with the latest `main`.

wrapper/src/test/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginTest.java
```diff
@@ -0,0 +1,142 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.plugin.iam;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.*;
```
karenc-bq: Please avoid * imports

danielfrankcom: Should be addressed by [e07c358](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/e07c3581471398b783a180bd4f69c139d85d89b5)

danielfrankcom:
Update to [e07c358](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/e07c3581471398b783a180bd4f69c139d85d89b5) addresses a few issues:

- Avoid wildcard imports as requested
- Add support for extracting region from connection string for DSQL
- Fix test issue caused by rebase

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlTokenUtility.java
```diff
@@ -0,0 +1,61 @@
...
+public class DsqlTokenUtility implements IamTokenUtility {
+
+  private DsqlUtilities utilities = null;
+
+  public DsqlTokenUtility() { }
+
+  // For testing only
+  DsqlTokenUtility(@NonNull final DsqlUtilities utilities) {
+    this.utilities = utilities;
+  }
```
deepakscomk: can we remove this and use static class mocks instead ?

danielfrankcom: Should be addressed by [bb1f7c5](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/bb1f7c54ac16df95de45065ecd90d23b9f728d37). This is cool, didn't know Mockito could do that.

danielfrankcom:
Update to [a1c6593](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/a1c6593b98bc883ce4ee9ec3ddf9213ed42d1bf1) fixes some formatting issues which were flagged by the GitHub actions.

danielfrankcom:
Update to [7c9c3fc](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/7c9c3fc33825001ddb33451495b120a30d352f22) removes a unit test dependency on the system AWS credentials provider.

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import software.amazon.jdbc.ConnectionPluginFactory;
+import software.amazon.jdbc.PluginService;
+
+/**
+ * Provides {@link ConnectionPlugin} instances which can be used to connect to Amazon Aurora DSQL.
+ */
+public class DsqlIamConnectionPluginFactory implements ConnectionPluginFactory {
+  @Override
+  public ConnectionPlugin getInstance(@NonNull final PluginService pluginService, @NonNull final Properties props) {
+    return new IamAuthConnectionPlugin(pluginService, new DsqlTokenUtility());
```
sergiyvamz: `DsqlTokenUtility` requires AWS SDK dependencies. Since `DsqlIamConnectionPluginFactory` is instantiated all the time, despite the user uses the plugin or not, this AWS SDK dependency becomes mandatory. A simplest fix would be for replace a token utility instance with supplier lambda and instantiate it later when the plugin (not the plugin factory) is loaded.

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlTokenUtility.java
```diff
@@ -0,0 +1,56 @@
...
+      @NonNull final String hostname,
+      final int port,
+      @NonNull final String username) {
+    if (this.utilities == null) {
+      this.utilities = DsqlUtilities.builder()
+          .credentialsProvider(credentialsProvider)
+          .region(region)
+          .build();
+    }
+    if (username.equals("admin")) {
```
sergiyvamz, line 43: equalsIgnoreCase() ?

danielfrankcom: I don't think so, DSQL seems to be picky about the username case.

```
> PGUSER=ADMIN psql

psql: error: connection to server at "..." (...), port 5432 failed: FATAL:  unable to accept connection, access denied
DETAIL:  Session Id: ...
HINT:  Wrong user to action mapping. user: ADMIN, action: DbConnectAdmin
connection to server at "..." (...), port 5432 failed: FATAL:  unable to accept connection, SSL is mandatory. AWS Authentication is required.
DETAIL:  Session Id: ...
```

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginFactory.java
danielfrankcom: Unless I'm misunderstanding, I don't think this is a problem since the `DsqlTokenUtility` isn't used during construction. I ran the `AwsIamAuthenticationPostgresqlExample` with verbose class loading enabled to confirm, and the `DsqlIamConnectionPluginFactory` class is loaded but not the `DsqlTokenUtility`.

I could add a `getDsqlTokenUtility()` method to the `IamAuthUtils` and use that here instead, but I think that would have the same behavior in practice. Is there a different approach you would suggest here?

sergiyvamz: Right, you're correct. Utility class isn't instantiated right away when the factory is created so it doesn't require those dependencies. My bad. However we usually do a dependency check before instantiated a class that has some external dependencies and raise an error with a good message. An example is here. If it's possible to use a similar approach here that'd be great. https://github.com/aws/aws-advanced-jdbc-wrapper/blob/4df7ce4b1bd494f089b5478f5d8187f0bc8724fc/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java#L137

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlTokenUtility.java
sheludko166: DbConnectAdmin can be granted to any IAM role and linked to any DB username – it's not limited to "admin".

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/DsqlIamConnectionPluginFactory.java
danielfrankcom: Should be addressed by [3e6bd2a](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/3e6bd2ad5acbccd09ff0119bb4ff16d5a9f2e585).

danielfrankcom:
Update to [3e6bd2a](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/3e6bd2ad5acbccd09ff0119bb4ff16d5a9f2e585) addresses a comment by adding a user-facing warning message if the DSQL SDK dependency is missing.

danielfrankcom:
Update to [4dfd2a0](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/4dfd2a08a63601f4200a2b90e4fea761261d09b0) addresses an offline comment from @deepakscomk about a possible error for subsequent calls to `generateAuthenticationToken` with different `credentialsProvider` or `region`.

danielfrankcom:
Update to [8c3d0a7](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/8c3d0a7b1fef31e59615508deee502409348f5a6) performs a rebase to pull in a fix from `main` for a javadoc error.

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
```diff
@@ -86,11 +86,7 @@ public class IamAuthConnectionPlugin extends AbstractConnectionPlugin {
 
   private final IamTokenUtility iamTokenUtility;
 
-  public IamAuthConnectionPlugin(final @NonNull PluginService pluginService) {
-    this(pluginService, IamAuthUtils.getTokenUtility());
-  }
-
-  IamAuthConnectionPlugin(final @NonNull PluginService pluginService, IamTokenUtility utility) {
+  IamAuthConnectionPlugin(final @NonNull PluginService pluginService, final IamTokenUtility utility) {
```
sergiyvamz: nit: public

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
danielfrankcom: Should be addressed by [b879640](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/b879640dfac11f978d32d0e12cb3ee3f8a8f6ca4).

PR #1478, opened 2025-07-16
add user documentation for DSQL plugin

danielfrankcom:
### Summary

This PR adds user documentation and associated examples for using the AWS Aurora DSQL IAM authentication plugin.

This accompanies the change in  #1477, which adds the `dsql` plugin implementation and tests.

### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1481, opened 2025-07-17
add DSQL example for Spring Boot Hikari

danielfrankcom:
### Summary

This PR adds sample code for DSQL, which shows how to use the `dsql` plugin with Spring Boot and Hikari.

### Description

The change includes a new sample directory, which shows a basic Spring Boot application which connects to a DSQL cluster. This sample differs from the samples in #1478, as the plugins and database configuration and set in a `.yml` file which is consumed by the application framework.

I have validated this example locally against a DSQL cluster, but it requires #1477. This PR should not be merged until the dependency PR has been merged.

### DIFF
--- a/examples/SpringBootHikariDsqlExample/gradle.properties
+++ b/examples/SpringBootHikariDsqlExample/gradle.properties
@@ -1,0 +1,16 @@
+#    Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License").
+#    You may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+# Do not publish the Jar file for this subproject
+nexus.publish=false
--- a/examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
+++ b/examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
@@ -1,0 +1,14 @@
+spring:
+  datasource:
+    url: jdbc:aws-wrapper:postgresql://cluster-identifier.dsql.us-east-1.on.aws:5432/postgres
+    username: admin
+    driver-class-name: software.amazon.jdbc.Driver
+    hikari:
+      data-source-properties:
+        wrapperPlugins: iamDsql
+        wrapperDialect: pg
+      max-lifetime: 840000
+      minimum-idle: 20
+      maximum-pool-size: 20
+      idle-timeout: 900000
+      read-only: true
--- a/examples/SpringBootHikariDsqlExample/build.gradle.kts
+++ b/examples/SpringBootHikariDsqlExample/build.gradle.kts
@@ -15,14 +15,14 @@
  */
 
 plugins {
-	id("org.springframework.boot") version "2.7.0"
-	id("io.spring.dependency-management") version "1.1.7"
+    id("org.springframework.boot") version "2.7.0"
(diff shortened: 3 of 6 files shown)


examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
```diff
@@ -0,0 +1,14 @@
+spring:
+  datasource:
+    url: jdbc:aws-wrapper:postgresql://cluster-identifier.dsql.us-east-1.on.aws:5432/postgres
+    username: admin
+    driver-class-name: software.amazon.jdbc.Driver
+    hikari:
+      data-source-properties:
+        wrapperPlugins: iamDsql
+        wrapperDialect: pg
```
deepakscomk, line 9: can we integrate our custom dsql dialect into advance jdbc wrapper here , any thoughts ? this make customer experience simpler :) .

examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
```diff
@@ -0,0 +1,14 @@
+spring:
+  datasource:
+    url: jdbc:aws-wrapper:postgresql://cluster-identifier.dsql.us-east-1.on.aws:5432/postgres
```
deepakscomk, line 3: as discussed offline, can we think of simplifying this, like just passing the clusterId and region

examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
danielfrankcom: Discussed offline - the dialect configures the wrapper itself with details like default port which are shared by the existing PostgreSQL dialect.

examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
danielfrankcom: The `iamDsql` plugin introduced by this change focuses on the authentication aspect of connection to DSQL. Adding custom Spring components to simplify the `.yml` configuration would make the example code quite a bit more complex imo, and would result in users having to copy/paste more code to get started.

I believe it is possible to do what you are suggesting, but it would require a custom `@ConfigurationProperties` and also a custom `@Configuration`. This seems out of scope for the wrapper library to me, and if we wanted to provide those implementations for Spring then I would think we should have them in a dedicated library somewhere they can be referenced by anyone using Spring, not just those using this JDBC wrapper library.

danielfrankcom:
Update to [fe9b450](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/fe9b4506c35429b6b2116f9e6dc1592b67e317d3) is a simple rebase, mostly to trigger CI jobs.

examples/SpringBootHikariDsqlExample/src/main/resources/application.yml
danielfrankcom: We discussed this a fair amount offline. Summarizing here for future reference.

The goal is to have a way to specify things like the host and other details as individual properties, without specifying the JDBC URL. This is perceived as simpler for the user.

While this is possible with the `DataSource` approach shown in #1478, Spring seems to require the `url` property. If it is missing, the application refuses to start:

```
Failed to configure a DataSource: 'url' attribute is not specified and no embedded datasource could be configured.
Reason: Failed to determine suitable jdbc url
```

It is possible to define a custom data source as shown in [this gist](https://gist.github.com/danielfrankcom/7134a2255f671990548b64850eaa6484), but users would need to copy/paste the definitions into their application in order to use them. Ultimately it still creates a URL for Spring, just with a step between.

While it might make sense to publish a separate project to provide these data wrappers for users, it doesn't seem to belong in the `aws-advanced-jdbc-wrapper` given this is just a code sample for using the DSQL auth plugin.

examples/SpringBootHikariDsqlExample/gradle.properties
```diff
@@ -0,0 +1,16 @@
+#    Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License").
```
karenc-bq, line 3: Can we make sure the copyright headers are consistent across the sample files

PR #1483, opened 2025-07-18
run integration tests against DSQL

danielfrankcom:
### Summary

This PR modifies the existing integration tests and surrounding infrastructure to work with Aurora DSQL where they are compatible.

### Description

The changes include:

- Added feature flags to select tests which are compatible with DSQL
- Added logic to create/delete DSQL clusters as part of test environment configuration
- Enabled DSQL plugin for tests which are compatible with DSQL to handle authentication
- Added retry mechanism for DDL operations in `DataCachePluginTests` to avoid issues with DSQL's optimistic concurrency control

### DIFF
--- a/wrapper/src/test/java/integration/container/tests/AwsDsqlIntegrationTest.java
+++ b/wrapper/src/test/java/integration/container/tests/AwsDsqlIntegrationTest.java
@@ -16,28 +16,43 @@
 
 package integration.container.tests;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import com.amazonaws.arn.Arn;
 import integration.DriverHelper;
 import integration.TestEnvironmentFeatures;
 import integration.container.ConnectionStringHelper;
 import integration.container.TestDriverProvider;
 import integration.container.TestEnvironment;
 import integration.container.condition.EnableOnTestFeature;
+
 import java.sql.Connection;
 import java.sql.DriverManager;
+import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.Properties;
+import java.util.logging.Logger;
+
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.TestMethodOrder;
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.GetCallerIdentityResponse;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.HostSpecBuilder;
 import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.authentication.AwsCredentialsManager;
 import software.amazon.jdbc.ds.AwsWrapperDataSource;
+import software.amazon.jdbc.hostavailability.SimpleHostAvailabilityStrategy;
 import software.amazon.jdbc.plugin.iam.IamAuthConnectionPlugin;
 
 @TestMethodOrder(MethodOrderer.MethodName.class)
@@ -45,6 +60,8 @@ import software.amazon.jdbc.plugin.iam.IamAuthConnectionPlugin;
 @EnableOnTestFeature(TestEnvironmentFeatures.RUN_DSQL_TESTS_ONLY)
 @Order(3)
 public class AwsDsqlIntegrationTest {
+
+  private static final Logger LOGGER = Logger.getLogger(AwsDsqlIntegrationTest.class.getName());
 
   @BeforeEach
   public void beforeEach() {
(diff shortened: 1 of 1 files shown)


danielfrankcom:
Please note this change depends on #1477, and the tests will fail until the dependency is merged.

danielfrankcom:
Integration tests passed in a custom run [here](https://github.com/aws/aws-advanced-jdbc-wrapper/actions/runs/16426616321/job/46428484961), with all the changes integrated. Once the dependency PR has been merged they should pass here too.

wrapper/src/test/java/integration/host/TestEnvironment.java
```diff
@@ -93,6 +95,7 @@ public class TestEnvironment implements AutoCloseable {
   private String rdsDbName; // "cluster-mysql", "instance-name", "rds-multi-az-cluster-name"
   private String rdsDbDomain; // "XYZ.us-west-2.rds.amazonaws.com"
   private String rdsEndpoint; // "https://rds-int.amazon.com"
+  private String dsqlEndpoint; // "cluster-identifier.dsql.us-east-1.on.aws"
```
sergiyvamz: I'd prefer to reuse `rdsDbDomain` and `rdsDbName` if it's possible. If `rds` prefix in variable names looks odd we can remove it.

wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
```diff
@@ -186,6 +192,15 @@ public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContex
                                 || config.noIam
                                 ? null
                                 : TestEnvironmentFeatures.IAM,
+                            deployment == DatabaseEngineDeployment.DSQL
```
sergiyvamz, line 195: It seems that the following features are always on for DSQL:
REQUIRES_IAM_AUTH
REQUIRES_TLS_SNI
SUPPORTS_FOREIGN_KEYS

What is the purpose of them?

danielfrankcom: These are used to exclude specific tests from running. I could have added a feature that indicates DSQL is being used and excluded those tests based on that, but I thought it made more sense to have features for the actual reasons why they are incompatible with DSQL.

DSQL does not support foreign keys, it only supports IAM authentication (no password), and it requires the TLS SNI to be provided so it doesn't support connecting via IP address (unless you hard code the SNI for the IP address). These features are what makes the tests I annotated incompatible with DSQL.

I'm happy to change this to be DSQL-specific and remove these more granular features, but if any other database has a similar restriction in the future then we will need to add additional exclusions for that database too rather than reusing existing features.

wrapper/src/test/java/integration/container/ConnectionStringHelper.java
```diff
@@ -190,13 +192,38 @@ public static String getWrapperClusterEndpointUrl() {
...
+    final TestEnvironment env = TestEnvironment.getCurrent();
+    final TestEnvironmentInfo envInfo = TestEnvironment.getCurrent().getInfo();
     props.setProperty(PropertyDefinition.USER.name, envInfo.getDatabaseInfo().getUsername());
-    props.setProperty(PropertyDefinition.PASSWORD.name, envInfo.getDatabaseInfo().getPassword());
 
-    final Set<TestEnvironmentFeatures> features = envInfo.getRequest().getFeatures();
+    final TestEnvironmentRequest request = envInfo.getRequest();
+
+    // DSQL only supports IAM authentication.
+    if (request.getDatabaseEngineDeployment() == DatabaseEngineDeployment.DSQL) {
```
sergiyvamz: This logic is slightly off here. Would you consider to move it to other place? Usually each test set up a list of plugins depending on their needs. This method is for common (i.e. "never change") general properties.

wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
```diff
@@ -300,8 +303,13 @@ void test_TokenGenerators() {
   }
 
   protected Properties initAwsIamProps(String user, String password) {
+    final TestEnvironmentInfo envInfo = TestEnvironment.getCurrent().getInfo();
+    final DatabaseEngineDeployment deployment = envInfo.getRequest().getDatabaseEngineDeployment();
+    final boolean isDsql = (deployment == DatabaseEngineDeployment.DSQL);
+
     final Properties props = ConnectionStringHelper.getDefaultProperties();
-    props.setProperty(PropertyDefinition.PLUGINS.name, "iam");
+    props.setProperty(PropertyDefinition.PLUGINS.name, isDsql ? "iamDsql" : "iam");
```
sergiyvamz: I'd prefer a separate test class for DSQL IAM scenarios rather to make existing test more complex and more convoluted. Code repetition is quite possible in such a case but I believe that's acceptable price to pay to have test clean and straightforward.

wrapper/src/test/java/integration/container/tests/DataCachePluginTests.java
```diff
@@ -55,35 +56,91 @@ public class DataCachePluginTests {
...
   private static final Logger LOGGER = Logger.getLogger(DataCachePluginTests.class.getName());
 
+  private static final int ARBITRARY_NUM_RETRIES = 5;
+
   @BeforeEach
   public void beforeEach() {
     DataCacheConnectionPlugin.clearCache();
   }
 
+  private static void runWithRetries() {
```
sergiyvamz: Empty method? Do we need it?

wrapper/src/test/java/integration/container/ConnectionStringHelper.java
danielfrankcom: The reason I put this here is because there are quite a few tests which are compatible with DSQL, but don't know anything about the `iam` or `iamDsql` plugins.

This method gets a default `Properties` instance which contains a general connection configuration which is supposed to work with any DB theoretically. This default general-purpose config doesn't work with DSQL, because it doesn't support password authentication and therefore requires the `iamDsql` plugin to connect.

Without this change, we would either (a) not be able to run generally applicable tests against DSQL, or (b) need to change individual tests to know about the `iamDsql` plugin to have them work with DSQL.

I do agree that it feels a bit weird to modify this file in this way, but I'm not sure what a better alternative is. We could add some kind of configuration for the default plugin depending on the `DatabaseEngineDeployment`, but then we would still be specifying a password for DSQL which wouldn't break anything I don't think but also wouldn't make a lot of sense.

wrapper/src/test/java/integration/container/tests/DataCachePluginTests.java
```diff
@@ -55,35 +56,91 @@ public class DataCachePluginTests {
...
+  public void testQueryCacheable() throws SQLException, InterruptedException {
 
     DataCacheConnectionPlugin.clearCache();
 
     final Properties props = ConnectionStringHelper.getDefaultProperties();
     PropertyDefinition.CONNECT_TIMEOUT.set(props, "30000");
     PropertyDefinition.SOCKET_TIMEOUT.set(props, "30000");
 
-    props.setProperty(PropertyDefinition.PLUGINS.name, "dataCache");
+    ConnectionStringHelper.addPlugin(props, "dataCache");
```
sergiyvamz: I like a simplicity of original code.

sergiyvamz: DSQL IAM scenarios could be moved to a separate test class.

wrapper/src/test/java/integration/container/tests/DataCachePluginTests.java
danielfrankcom: Thanks, this is left over from a previous version and I missed it

danielfrankcom:
Update to [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19) reworks the integration test changes to separate the DSQL test logic, and avoid changes to other integration tests that don't support IAM.

wrapper/src/test/java/integration/container/tests/DataCachePluginTests.java
danielfrankcom: Should be addressed by [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19).

wrapper/src/test/java/integration/container/tests/DataCachePluginTests.java
danielfrankcom: Should be addressed by [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19).

wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
danielfrankcom: Should be addressed by [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19).

wrapper/src/test/java/integration/container/ConnectionStringHelper.java
danielfrankcom: Should be addressed by [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19).

wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
danielfrankcom: Should be addressed by [0a2f489](https://github.com/aws/aws-advanced-jdbc-wrapper/commit/0a2f489d8161525cb19278760d929ae240cbdd19).

wrapper/src/test/java/integration/host/TestEnvironment.java
danielfrankcom: Should be addressed by [c240647](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1483/commits/c2406472813828cdc9aec6244d7b7ada1444c42c).

The RDS prefix is a bit confusing but it's used in a lot of places so renaming it would complicate the diff here quite a bit. Given it's only for testing and use by this team, it should be clear enough as-is. I added some comments to the domain variables as a reminder it also supports the DSQL format.

sergiyvamz (approved):

danielfrankcom:
[2080b29](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1483/commits/2080b299232a066bdf9e032604af17c6bc6769a0) adds a new integration test which focuses on non-admin user connections.

Change is tested on integration branch [here](https://github.com/aws/aws-advanced-jdbc-wrapper/actions/runs/16510112031/job/46690173285).

Issue #1399, opened 2025-05-14
No wrapper class exists for 'com.pgvector.PGvector'


carolosf:
### Describe the feature

PGvector is useful for AI workloads but the lack of a wrapper class makes the AWS JDBC wrapper unusable for this kind of usecase. 

### Use Case

Retrieval-augmented generation support

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### The AWS Advanced JDBC Driver version used

software.amazon.jdbc:aws-advanced-jdbc-wrapper:2.5.6

### JDK version used

openjdk 22.0.1

### Operating System and version

linux

sergiyvamz:
Hi @carolosf 

Thank you reaching out with this issue. Would you mind to provide a concrete example of Java code that you'd like to be supported by AWS Wrapper?

carolosf:
Thanks for the response.

The use case I’m looking at is this:

https://github.com/quarkiverse/quarkus-langchain4j/tree/main/samples/movie-similarity-search

It should be possible to reproduce locally by just using the jdbc wrapper and setting jdbc properties.

sergiyvamz:
Hi @carolosf @onestardao 

We have implemented logic that allows the driver to accept third-party classes and pass them through without additional processing. We anticipate this enhancement will enable support for `PGVector`. While these changes have not yet been officially released, they are available in our snapshot builds. Could you please test this modification and provide your feedback? 

Thank you for your assistance!

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#enable-third-party-classes-and-packages

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md#using-a-snapshot-of-the-driver

onestardao:
thanks for the quick follow up. i’ll pull the snapshot and run a quick repro against our pgvector case. i’ll post logs and a minimal repro if i hit issues, otherwise i’ll report back that it’s fixed. much appreciated.

sergiyvamz:
Hi @carolosf 

Thank you for prompt check of our fix. I might guess those unexplained behaviour is caused by environment misconfigurations. And it sounds like the latest code works well for you. I'll keep you updated when the fix is released.

PR #1494, opened 2025-07-30
feat: add support of PGvector

sergiyvamz:
### Summary

Allow user to specify classes and java packages that don't require wrapping.
Specifying `PGvector` or `PostGIS` classes could be a good example of use.
https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1399


wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -177,6 +177,12 @@ public class PropertyDefinition {
       "Allow the driver to optimized work with ResultSets and skip tracking some of their methods"
           + " by assuming that entire data result are fetched at once.");
 
+  public static final AwsWrapperProperty SKIP_WRAPPING_FOR_PACKAGES =
+      new AwsWrapperProperty(
+          "skipWrappingForPackages", null,
+          "Register Java package names (coma separated list) which will be left unwrapped."
```
karenc-bq:
```suggestion
          "Register Java package names (comma separated list) which will be left unwrapped."
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -392,3 +393,18 @@ public static void main(String[] args) throws SQLException {
 
 ## Enable Logging
 To enable logging in the AWS JDBC Driver, change the `logger=StandardLogger` parameter to `wrapperLoggerLevel=FINEST`
+
+
+## Enable Third Party Classes and Packages
+A user application may require a special handling for particular classes or Java packages. In such case, wrapping of data objects by the AWS Advanced JDBC Driver might be crucial for a user application. The driver allows to specify classes and packages that should be left intact with no wrapping. Use the following code snippet as an example to register classes or an entire package if needed.
```
karenc-bq: The first two sentences might be a bit confusing due to lack of context, what do you think of
```suggestion
The AWS JDBC Driver performs special handling before or after operating on a database connection. For example, the IAM Authentication plugin generates a special IAM authentication token before establishing a connection. This is achievable by the AWS JDBC Driver wrapper over relevant data objects.

Not all data objects require special handling. Therefore, they can be left unwrapped. An example of this are classes from the PGVector package.

The driver allows you to specify classes and packages that should be left intact with no wrapping. Use the following code snippet as an example to register classes or an entire package if needed.
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -392,3 +393,18 @@ public static void main(String[] args) throws SQLException {
...
+
+```java
+Driver.skipWrappingForType(com.pgvector.PGvector.class);
+Driver.skipWrappingForType(com.pgvector.PGhalfvec.class);
+Driver.skipWrappingForType(com.pgvector.PGbit.class);
+    
+Driver.skipWrappingForPackage("com.pgvector");
+```
+This feature can be used to allow the AWS Advanced JDBC Driver to properly handle popular database extensions like [PGvector](https://github.com/pgvector/pgvector-java) and [PostGIS](https://github.com/postgis/postgis-java).
```
karenc-bq, line 409: Curious if we need to expand on the consequences of wrapping these packages? If there are any.

wrapper/src/main/java/software/amazon/jdbc/PropertyDefinition.java
```diff
@@ -177,6 +177,12 @@ public class PropertyDefinition {
       "Allow the driver to optimized work with ResultSets and skip tracking some of their methods"
           + " by assuming that entire data result are fetched at once.");
 
+  public static final AwsWrapperProperty SKIP_WRAPPING_FOR_PACKAGES =
+      new AwsWrapperProperty(
+          "skipWrappingForPackages", null,
+          "Register Java package names (comma separated list) which will be left unwrapped."
+              + " This setting is global for the driver, not for a particular connection.");
```
karenc-bq:
```suggestion
              + " This setting modifies all future connections established by the driver, not just the current connection using this set of connection properties.");
```

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperDataSource.java
```diff
@@ -119,6 +119,14 @@ public Connection getConnection(final String username, final String password) th
 
     final Properties props = PropertyUtils.copyProperties(this.targetDataSourceProperties);
 
+    final String registerPackageNames = PropertyDefinition.SKIP_WRAPPING_FOR_PACKAGES.getString(props);
+    if (!StringUtils.isNullOrEmpty(registerPackageNames)) {
+      String[] packages = registerPackageNames.split(",");
```
karenc-bq:
```suggestion
      final String[] packages = registerPackageNames.split(",");
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -91,6 +91,7 @@ These parameters are applicable to any instance of the AWS JDBC Driver.
...
 | `wrapperCaseSensitive`,<br>`wrappercasesensitive` | `Boolean` | No       | Allows the driver to change case sensitivity for parameter names in the connection string and in connection properties. Set parameter to `false` to allow case-insensitive parameter names.                                                                                                                                                          | `true`        |
+| `skipWrappingForPackages`                         | `String`  | No       | Register Java package names (separated by comma) which will be left unwrapped. This setting is global for the driver, not for a particular connection.                                                                                                                                                                                               | `true`        |
```
karenc-bq:
```suggestion
| `skipWrappingForPackages`                         | `String`  | No       | Register Java package names (separated by comma) which will be left unwrapped. This setting modifies all future connections established by the driver, not just a particular connection.                                                                                                                                                                                              | `true`        |
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -392,3 +393,18 @@ public static void main(String[] args) throws SQLException {
 
 ## Enable Logging
 To enable logging in the AWS JDBC Driver, change the `logger=StandardLogger` parameter to `wrapperLoggerLevel=FINEST`
+
+
```
sophia-bq:
```suggestion
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
sophia-bq:
```suggestion
Depending on the requirements of a user's application, the AWS Advanced JDBC Driver performs special handling before and after operating on a database connection. This is achievable by wrapping over relevant data objects. For example, the IAM Authentication plugin generates a special IAM authentication token before establishing a connection. 

Not all data objects require special handling and thus can be left unwrapped. One example being classes from the PGVector package. The driver allows you to specify such classes and packages that should be left intact with no wrapping. If needed, use the following code snippet as an example to register data objects that should be left unwrapped.
```

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
sophia-bq:
```suggestion
| `skipWrappingForPackages`                         | `String`  | No       | Register Java package names (separated by comma) which will be left unwrapped. This setting is global for the driver, not for a particular connection.                                                                                                                                                                                               | `com.pgvector`        |
```
Might even be best to have another one in the example to show the comma separated list

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
sophia-bq: If there are consequences it might be worth documenting them

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -392,3 +393,18 @@ public static void main(String[] args) throws SQLException {
...
+```java
+Driver.skipWrappingForType(com.pgvector.PGvector.class);
+Driver.skipWrappingForType(com.pgvector.PGhalfvec.class);
+Driver.skipWrappingForType(com.pgvector.PGbit.class);
+    
+Driver.skipWrappingForPackage("com.pgvector");
+```
+This feature can be used to allow the AWS Advanced JDBC Driver to properly handle popular database extensions like [PGvector](https://github.com/pgvector/pgvector-java) and [PostGIS](https://github.com/postgis/postgis-java).
+
+Using `Driver.skipWrappingForPackage()` method and using driver configuration parameter `skipWrappingForPackages` are functionally similar. The configuration parameter receives a comma separated list of package names while `Driver.skipWrappingForPackage()` accepts just one package at at time.
```
sophia-bq, line 411: I'm curious on the choices behind this. It seems there is 1 way for skipWrappingForType and 2 ways for skipWrappingForPackage. If the reason we introduced skipWrappingForPackages is to be easier than writing a line of code for each package - we should definitely have that for type as well as already in this example the number of types to skip is greater than the number of packages. Suggestion: either one way or two ways for both. 

And, if we still want parameters.. 
i. should clarify whether or not skipWrappingForType has an equivalent? 
ii. should have an example of using the parameters?

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
sergiyvamz: I'm not sure what consequences it could be. Either the driver throw an exception since it doesn't have any suitable wrapper class, or it passes this classes intact. We can adjust the messages slightly:

`This feature can be used to allow the AWS Advanced JDBC Driver to pass classes intact and, thus, allow support of  popular database extensions like [PGvector](https://github.com/pgvector/pgvector-java) and [PostGIS](https://github.com/postgis/postgis-java).`

How does it sound?

docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
sergiyvamz: The intention was to let a user a solution to provide an exceptional classes. The current solution with `Driver.skipWrappingForType()` allows to do it with highest possible granularity. Some users may needs such precision due to high security requirements. A less granular and easier solution is to specify entire package name with `Driver.skipWrappingForPackage()`. It allows more classes to be left intact so it could be considered as a less secure approach. 

Both approaches above require a user to have access to source code of an application and that is not always the case. Many of out-of-box AI solutions provides a free access to pre-built versions, not to the sources. For such cases there's a configuration parameter `skipWrappingForPackages`.

sophia-bq (approved):

PR #1495, opened 2025-07-30
Adapt efm2, limitless, and fastest response plugins to refactoring work

aaron-congo:


wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouters.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package software.amazon.jdbc.plugin.limitless;
+
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import software.amazon.jdbc.HostSpec;
+import software.amazon.jdbc.hostlistprovider.Topology;
+
+public class LimitlessRouters extends Topology {
```
sergiyvamz: I wouldn't extend it from Topology. Inheriting form Topology may give an impression that there's logic relations between these two classes. Practically they are completely unrelated.

wrapper/src/main/java/software/amazon/jdbc/plugin/strategy/fastestresponse/NodeResponseTimeMonitor.java
```diff
@@ -85,55 +84,38 @@ public NodeResponseTimeMonitor(
 
     // Report current response time (in milliseconds) to telemetry engine.
     // Report -1 if response time couldn't be measured.
-    this.responseTimeMsGauge = this.telemetryFactory.createGauge(
+    // TODO: this is not used anywhere, should we remove it?
```
aaron-congo: Should we remove this gauge or add logic to actually use it (it is currently unused)?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/HostMonitorImpl.java
```diff
@@ -335,7 +335,6 @@ ConnectionStatus checkConnectionStatus(final long shortestFailureDetectionInterv
 
         LOGGER.finest(() -> "Opening a monitoring connection to " + this.hostSpec.getUrl());
         startNano = this.getCurrentTimeNano();
-        // TODO: replace with ConnectionService#open
```
aaron-congo: I believe we decided to not adapt the EFM1 monitors because they are old, have been superseded by EFM2, and do not fit the structure of the refactoring well. However, should we change this to use ConnectionService instead of forceConnect?

sergiyvamz: I wouldn't consider EFM2 as a successor of EFM1. They're both functional and EFM1 isn't an old/outdated one. I'd still plan to migrate EFM1 to new services. It might take a bit more efforts to adopt it.

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
```diff
@@ -45,69 +49,51 @@ public class LimitlessRouterServiceImpl implements LimitlessRouterService {
...
-            intervalMs) ->
-            new LimitlessRouterMonitor(
-                pluginService,
-                hostSpec,
-                routerCache,
-                routerCacheKey,
-                props,
-                intervalMs),
-        new LimitlessQueryHelper(pluginService));
+  public LimitlessRouterServiceImpl(
```
sergiyvamz, line 63: Do we need `closeAllMonitors()` method similar to `HostMonitorServiceImpl.closeAllMonitors()`?

wrapper/src/main/java/software/amazon/jdbc/plugin/strategy/fastestresponse/NodeResponseTimeMonitor.java
sergiyvamz: The gauge is initialized and call provided lambda to get values and report them. It's in use.

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/strategy/fastestresponse/NodeResponseTimeMonitor.java
aaron-congo: Ah okay, sounds like the callback is used internally, didn't realize

wrapper/src/main/java/software/amazon/jdbc/plugin/limitless/LimitlessRouterServiceImpl.java
aaron-congo: I don't think so, HostMonitorServiceImpl.closeAllMonitors() is only called in the performance tests and in Driver.releaseResources, but Driver.releaseResources already contains a call to stop all monitors in the monitor service

PR #1514, opened 2025-08-20
[wip] refactor: failover handler tasks use separate PluginService instances

aaron-congo:
refactor: failover handler tasks use separate PluginService instances


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -124,38 +130,43 @@ protected void setTimeoutMs(final int timeoutMs) {
...
    */
   @Override
-  public ReaderFailoverResult failover(final List<HostSpec> hosts, final HostSpec currentHost)
+  public ReaderFailoverResult failover(
+      final ConnectionService connectionService, final List<HostSpec> hosts, final HostSpec currentHost)
       throws SQLException {
     if (Utils.isNullOrEmpty(hosts)) {
       LOGGER.fine(() -> Messages.get("ClusterAwareReaderFailoverHandler.invalidTopology", new Object[] {"failover"}));
-      return FAILED_READER_FAILOVER_RESULT;
+      return new ReaderFailoverResult(null, null, false, null, null);
```
sergiyvamz: why?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -164,24 +175,25 @@ private Future<ReaderFailoverResult> submitInternalFailoverTask(
...
-      final Future<ReaderFailoverResult> future) throws SQLException {
+      final Future<ReaderFailoverResult> future,
+      final Map<String, HostAvailability> availabilityMap) throws SQLException {
     try {
       final ReaderFailoverResult result = future.get(this.maxFailoverTimeoutMs, TimeUnit.MILLISECONDS);
       if (result == null) {
         LOGGER.warning(
             Messages.get("ClusterAwareReaderFailoverHandler.timeout", new Object[] {this.maxFailoverTimeoutMs}));
-        return FAILED_READER_FAILOVER_RESULT;
+        return new ReaderFailoverResult(null, null, false, null, availabilityMap);
```
sergiyvamz: Do we need to return availabilityMap?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -239,18 +255,20 @@ public List<HostSpec> getHostsByPriority(final List<HostSpec> hosts) {
...
           () -> Messages.get(
               "ClusterAwareReaderFailover.invalidTopology",
               new Object[] {"getReaderConnection"}));
-      return FAILED_READER_FAILOVER_RESULT;
+      return new ReaderFailoverResult(null, null, false, null, null);
     }
 
     final List<HostSpec> hostsByPriority = getReaderHostsByPriority(hostList);
-    return getConnectionFromHostGroup(hostsByPriority);
+    final Map<String, HostAvailability> availabilityMap = new ConcurrentHashMap<>();
```
sergiyvamz: I think we can initialize this map in the constructor.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -610,6 +619,23 @@ protected <E extends Exception> void dealWithIllegalStateException(
   protected void failover(final HostSpec failedHost) throws SQLException {
     this.pluginService.setAvailability(failedHost.asAliases(), HostAvailability.NOT_AVAILABLE);
 
+    if (this.connectionService == null) {
+      TargetDriverHelper helper = new TargetDriverHelper();
+      java.sql.Driver driver = helper.getTargetDriver(this.pluginService.getOriginalUrl(), properties);
+      final ConnectionProvider defaultConnectionProvider = new DriverConnectionProvider(driver);
```
sergiyvamz: I believe that default connection provider is stateless and can be reused.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ReaderFailoverHandler.java
```diff
@@ -36,7 +37,8 @@ public interface ReaderFailoverHandler {
    * @return {@link ReaderFailoverResult} The results of this process.
    * @throws SQLException indicating whether the failover attempt was successful.
    */
-  ReaderFailoverResult failover(List<HostSpec> hosts, HostSpec currentHost) throws SQLException;
+  ReaderFailoverResult failover(
+      ConnectionService connectionService, List<HostSpec> hosts, HostSpec currentHost) throws SQLException;
```
sergiyvamz: Can we pass ConnectionService as constructor parameter parameter to a class that implements this interface? I think the reference to ConnectionService never changes.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -791,6 +823,20 @@ protected void failoverWriter() throws SQLException {
     }
   }
 
+  private void updateHostAvailability(Map<String, HostAvailability> hostAvailabilityMap) {
+    if (hostAvailabilityMap != null && !hostAvailabilityMap.isEmpty()) {
```
sergiyvamz: I feel that it'd be easier to iterate through hostAvailabilityMap and apply availability status to hosts from pluginService.

sergiyvamz: Nested loops :(

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: The majority of cases return the availability map, there are only two places in the file where we return null for the availability map arg. I can keep the FAILED_READER_FAILOVER_RESULT variable if you'd like but we will only be using it in 2 places.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: We don't need to but even if the failover task timed out the map may still have info about which hosts are up/down which could be used. Although maybe this isn't useful if failover failed since we're going to be throwing a FailoverFailed exception anyway?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: We could but then we'd have to make sure it was cleared after each failover because we do not want there to be data from a previous failover in the map if a new failover takes place. Let me know if you prefer this still.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
aaron-congo: I think you're right but I'm not sure how to grab it here. `this.pluginService.getConnectionProvider()` would grab it but that method is deprecated.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ReaderFailoverHandler.java
aaron-congo: I'll look into that

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
aaron-congo: I think that would be slower because we wouldn't get the advantage of being able to check the map for matching hosts in O(1) time and the hosts are in a list instead of a map. If we say `h` is the number of hosts, `a` is the number of aliases each host has, and `m` is the number of entries in the map, the current code executes in `O(h*a)` time, but iterating through the map would execute in `O(h*a*m)` time.

aaron-congo: Yeah not ideal maybe, let me know if you have a better solution. I'm not sure if its necessary to iterate through the host aliases, if we can get rid of that we wouldn't have nested loops. Also note that there usually should not be many hosts or many aliases, maybe ~3 hosts and ~3 aliases per host, so this should be fairly fast still.

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -372,13 +396,37 @@ private ReaderFailoverResult getNextResult(final CompletionService<ReaderFailove
     }
   }
 
-  private class ConnectionAttemptTask implements Callable<ReaderFailoverResult> {
+  protected PluginService getNewPluginService() {
+    return new PartialPluginService(
```
sergiyvamz: What's not multi-thread safe inside PartialPluginService? I see that these two new instances of PartialPluginService shares the same ConnectionPluginManager instance. Is it ok?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
```diff
@@ -134,7 +145,7 @@ public WriterFailoverResult failover(final List<HostSpec> currentTopology)
     }
   }
 
-  private HostSpec getWriter(final List<HostSpec> topology) {
+  private static HostSpec getWriter(final List<HostSpec> topology) {
```
sergiyvamz: why static?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -617,6 +631,23 @@ protected void failover(final HostSpec failedHost) throws SQLException {
     }
   }
 
+  protected ConnectionService getConnectionService() throws SQLException {
+    TargetDriverHelper helper = new TargetDriverHelper();
+    java.sql.Driver driver = helper.getTargetDriver(this.pluginService.getOriginalUrl(), properties);
+    final ConnectionProvider defaultConnectionProvider = new DriverConnectionProvider(driver);
```
sergiyvamz: Why we can't reuse a connection provider from the current PluginService?

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorServiceImpl.java
```diff
@@ -228,6 +224,25 @@ public <T extends Monitor> T runIfAbsent(
         Messages.get("MonitorServiceImpl.unexpectedMonitorClass", new Object[] {monitorClass, monitor}));
   }
 
+  protected ConnectionService getConnectionService(StorageService storageService,
+      TelemetryFactory telemetryFactory, String originalUrl, String driverProtocol, TargetDriverDialect driverDialect,
+      Dialect dbDialect, Properties originalProps) throws SQLException {
+    TargetDriverHelper helper = new TargetDriverHelper();
+    java.sql.Driver driver = helper.getTargetDriver(originalUrl, originalProps);
+    final ConnectionProvider defaultConnectionProvider = new DriverConnectionProvider(driver);
```
sergiyvamz: Why we can't reuse the same connection provider?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
aaron-congo: I now moved this method to the Utils class since we had a lot of duplicated instances of this logic

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: I've changed the code so that each plugin service gets a new plugin manager now

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -326,13 +339,29 @@ private ReaderFailoverResult getResultFromNextTaskBatch(
       final List<HostSpec> hosts,
       final ExecutorService executor,
       final CompletionService<ReaderFailoverResult> completionService,
+      final List<PluginService> pluginServices,
```
sergiyvamz: Can we pass just two parameters instead of array?

sergiyvamz (approved):

Issue #1513, opened 2025-08-20
customEndpoint plugin defaults don't work


er1c:
### Describe the bug

We added the `customEndpoint` plugin and quickly ran into:

1.  It needs an IAM permission (`rds:DescribeDBClusterEndpoints`) that isn't documented:
```
software.amazon.awssdk.services.rds.model.RdsException: User: arn:aws:sts::....:assumed-role/....-service/.... is not authorized to perform: rds:DescribeDBClusterEndpoints on resource: arn:aws:rds:us-east-1:....:cluster-endpoint:svc-....-reader because no identity-based policy allows the rds:DescribeDBClusterEndpoints action (Service: Rds, Status Code: 403, Request ID: aedfbcfc-...) 
```

2. After a few minutes it will start throwing other exceptions:

```
[ip-10-2-1-116.ec2.internal] [severe] s.a.j.p.c.CustomEndpointPlugin - Encountered an exception while monitoring custom endpoint 'svc-...-reader.cluster-custom-....us-east-1.rds.amazonaws.com'. 
[ip-10-2-1-116.ec2.internal] [error] s.a.j.p.c.CustomEndpointPlugin - Encountered an exception while monitoring custom endpoint 'svc-...-reader.cluster-custom-....us-east-1.rds.amazonaws.com'. 
software.amazon.awssdk.services.rds.model.RdsException: Rate exceeded (Service: Rds, Status Code: 400, Request ID: 3d0b5619-42bc-4039-9045-4d6bcdc32ae4) 
``` 

### Expected Behavior

I would expect the default values would work, and not cause rate limiting problems.

### What plugins are used? What other connection properties were set?

initialConnection,auroraConnectionTracker,failover2,efm2,customEndpoint

### Current Behavior

```
[ip-10-2-1-116.ec2.internal] [severe] s.a.j.p.c.CustomEndpointPlugin - Encountered an exception while monitoring custom endpoint 'svc-...-reader.cluster-custom-....us-east-1.rds.amazonaws.com'. 
[ip-10-2-1-116.ec2.internal] [error] s.a.j.p.c.CustomEndpointPlugin - Encountered an exception while monitoring custom endpoint 'svc-...-reader.cluster-custom-....us-east-1.rds.amazonaws.com'. 
[...]

PR #1521, opened 2025-09-02
bug: handle unauthorized exceptions while getting custom endpoint details

sergiyvamz:
### Summary

handle unauthorized exceptions while getting custom endpoint details
https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1513

### Description

- handle RdsException associated with user lack of permissions; use some delay before trying again
- use reduced refresh rate (5min) when IAM user has no permissions to get custom endpoint details
- handle throttle exception and reduce refresh rate

### Additional Reviewers

@karenc-bq 

### DIFF
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheCustomEndpointPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheCustomEndpointPlugin.md
@@ -26,3 +26,11 @@ The Custom Endpoint Plugin adds support for RDS custom endpoints. When the Custo
 | `customEndpointMonitorExpirationMs`  | Integer |    No    | Controls how long a monitor should run without use before expiring and being removed, in milliseconds.                                                                                                                                                                                                                               | `900000` (15 minutes) | `600000`      |
 | `waitForCustomEndpointInfo`          | Boolean |    No    | Controls whether to wait for custom endpoint info to become available before connecting or executing a method. Waiting is only necessary if a connection to a given custom endpoint has not been opened or used recently. Note that disabling this may result in occasional connections to instances outside of the custom endpoint. | `true`                | `true`        |
 | `waitForCustomEndpointInfoTimeoutMs` | Integer |    No    | Controls the maximum amount of time that the plugin will wait for custom endpoint info to be made available by the custom endpoint monitor, in milliseconds.                                                                                                                                                                         | `5000`                | `7000`        |
+
+### Use IAM authentication with the Custom Endpoint Plugin
+
+When using IAM authentication make sure that IAM user has `rds:DescribeDBClusterEndpoints` permission granted. You may see a corresponding exception in the driver logs if IAM user doesn't have this permission:
+
+```
+software.amazon.awssdk.services.rds.model.RdsException: User: arn:aws:sts:...:assumed-role/.... is not authorized to perform: rds:DescribeDBClusterEndpoints on resource: arn:aws:rds:.... because no identity-based policy allows the rds:DescribeDBClusterEndpoints action (Service: Rds, Status Code: 403, Request ID: ...) 
+```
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -168,12 +171,30 @@ public void monitor() {
...
         } catch (InterruptedException e) {
           throw e;
+        } catch (RdsException ex) {
+          LOGGER.log(Level.SEVERE,
+              Messages.get(
+                  "CustomEndpointMonitorImpl.exception",
+                  new Object[] {this.customEndpointHostSpec.getUrl()}), ex);
+
+          if (ex.isThrottlingException()) {
+            this.refreshRateNano *= 2; // Reduce the refresh rate.
```
karenc-bq, line 181: Will this number increase exponentially if the requests keep failing, will it cause the application to "hang"?

docs/using-the-jdbc-driver/using-plugins/UsingTheCustomEndpointPlugin.md
```diff
@@ -26,3 +26,11 @@ The Custom Endpoint Plugin adds support for RDS custom endpoints. When the Custo
...
 | `waitForCustomEndpointInfo`          | Boolean |    No    | Controls whether to wait for custom endpoint info to become available before connecting or executing a method. Waiting is only necessary if a connection to a given custom endpoint has not been opened or used recently. Note that disabling this may result in occasional connections to instances outside of the custom endpoint. | `true`                | `true`        |
 | `waitForCustomEndpointInfoTimeoutMs` | Integer |    No    | Controls the maximum amount of time that the plugin will wait for custom endpoint info to be made available by the custom endpoint monitor, in milliseconds.                                                                                                                                                                         | `5000`                | `7000`        |
+
+### Use IAM authentication with the Custom Endpoint Plugin
+
+When using IAM authentication make sure that IAM user has `rds:DescribeDBClusterEndpoints` permission granted. You may see a corresponding exception in the driver logs if IAM user doesn't have this permission:
```
karenc-bq, line 32: Just as a sanity check, have we confirmed this is the only permission required?

sergiyvamz: Yes, that's the only one we need.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
sergiyvamz: No, it slows down a custom endpoint monitoring rate. Since such monitoring occurs in a separate thread there's no risk of "hanging" of the user application.

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -168,12 +171,30 @@ public void monitor() {
           TimeUnit.NANOSECONDS.sleep(sleepDuration);
         } catch (InterruptedException e) {
           throw e;
+        } catch (RdsException ex) {
+          LOGGER.log(Level.SEVERE,
```
er1c, line 175: should this still be SEVERE if it's a throttling exception? especially given the full stack trace

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
er1c: This potentially always increases the refresh interval and never decreases it?

For a long-running process this just keeps going up and up; even if there are other processes that are causing a short-term rate throttling error.

Doing any type of rolling-restart of a service with many processes that all use the same custom endpoint will inevitably always start hitting the rate limiting and see these logger severe exceptions

PR #1523, opened 2025-09-04
fix: writer host comparison in AuroraConnectionTrackerPlugin

aaron-congo:
- before these changes, the AuroraConnectionTrackerPlugin would use the HostSpec's equals method to determine whether the writer had changed. This was causing false positives because the weight of each HostSpec is compared in the equals method, and the weight of a given host before/after failover may be different even if the host itself is the same. To fix the issue we compare the URLs of both hosts to determine if the writer has changed
- also updated and refactored other areas of the code to compare host URLs instead of using HostSpec#equals to avoid this same issue


wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraConnectionTrackerPlugin.java
```diff
@@ -156,13 +156,16 @@ private void checkWriterChanged(boolean needRefreshHostLists) {
...
+    if (hostSpecAfterFailover == null) {
+      return;
+    }
 
     if (this.currentWriter == null) {
       this.currentWriter = hostSpecAfterFailover;
       this.needUpdateCurrentWriter = false;
-
-    } else if (!this.currentWriter.equals(hostSpecAfterFailover)) {
+    } else if (!this.currentWriter.getUrl().equals(hostSpecAfterFailover.getUrl())) {
```
sergiyvamz: getHostAndPort() ?

sergiyvamz (approved):

PR #1531, opened 2025-09-12
Remote query cache plugin

QuChen88:
### Summary

Remote caching plugin implementation. The goal of the plugin is to cache query result in a remote cache server for certain read-only queries that are deemed to be cacheable. All incoming queries that contain a query hint with caching parameter settings specified will have their responses cached with the defined TTL. Further requests with the same query get their result served from the cache instead of from the backend database. The remote cache server runs on Valkey / Redis and is assumed to be able to scale and handle eviction of data when the cache is full, so the size limitation is no longer an issue on the client side.

### Basic walkthrough of code:

- The entry point for the code is `DataRemoteCachePlugin.java`. It contains the core logic for the remote caching on both the query read path and write path. It will call into `CacheConnection.java` to handle reading and writing to the remote cache server.
- `CacheConnection.java` mainly handles interaction between the client and the cache server via a connection pool. On the write path, the query result gets serialized into raw bytes which gets stored in the cache server. On the read path, the query result gets fetched from the cache server, de-serialized into Java objects and returned back to the driver application for consumption.
- The logic to handle cached query result is `CachedResultSet.java`. `getXXX` APIs there for supported for various data types (e.g. String, boolean, int, double, float, Date, Timestamp, etc) along with serialization and de-serialization logic for `CachedResultSet`.
- Supports plain username/password authentication to the cache server
- Supports IAM authentication for cache connections to AWS Elasticache
- Caching failure handling
- The existing `DataCacheConnectionPlugin.java` has been re-factored into `DataLocalCacheConnectionPluginFactory.java` to have a more descriptive name indicating it is storing cached results locally. 
- Unit testing and integration testing for the new logic is included
- Example program is added to indicate how to use the new `DataRemoteCachePlugin`. Specifically a few additional properties are required for the plugin to work. 
- Implement cache connection via [ValkeyGlide](https://github.com/valkey-io/valkey-glide) client


examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
```diff
@@ -0,0 +1,67 @@
+package software.amazon;
+
+import software.amazon.util.EnvLoader;
+import java.sql.*;
+import java.util.*;
+import java.util.logging.Logger;
```
alexey-amazon, line 16: java.util.logging is extremely ancient thing and no new apps should be using it.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,266 @@
...
+import io.lettuce.core.codec.ByteArrayCodec;
+import io.lettuce.core.resource.ClientResources;
+import software.amazon.jdbc.AwsWrapperProperty;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
+import java.util.Properties;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.logging.Logger;
```
alexey-amazon, line 48: We should find ways to use something else here, perhaps SLF4J?

alexey-amazon, line 48: Please consult with Dave what and how we can log in the driver if anything, but it seems JUL is not what we should be using there?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,266 @@
...
+  With isRead we check if we need to initialise connection pool for read or write to cache.
+  If isRead is true, we initialise connection pool for read.
+  If isRead is false, we initialise connection pool for write.
+   */
+  private void initializeCacheConnectionIfNeeded(boolean isRead) {
+    if (StringUtils.isNullOrEmpty(cacheRwServerAddr)) return;
+    // Initialize the message digest
+    if (msgHashDigest == null) {
+      try {
+        msgHashDigest = MessageDigest.getInstance("SHA-384");
```
alexey-amazon: So this is a crypto use-case where we really want it to be super low probability of collisions?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,266 @@
...
+      // If read-only server is specified, use it for the read-only connections
+      if (isRead && !StringUtils.isNullOrEmpty(cacheRoServerAddr)) {
+        serverAddr = cacheRoServerAddr;
+      }
+      String[] hostnameAndPort = serverAddr.split(":");
+      RedisURI redisUriCluster = RedisURI.Builder.redis(hostnameAndPort[0])
+          .withPort(Integer.parseInt(hostnameAndPort[1]))
+          .withSsl(useSSL).withVerifyPeer(false).withLibraryName("aws-jdbc-lettuce").build();
+
+      RedisClient client = RedisClient.create(resources, redisUriCluster);
```
alexey-amazon: Do we just compile with arbitrary version and let customer provide actual Lettuce client of version they so choose hoping that it's good enough?

wrapper/build.gradle.kts
```diff
@@ -48,9 +48,11 @@ dependencies {
     compileOnly("org.osgi:org.osgi.core:6.0.0")
     compileOnly("org.osgi:org.osgi.core:6.0.0")
     compileOnly("com.amazonaws:aws-xray-recorder-sdk-core:2.18.2")
+    compileOnly("io.lettuce:lettuce-core:6.6.0.RELEASE")
     compileOnly("io.opentelemetry:opentelemetry-api:1.52.0")
     compileOnly("io.opentelemetry:opentelemetry-sdk:1.52.0")
     compileOnly("io.opentelemetry:opentelemetry-sdk-metrics:1.52.0")
+    compileOnly("org.apache.commons:commons-pool2:2.11.1")
```
alexey-amazon: Will this break customers who happen to have bad version of commons-pool2 in their classpath only if they activate this caching plugin?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,266 @@
...
+            public StatefulRedisConnection<byte[], byte[]> create() {
+              StatefulRedisConnection<byte[], byte[]> connection = client.connect(new ByteArrayCodec());
+              // In cluster mode, we need to send READONLY command to the server for reading from replica.
+              // Note: we gracefully ignore ERR reply to support non cluster mode.
+              if (isRead) {
+                try {
+                  connection.sync().readOnly();
+                } catch (RedisCommandExecutionException e) {
+                  if (e.getMessage().contains("ERR This instance has cluster support disabled")) {
+                    LOGGER.fine("------ Note: this cache cluster has cluster support disabled ------");
```
alexey-amazon: Where this is getting written? Why this is better than using SLf4J?

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
```diff
@@ -0,0 +1,67 @@
...
+    properties.setProperty("user", USERNAME);
+    properties.setProperty("password", PASSWORD);
+
+    // Configuring connection properties for the JDBC Wrapper.
+    properties.setProperty("wrapperPlugins", "dataRemoteCache");
+    properties.setProperty("cacheEndpointAddrRw", CACHE_RW_SERVER_ADDR);
+    properties.setProperty("cacheEndpointAddrRo", CACHE_RO_SERVER_ADDR);
+    properties.setProperty("cacheUseSSL", USE_SSL); // "true" or "false"
+    properties.setProperty("wrapperLogUnclosedConnections", "true");
+    String queryStr = "/*+ CACHE_PARAM(ttl=300s) */ select * from cinemas";
```
alexey-amazon, line 83: Is there any TTL configuration we can set for the entire driver?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,335 @@
...
+import software.amazon.jdbc.util.StringUtils;
+import software.amazon.jdbc.util.WrapperUtils;
+import software.amazon.jdbc.util.telemetry.TelemetryCounter;
+import software.amazon.jdbc.util.telemetry.TelemetryFactory;
+import software.amazon.jdbc.util.telemetry.TelemetryContext;
+import software.amazon.jdbc.util.telemetry.TelemetryTraceLevel;
+
+public class DataRemoteCachePlugin extends AbstractConnectionPlugin {
+  private static final Logger LOGGER = Logger.getLogger(DataRemoteCachePlugin.class.getName());
+  private static final int MAX_CACHEABLE_QUERY_SIZE = 16000;
```
alexey-amazon: Usually people use 16384 instead. Otherwise I feel I am being short-changed ;)

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -0,0 +1,335 @@
...
+
+    TelemetryContext cacheContext = null;
+    TelemetryContext dbContext = null;
+    // If the query is cacheable, we try to fetch the query result from the cache.
+    boolean isInTransaction = pluginService.isInTransaction();
+    // Get the query hint part in front of the query itself
+    String mainQuery = sql; // The main part of the query with the query hint prefix trimmed
+    int endOfQueryHint = 0;
+    Integer configuredQueryTtl = null;
+    // Queries longer than 16KB is not cacheable
```
alexey-amazon, line 339: > // Queries longer than 16KB is not cacheable
 
 This is configurable, right? Why this comment has specific value?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,335 @@
...
+    boolean isInTransaction = pluginService.isInTransaction();
+    // Get the query hint part in front of the query itself
+    String mainQuery = sql; // The main part of the query with the query hint prefix trimmed
+    int endOfQueryHint = 0;
+    Integer configuredQueryTtl = null;
+    // Queries longer than 16KB is not cacheable
+    if ((sql.length() < MAX_CACHEABLE_QUERY_SIZE) && sql.startsWith(QUERY_HINT_START_PATTERN)) {
+      endOfQueryHint = sql.indexOf(QUERY_HINT_END_PATTERN);
+      if (endOfQueryHint > 0) {
+        configuredQueryTtl = getTtlForQuery(sql.substring(2, endOfQueryHint).trim());
```
alexey-amazon: Should we use QUERY_HINT_START_PATTERN.length() instead of 2?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,335 @@
...
+        mainQuery = sql.substring(endOfQueryHint + 2).trim();
```
alexey-amazon: here too

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -0,0 +1,335 @@
...
+            return null;
+          }
+        }
+      }
+    }
+    return ttlValue;
+  }
+
+  @Override
+  public <T, E extends Exception> T execute(
```
alexey-amazon, line 313: My reading of this code let me conclude that we ask customer to specify this hint with TTL, which will actually enable caching.

Some folks might find it difficult to use, because you have to have these TTLs though out the app in many SQL queries. However, even worse part of it is that it cannot be disabled, because if caching plugin is turned off, these hints will go through to the database and error out. 

There might be an alternative way to do it such as configuring TTL for the connection or the entire pool of connections without amending every query with a hint. In such case all customer will need to do is to use different pool for their cachable use-case.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: So far this is a hard-coded 16KB value and not configurable. We have been using 16KB as the limit for long command inputs in both elasticache and RDS proxy and so far it worked out well. 

The more knobs we expose to the user, the more likely they get confused.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Good idea. Will fix.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: I executed the query with the cache hints against Postgres. The query was able to get executed without any problems. The database will simply treat it just like any other query comments and ignore the stuff it doesn't recognize.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: will use 16384.

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
QuChen88: Good question. 

We considered allowing the user to set a global TTL configuration for all the queries. However that has not been very well thought out in terms of what happens when the global TTL conflicts with the query hint for a particular SQL? Also we believe not all SQLs are good to be cached and the user needs to opt into this feature.

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
QuChen88: I have been simply following the current convention in the driver code and using `java.util.logging.Logger`. 

What alternative should I be using? Can you point me to an example code in this driver?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Correct. AWS Crypto BRs recommend SHA-384 as hash function which is collision resistant.

wrapper/build.gradle.kts
QuChen88: Based on my testing, yes. 

If I don't enable the cache plugin, then Lettuce and the pool dependency won't get invoked.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Will talk to Dave about what is the recommended way to log in the driver. Right now lots of places are using JUL.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: No, the driver is specifically running on lettuce version 6.6.0 so the client APIs being used here are compatible with that.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Not sure I understand.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
alexey-amazon: ok, cool, good to know!

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
alexey-amazon: I think we should have it configurable with this 16K default. Imagine you have a query which is very close to 16K and then after adding just a couple of characters it cannot be cached any more and there is no way to configure any different limits. That would be very difficult.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
alexey-amazon: these are optional dependencies. What will happen if you add this wrapper into a project which already depends on Lettuce? Does it have to be the same version as driver wants?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
alexey-amazon: I think the current story is to rely on JUL. Ok then.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
alexey-amazon: Can we also please check with MySQL?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: If that happens then the user needs to resolve the version dependency conflict via gradle using the higher version of the 2.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: Made it configurable.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: Thanks for the reminder. I set up MySQL locally and tested my program against it.

- In terms of query hints, MySQL was able to ignore query hints that is not recognized similar to Postgres (e.g. caching hints), so no concerns there.
- However, I noticed that `Connection.getCatalog()` and `Connection.getSchema()` behave a bit differently from Postgres, also the database username is actually `<username>@<ip>` instead of just the plain username as in postgres. Those required a small bit of functional tweak in the caching logic. Good to catch it!

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+                "DataLocalCacheConnectionPlugin.queryResultsCached",
+                new Object[]{methodName, sql}));
+      } else {
+        if (this.hitCounter != null) {
+          this.hitCounter.inc();
+        }
+        try {
+          result.beforeFirst();
+        } catch (final SQLException ex) {
+          if (exceptionClass.isAssignableFrom(ex.getClass())) {
```
sergiyvamz: You can use WrapperUtils.wrapExceptionIfNeeded()

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+  public DataRemoteCachePlugin(final PluginService pluginService, final Properties properties) {
+    try {
+      Class.forName("io.lettuce.core.RedisClient"); // Lettuce dependency
+      Class.forName("org.apache.commons.pool2.impl.GenericObjectPool"); // Object pool dependency
+    } catch (final ClassNotFoundException e) {
+      throw new RuntimeException(Messages.get("DataRemoteCachePlugin.notInClassPath", new Object[] {e.getMessage()}));
+    }
+    this.pluginService = pluginService;
+    this.telemetryFactory = pluginService.getTelemetryFactory();
+    this.cacheHitCounter = telemetryFactory.createCounter("JdbcCachedQueryCount");
```
sergiyvamz: Other counters of the driver uses notation with dot, like "dataCache.cache.miss". Can we change those names accordingly?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+    try {
+      Connection currentConn = pluginService.getCurrentConnection();
+      DatabaseMetaData metadata = currentConn.getMetaData();
+      // Fetch and record the schema name if the session state doesn't currently have it
+      SessionStateService sessionStateService = pluginService.getSessionStateService();
+      String catalog = sessionStateService.getCatalog().orElse(null);
+      String schema = sessionStateService.getSchema().orElse(null);
+      if (catalog == null && schema == null) {
+        // Fetch the current schema name and store it in sessionStateService
+        catalog = currentConn.getCatalog();
```
sergiyvamz: Returned catalog and schema could be null. If you set nulls to session state service then then time you call this method you get nulls from session state service and do this logic again. I'd suggest to pay attention on the following statements and use isPresent() instead of orElse(null).

```
String catalog = sessionStateService.getCatalog().orElse(null);
String schema = sessionStateService.getSchema().orElse(null);

```

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+        schema = currentConn.getSchema();
+        if (catalog != null) sessionStateService.setCatalog(catalog);
+        if (schema != null) sessionStateService.setSchema(schema);
+      }
+
+      if (dbUserName == null) {
+        // For MySQL, metadata username is actually <UserName>@<ip>. We just need the part before '@'.
+        dbUserName = metadata.getUserName();
```
sergiyvamz: Why not to use username from connection properties?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+        int nameIndexEnd = dbUserName.indexOf('@');
+        if (nameIndexEnd > 0) {
+          dbUserName = dbUserName.substring(0, nameIndexEnd);
+        }
+      }
+      LOGGER.finest("DB driver protocol " + pluginService.getDriverProtocol()
```
sergiyvamz: All log statements should use the following pattern (use lambda). As an example:

```
LOGGER.finest(() -> "DB driver protocol " + pluginService.getDriverProtocol()
+ ", database product: " + metadata.getDatabaseProductName() + " " + metadata.getDatabaseProductVersion()
          + ", catalog: " + catalog + ", schema: " + schema + ", user: " + dbUserName
          + ", driver: " + metadata.getDriverName() + " " + metadata.getDriverVersion());

```

Also for such string concatenations is better use String.format() or StringBuilder.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+      String[] words = {catalog, schema, dbUserName, query};
+      return String.join("_", words);
+    } catch (SQLException e) {
+      LOGGER.warning("Error getting session state: " + e.getMessage());
+      return null;
+    }
+  }
+
+  private ResultSet fetchResultSetFromCache(String queryStr) throws SQLException {
+    if (cacheConnection == null) return null;
```
sergiyvamz: Why not to throw an exception? It seems it's abnormal case when connection object is null. Or, should we reconnect in this case?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+
+    String cacheQueryKey = getCacheQueryKey(queryStr);
+    if (cacheQueryKey == null) return null; // Treat this as a cache miss
+    byte[] cachedResult = cacheConnection.readFromCache(cacheQueryKey);
+    if (cachedResult == null) return null;
+    // Convert result into ResultSet
+    try {
+      return CachedResultSet.deserializeFromByteArray(cachedResult);
+    } catch (Exception e) {
+      LOGGER.warning("Error de-serializing cached result: " + e.getMessage());
```
sergiyvamz: I'd suggest to log original exception instead of just its message. It's valuable for troubleshooting.

`LOGGER.log(Level.WARNING, "Error de-serializing cached result: " + e.getMessage(), e);`

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+   *  The ResultSet object passed in would be consumed to create a CacheResultSet object. It is returned
+   *  for consumer consumption.
+   */
+  private ResultSet cacheResultSet(String queryStr, ResultSet rs, int expiry) throws SQLException {
+    // Write the resultSet into the cache as a single key
+    String cacheQueryKey = getCacheQueryKey(queryStr);
+    if (cacheQueryKey == null) return rs; // Treat this condition as un-cacheable
+    CachedResultSet crs = new CachedResultSet(rs);
+    byte[] jsonString = crs.serializeIntoByteArray();
+    cacheConnection.writeToCache(cacheQueryKey, jsonString, expiry);
```
sergiyvamz: Possible NPE

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+
+  /**
+   * Determine the TTL based on an input query
+   * @param queryHint string. e.g. "CACHE_PARAM(ttl=100s, key=custom)"
+   * @return TTL in seconds to cache the query.
+   *         null if the query is not cacheable.
+   */
+  protected Integer getTtlForQuery(String queryHint) {
+    // Empty query is not cacheable
+    if (StringUtils.isNullOrEmpty(queryHint)) return null;
```
sergiyvamz: For redability and to follow our codestyle please use the following pattern:

```
if (condition) {
   return;
}

```

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+  protected Integer getTtlForQuery(String queryHint) {
+    // Empty query is not cacheable
+    if (StringUtils.isNullOrEmpty(queryHint)) return null;
+    // Find CACHE_PARAM anywhere in the hint string (case insensitive)
+    String upperHint = queryHint.toUpperCase();
+    int cacheParamStart = upperHint.indexOf(CACHE_PARAM_PATTERN);
+    if (cacheParamStart == -1) return null;
+
+    // Find the matching closing parenthesis
+    int paramsStart = cacheParamStart + CACHE_PARAM_PATTERN.length();
```
sergiyvamz, line 250: Using regular expression is preferable.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+
+    if (resultClass != ResultSet.class) {
+      return jdbcMethodFunc.call();
+    }
+
+    incrCounter(totalQueryCounter);
+
+    ResultSet result;
+    boolean needToCache = false;
+    final String sql = getQuery(methodInvokeOn, jdbcMethodArgs);
```
sergiyvamz: It's recommended to use `this` when calling method or accessing member variables.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+          } else { // Cache miss - leave context open
+            cacheContext.setSuccess(false);
+          }
+        }
+      }
+    } else {
+      incrCounter(cacheBypassCounter);
+    }
+
+    dbContext = telemetryFactory.openTelemetryContext(
```
sergiyvamz, line 393: Why TOP_LEVEL?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
```diff
@@ -0,0 +1,382 @@
...
+      return jdbcMethodArgs[0].toString().trim();
+    }
+
+    // If the query is not in the method arguments, check for prepared statement query. Get the query
+    // string from the prepared statement. The exact query string is dependent on the underlying driver.
+    if (methodInvokeOn instanceof PreparedStatement) {
+      // For postgres, this gives the raw query itself. i.e. "select * from T where A = 1".
+      // For MySQL, this gives "com.mysql.cj.jdbc.ClientPreparedStatement: select * from T where A = 1"
+      // For MariaDB, this gives "ClientPreparedStatement{sql:'select * from T where A=1', parameters:[]}"
+      return methodInvokeOn.toString();
```
sergiyvamz: You should move this logic to TargetDriverDialect.

What is the logic if it return null instead of query, if it return something else like class name?

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/PgCacheBenchmarks.java
```diff
@@ -0,0 +1,144 @@
...
+    for (int i = 0; i < 400000; i++) {
+      Statement stmt = connection.createStatement();
+      String description = "description " + i;
+      String text = "here is my text data " + i;
+      String query = "insert into test values (" + i + ", " + i * 10 + ", '" + description + "', '" + text + "', " + i * 100 + 0.1234 + ", '2024-01-10', '10:00:00', '10:00:00-07', '2025-07-15 10:00:00', '2025-07-15 10:00:00-07'" + ", '" + desc_1KB + "');";
```
sergiyvamz: String.format()

examples/AWSDriverExample/src/main/java/software/amazon/util/EnvLoader.java
```diff
@@ -0,0 +1,83 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A simple utility class to load environment variables from a .env file.
+ */
+public class EnvLoader {
```
sergiyvamz: Is it for example code? Does it make sense to use existing loader like this

https://github.com/cdimascio/dotenv-java

?

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
```diff
@@ -0,0 +1,186 @@
+package software.amazon;
+
+import software.amazon.util.EnvLoader;
+import java.sql.*;
+import java.util.*;
+import java.util.logging.Logger;
+
+public class DatabaseConnectionWithCacheExample {
```
sergiyvamz, line 18: Did you run this code? Just curious about performance numbers.

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
```diff
@@ -0,0 +1,186 @@
...
+            }
+          }
+        } catch (Exception e) {
+          LOGGER.warning("Error: " + e.getMessage());
+        }
+      });
+      threads[t].start();
+    }
+    // Wait for all threads to complete
+    for (Thread thread : threads) {
```
sergiyvamz: Modern Java recommends to use Executors to start threads.

examples/AWSDriverExample/src/main/resources/logback.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sergiyvamz: Why do you need this file? I think logger is already configured with `logging-example.properties` file.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+          JdbcMethod.STATEMENT_EXECUTEQUERY.methodName,
+          JdbcMethod.STATEMENT_EXECUTE.methodName,
+          JdbcMethod.PREPAREDSTATEMENT_EXECUTE.methodName,
+          JdbcMethod.PREPAREDSTATEMENT_EXECUTEQUERY.methodName,
+          JdbcMethod.CALLABLESTATEMENT_EXECUTE.methodName,
+          JdbcMethod.CALLABLESTATEMENT_EXECUTEQUERY.methodName
+      )));
+
+  public static final AwsWrapperProperty DATA_CACHE_TRIGGER_CONDITION = new AwsWrapperProperty(
+      "dataCacheTriggerCondition", "false",
```
sergiyvamz: Default value is `false`? That doesn't correspond to parameter description.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+
+  protected static final AwsWrapperProperty CACHE_NAME =
+      new AwsWrapperProperty(
+          "cacheName",
+          null,
+          "Explicit cache name for ElastiCache IAM authentication. ");
+
+  protected static final AwsWrapperProperty CACHE_CONNECTION_TIMEOUT =
+      new AwsWrapperProperty(
+          "cacheConnectionTimeout",
```
sergiyvamz: It's better to give user a hint what time unit to use, seconds, ms, etc. As a suggestion:

`cacheConnectionTimeout` -> `cacheConnectionTimeoutMs`

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+          "2000",
+          "Cache connection request timeout duration in milliseconds.");
+
+  protected static final AwsWrapperProperty CACHE_CONNECTION_POOL_SIZE =
+      new AwsWrapperProperty(
+          "cacheConnectionPoolSize",
```
sergiyvamz, line 133: Is it initial size or max size?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+      new AwsWrapperProperty(
+          "cacheKeyPrefix",
+          null,
+          "Optional prefix for cache keys (max 10 characters). Enables multi-tenant cache isolation.");
+
+  // Adding support for read and write connection pools to the remote cache server
+  private volatile GenericObjectPool<StatefulConnection<byte[], byte[]>> readConnectionPool;
+  private volatile GenericObjectPool<StatefulConnection<byte[], byte[]>> writeConnectionPool;
+  // Cache endpoint registry to hold connection pools for multi end points
+  private static final ConcurrentHashMap<String, GenericObjectPool<StatefulConnection<byte[], byte[]>>> endpointToPoolRegistry = new ConcurrentHashMap<>();
```
sergiyvamz: Static variable should go before instance members.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+
+  static {
+    PropertyDefinition.registerPluginProperties(CacheConnection.class);
+  }
+
+  /**
+   * Wraps a StatefulConnection (either StatefulRedisConnection or StatefulRedisClusterConnection)
+   * and exposes only ping functionality.
+   */
+  private static class PingConnection implements CachePingConnection {
```
sergiyvamz: Usually nested class definition should be at the bottom of main class.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+    this.useSSL = Boolean.parseBoolean(CACHE_USE_SSL.getString(properties));
+    this.cacheName = CACHE_NAME.getString(properties);
+    this.cacheIamRegion = CACHE_IAM_REGION.getString(properties);
+    this.cacheUsername = CACHE_USERNAME.getString(properties);
+    this.cachePassword = CACHE_PASSWORD.getString(properties);
+    this.cacheConnectionTimeout = Duration.ofMillis(CACHE_CONNECTION_TIMEOUT.getInteger(properties));
+    this.cacheConnectionPoolSize = CACHE_CONNECTION_POOL_SIZE.getInteger(properties);
+    if (this.cacheConnectionPoolSize <= 0 || this.cacheConnectionPoolSize > DEFAULT_MAX_POOL_SIZE) {
+      throw new IllegalArgumentException(
+          "Cache connection pool size must be within valid range: 1-" + DEFAULT_MAX_POOL_SIZE + ", but was: " + this.cacheConnectionPoolSize);
```
sergiyvamz: Message should be in resource file. Use `Messages.get()` to get a message from resource.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+    }
+    this.failWhenCacheDown = FAIL_WHEN_CACHE_DOWN.getBoolean(properties);
+    this.cacheKeyPrefix = CACHE_KEY_PREFIX.getString(properties);
+    if (this.cacheKeyPrefix != null) {
+      if (this.cacheKeyPrefix.isEmpty() || this.cacheKeyPrefix.trim().isEmpty()) {
```
sergiyvamz: The first condition is covered by the second one.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+      );
+    } else {
+      this.awsProfileProperties = null;
+    }
+    if (this.iamAuthEnabled) {
+      // Handle null case
+      Properties propsToPass = (this.awsProfileProperties != null)
+          ? this.awsProfileProperties
+          : new Properties();
+      this.credentialsProvider = AwsCredentialsManager.getProvider(null, propsToPass);
```
sergiyvamz: You can pass original `properties` to getProvider().

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+    ClientResources resources = ClientResources.builder().build();
+
+    try (RedisClient client = RedisClient.create(resources, redisUri);
+         StatefulRedisConnection<byte[], byte[]> conn = client.connect(new ByteArrayCodec())) {
+
+      String infoOutput = conn.sync().info("cluster");
+      boolean clusterEnabled = false;
+      if (infoOutput != null) {
+        // Parse the INFO output line by line
+        for (String line : infoOutput.split("\r?\n")) {
```
sergiyvamz, line 274: Regular expression maybe?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+        throw new RuntimeException("SHA-384 not supported", e);
+      }
+    }
+
+    // return early if connection pool is already initialized before acquiring the lock
+    if ((isRead && this.readConnectionPool != null) || (!isRead && this.writeConnectionPool != null)) {
+      return;
+    }
+
+    connectionInitializationLock.lock();
```
sergiyvamz: You can use ResourceLock

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,832 @@
...
+    try {
+      // Double check after lock is acquired
+      if ((isRead && this.readConnectionPool != null) || (!isRead && this.writeConnectionPool != null)) {
+        return;
+      }
+      // Detect cluster mode first (cached for reuse)
+      detectClusterMode();
+
+      // Register cluster with CacheMonitor on first cache operation (skip if telemetryFactory is null = test mode)
+      if (telemetryFactory != null && !this.cacheMonitorRegistered) {
```
sergiyvamz: What if telemetryFactory is null since user disabled telemetry?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
```diff
@@ -0,0 +1,1441 @@
...
+
+  private Time convertToTime(Object timeObj, Calendar cal) throws SQLException {
+    if (timeObj == null) return null;
+    if (timeObj instanceof Time) return (Time) timeObj;
+    if (timeObj instanceof Number) return new Time(((Number)timeObj).longValue()); // TODO: test
+    if (timeObj instanceof LocalTime) {
+      // Convert the LocalTime for the specified time zone into Time representing
+      // the same instant of time for the default time zone.
+      LocalTime localTime = (LocalTime)timeObj;
+      if (cal == null) return Time.valueOf(localTime);
```
sergiyvamz: I'm not sure such logic is necessary here. Is to support scenario when one application stores data to the remote cache with one time zone settings and the other app gets data with different time zone?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedSupplier.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package software.amazon.jdbc.plugin.cache;
+
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+
+public final class CachedSupplier {
```
sergiyvamz: Is it similar object?
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/utils/cache/CachedSupplier.html

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheMonitor.java
```diff
@@ -0,0 +1,527 @@
...
+    if (cluster == null) {
+      LOGGER.warning("Error report for unregistered cluster: RW=" + rwEndpoint + ", RO=" + roEndpoint);
+      return;
+    }
+    ErrorCategory category = classifyError(error);
+    if (errorCounter != null) {
+      errorCounter.inc();
+    }
+    if (!isRecoverableError(category)) {
+      LOGGER.info(() -> "Non-recoverable error (" + category + ") for " +
```
sergiyvamz: Should it be SEVERE instead of info?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheMonitor.java
```diff
@@ -0,0 +1,527 @@
...
+import software.amazon.jdbc.util.telemetry.TelemetryFactory;
+import software.amazon.jdbc.util.telemetry.TelemetryGauge;
+
+/**
+ * This class uses a background thread to monitor cache cluster health and manage state transitions.
+ *
+ * Implements a three-state machine (HEALTHY → SUSPECT → DEGRADED) with proactive health checks
+ * that only run when clusters are in SUSPECT or DEGRADED states.
+ */
+public class CacheMonitor implements Runnable {
```
sergiyvamz: You need to extend this class from software.amazon.jdbc.util.monitoring.AbstractMonitor and use MonitorService to start an instance.

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/ElastiCacheIamTokenUtility.java
```diff
@@ -0,0 +1,125 @@
...
+
+  // For testing only
+  public ElastiCacheIamTokenUtility(String cacheName, Instant fixedInstant, Aws4Signer signer) {
+    this.cacheName = Objects.requireNonNull(cacheName, "cacheName cannot be null");
+    this.clock = Clock.fixed(fixedInstant, ZoneId.of("UTC"));
+    this.signer = signer;
+  }
+
+  @Override
+  public String generateAuthenticationToken(
```
sergiyvamz: Host this token is different from IAM token?

wrapper/src/test/java/integration/TestEnvironmentInfo.java
```diff
@@ -47,6 +48,10 @@ public TestDatabaseInfo getDatabaseInfo() {
     return this.databaseInfo;
   }
 
+  public TestDatabaseInfo getDbCacheInfo() {
```
sergiyvamz: Can we rename it to something more obvious like `getValkeyServerInfo`?

sergiyvamz: Do you actually need all those fields in TestDatabaseInfo? Does it make sense to create a small class with valkey-related properties? It seems that you need just a server name (network name) and port.

.env.example
```diff
@@ -0,0 +1,5 @@
+DB_CONNECTION_STRING=jdbc:aws-wrapper:postgresql://localhost:5432/dbname
```
sergiyvamz: Is it a part of some code example? Should we move it to appropriate folder?

sergiyvamz:
Can we have documentation for the new plugin?

examples/AWSDriverExample/src/main/java/software/amazon/util/EnvLoader.java
QuChen88: Yes this is for the example code. Will use an existing loader for the .env configuration file.

benchmarks/src/jmh/java/software/amazon/jdbc/benchmarks/PgCacheBenchmarks.java
QuChen88: Will do

examples/AWSDriverExample/src/main/java/software/amazon/DatabaseConnectionWithCacheExample.java
QuChen88: This is example program and not benchmark code. 

`PgCacheBenchmark.java` is for benchmarking. I will put performance numbers on the documentation PR that follows.

examples/AWSDriverExample/src/main/resources/logback.xml
QuChen88: If I remember, this was mainly for me to suppress the debug logs from the Lettuce client when I ran the example program as it was clogging the console output which was making debugging difficult. Couldn't find any other way to suppress the logs from Lettuce.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Will change this

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: The max size.

QuChen88: Actually both initial and max size. It doesn't currently change dynamically

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Fixed

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Moved

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Removed the first condition.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: The original set of properties contain a lot of other information that don't need to get passed and stored in the credentials provider. Want to be a bit efficient here.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: I think regular expression has overhead compared to simpler `String.split()` and is a lot more error prone as well. The parsing here is relatively simple so there is no need for a more complex regex

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Will look at moving the exception messages into [wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties]

> Message should be in resource file. Use `Messages.get()` to get a message from resource.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
QuChen88: Fixed

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Fixed. Will use `isPresent()` to determine if I need to fetch them from `currentConn` or not. 

Sometimes I noticed that one of them can be `null` for PG/MySQL, which is normal. This logic should handle that (i.e. indicate `null` in that case).

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: I discussed this with Dave previously. The original idea was that if there is something wrong with the cache (e.g. Valkey cache server is not present or the cache connection is not initialized), then we fall back to the database instead of bubbling up the exception and causing the query to fail. 

Looking at the implementation, in the constructor of `DataRemoteCachePlugin()`, I always initialize `cacheConnection` field, so this should not happen. I can remove this check and cause an exception to be thrown if it is `null` somehow.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Fixed

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Nice catch. I am handling NPE now and wrapping it inside SQLException.

examples/AWSDriverExample/src/main/resources/logback.xml
QuChen88: Will remove.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Added a `connect()` method to fetch the DB username from connection properties. 

However, I noticed from local PG testing that I didn't have to specify `user` property in my connection properties. If `user` is not specified, the underlying driver was able to discover my username `quchen` and use it to connect to my database, so I think the discovery logic here is still required.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
QuChen88: I didn't change this code. It was re-factored from its original place. See https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main/wrapper/src/main/java/software/amazon/jdbc/plugin/DataCacheConnectionPlugin.java#L75

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: I tried this way, however I ran into some limitations when string variables in a lambda expression need to be `final`. But a lot of them are not.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: Like I said in the other one, the parsing logic here is relatively simple. I don't feel strongly about using a regular expression here because it is more complex and error prone.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
QuChen88: Yes, it is mainly to support scenarios when the application tries to retrieve `Time` in a different time zone from the time zone that it was stored as. This can be done by specifying `cal` parameter in `convertToTime(Object timeObj, Calendar cal)` API.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataRemoteCachePlugin.java
QuChen88: Moved this logic into `TargetDriverDialect.java` with some sanity checks for when `toString()` doesn't return the expected format. Treated `null` return value as the query not cacheable.

Added unit tests.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: changed to `NESTED`

.env.example
sergiyvamz: Let's move this file closer to a corresponding code.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
sergiyvamz: Yes I see. Since you're already working on both plugins, please fix it. Lets make default value null. Thank you.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -0,0 +1,461 @@
...
+      DatabaseMetaData metadata = currentConn.getMetaData();
+      // Fetch and record the schema name if the session state doesn't currently have it
+      SessionStateService sessionStateService = pluginService.getSessionStateService();
+      Optional<String> catalog = sessionStateService.getCatalog();
+      Optional<String> schema = sessionStateService.getSchema();
+      String catalogName = catalog.orElse(null);
+      String schemaName = schema.orElse(null);
+      // If catalog and schema names are not present in sessionStateService,
+      // fetch them from the DB connection and store them in sessionStateService.
+      if (!catalog.isPresent() && !schema.isPresent()) {
```
karenc-bq, line 158: Do we need to worry about there case where one is present but the other one is not? This only updates the value if both are null

karenc-bq: What do you think of something like
```java
      String catalogName = catalog.orElse(currentConn.getCatalog());
      String schemaName = schema.orElse(currentConn.getSchema());
      // If catalog and schema names are not present in sessionStateService,
      // fetch them from the DB connection and store them in sessionStateService.
      if (!catalog.isPresent()) {
        sessionStateService.setCatalog(catalogName);
      }
      if (!schema.isPresent()) {
        sessionStateService.setSchema(schemaName);
      }
```

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,855 @@
...
+          ? this.awsProfileProperties
+          : new Properties();
+      this.credentialsProvider = AwsCredentialsManager.getProvider(null, propsToPass);
+    } else {
+      this.credentialsProvider = null;
+    }
+  }
+
+  // for unit testing only
+  public CacheConnection(final Properties properties) {
```
karenc-bq:
```suggestion
  CacheConnection(final Properties properties) {
```

We can make this package private

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
```diff
@@ -0,0 +1,855 @@
...
+        // Ignore shutdown errors during cluster mode detection
+      }
+    }
+  }
+
+  /* Here we check if we need to initialise connection pool for read or write to cache.
+  With isRead we check if we need to initialise connection pool for read or write to cache.
+  If isRead is true, we initialise connection pool for read.
+  If isRead is false, we initialise connection pool for write.
+   */
```
karenc-bq:
```suggestion
  /**
   * Initializes the connection pool for reading from or writing to the cache, if not already initialized.
   *
   * @param isRead if {@code true}, initializes the read connection pool; otherwise, initializes the write connection pool
   */
```

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/ElastiCacheIamTokenUtility.java
```diff
@@ -0,0 +1,129 @@
...
+  private final Aws4Signer signer;
+
+  public ElastiCacheIamTokenUtility(String cacheName) {
+    this.cacheName = Objects.requireNonNull(cacheName, "cacheName cannot be null");
+    this.clock = Clock.systemUTC();
+    this.signer = Aws4Signer.create();
+  }
+
+  // For testing only
+  public ElastiCacheIamTokenUtility(String cacheName, Instant fixedInstant, Aws4Signer signer) {
```
karenc-bq: Aws4Signer is deprecated.
I know the existing `LightRdsUtility` also uses the deprecated module, changing it in that file would cause a breaking change, but does it make sense to use AwsV4HttpSigner in this new plugin?
@sergiyvamz what do you think?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: I did consider this scenario. As long as I try to fetch them at least once and get at least one of them filled in, then I am fine. That is why I only update when both are not present. 

To add, for both MySQL and Postgres, I noticed one of them can be null

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Will change

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/ElastiCacheIamTokenUtility.java
QuChen88: Good news - `ElastiCacheIamTokenUtility.java` will be removed in the revision tomorrow when we bring in glide :)

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CacheConnection.java
QuChen88: Fixed

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
QuChen88: OK. Made the default value null.

.env.example
QuChen88: Removed this file.

QuChen88: Removed this file. Added comment in the example program instead

wrapper/src/test/java/integration/TestEnvironmentInfo.java
QuChen88: Renamed to `getValkeyServerInfo()`

PR #1534, opened 2025-09-15
Service container utility

aaron-congo:


wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -370,6 +363,20 @@ private ReaderFailoverResult getResultFromNextTaskBatch(
     return new ReaderFailoverResult(null, null, false);
   }
 
+  protected FullServicesContainer getNewServicesContainer() throws SQLException {
+    return ServiceUtility.getInstance().createServiceContainer(
```
sergiyvamz: Just a suggestion. I've seen such piece of code at least twice. I think it makes sense to create a new method:
`ServiceUtility.createNewFromSource(FullServicesContainer source, PluginService pluginService, Properties props)`

wrapper/src/main/java/software/amazon/jdbc/util/ServiceUtility.java
```diff
@@ -0,0 +1,90 @@
...
+import software.amazon.jdbc.util.monitoring.MonitorService;
+import software.amazon.jdbc.util.storage.StorageService;
+import software.amazon.jdbc.util.telemetry.TelemetryFactory;
+
+public class ServiceUtility {
+  private static volatile ServiceUtility instance;
+  private static final ReentrantLock initLock = new ReentrantLock();
+
+  private ServiceUtility() {
+    if (instance != null) {
```
sergiyvamz: What is it for?

wrapper/src/main/java/software/amazon/jdbc/util/ServiceUtility.java
```diff
@@ -0,0 +1,90 @@
...
+      throw new IllegalStateException("ServiceContainerUtility singleton instance already exists.");
+    }
+  }
+
+  public static ServiceUtility getInstance() {
+    if (instance != null) {
```
sergiyvamz: This class is a simple stateless utility class. I think we can simplify logic a bit and use pre-initialized instance of ServiceUtility.


```
private static final ServiceUtility INSTANCE = new SericeUtility;

public ServiceUtility getInstance() { return INSTANCE; }
```

wrapper/src/main/java/software/amazon/jdbc/util/connection/ConnectionService.java
```diff
@@ -21,7 +21,16 @@
...
+import software.amazon.jdbc.util.FullServicesContainer;
 
+/**
+ * A service used to open new connections for internal driver use.
+ *
+ * @deprecated This interface is deprecated and will be removed in a future version. Use
+ *     {@link software.amazon.jdbc.util.ServiceUtility#createServiceContainer} followed by
+ *     {@link PluginService#forceConnect} instead.
+ */
+@Deprecated
```
sergiyvamz, line 33: We're going to release 3.0 with some breaking changes. Do we want to add one more breaking change and completely remove this interface and implementation class?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
sergiyvamz: Is it the same pluginService that is sitting in serviceContainer? Can we use
`this.servicesContainer.getPluginService()` instead?

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: Good idea, thanks!

wrapper/src/main/java/software/amazon/jdbc/util/connection/ConnectionService.java
aaron-congo: Sounds like we will keep these changes as part of a minor version bump instead of a major one, but I agree we should remove this when we bump to 3.0

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareReaderFailoverHandler.java
aaron-congo: I added the requested method, I did however keep the "getNewServicesContainer" method because it allows me to mock the service container creation in our unit tests

PR #1537, opened 2025-09-18
Feat/weighted random new

aaronchung-bitquill:


wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,168 @@
...
+        final int rangeEnd = counter + hostWeight - 1;
+        hostWeightRangeMap.put(host, new NumberRange(rangeStart, rangeEnd));
+        counter = counter + hostWeight;
+      } else {
+        hostWeightRangeMap.put(host, new NumberRange(counter, counter));
+        counter++;
+      }
+    }
+
+    // Check random number is in host weigh range map
```
aaron-congo: it seems like this comment is not accurate, should we remove it?

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,168 @@
...
+    if (this.random == null) {
+      this.random = new Random();
+    }
+    int randomInt = this.random.nextInt(counter);
+
+    // This block is for testing purposes
```
aaron-congo: Should we remove this block if its for testing purposes?

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,168 @@
...
+    if (this.randomFunc != null) {
+      try {
+        randomInt = this.randomFunc.call();
+      } catch (Exception e) {
+        // This should not happen
+      }
+    }
+
+    for (final Entry<HostSpec, NumberRange> entry : hostWeightRangeMap.entrySet()) {
+      if (hostWeightRangeMap.get(entry.getKey()).isInRange(randomInt)) {
```
aaron-congo:
```suggestion
      if (entry.getValue().isInRange(randomInt)) {
```

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,168 @@
...
+        .filter(hostSpec ->
+            role.equals(hostSpec.getRole()) && hostSpec.getAvailability().equals(HostAvailability.AVAILABLE))
+        .sorted(Comparator.comparing(HostSpec::getHost))
+        .collect(Collectors.toList());
+
+    if (eligibleHosts.isEmpty()) {
+      throw new SQLException(Messages.get("HostSelector.noHostsMatchingRole", new Object[] {role}));
+    }
+
+    final Map<HostSpec, NumberRange> hostWeightRangeMap = new HashMap<>();
```
aaron-congo: Something just occurred to me, I've read that mutable objects should not be used as keys in hash sets or hash maps because their hashcode changes when the object changes which can lead to lost entries, inconsistent lookup results etc. I looked in our code and I think there are some places where we use mutable keys, as a separate task we should have someone look through the code and clean these up

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
aaronchung-bitquill: Ah! That was supposed to be further down. I will move it

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
aaronchung-bitquill: This is used by unit tests so i can mock the random number generation. 
If its okay with you, I'll leave it in, but make the comment more clear.

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,168 @@
...
+    throw new SQLException(Messages.get("HostSelector.weightedRandomUnableToGetHost", new Object[] {role}));
+  }
+
+  private Map<String, Integer> getHostWeightPairMap(final String hostWeightMapString) throws SQLException {
+    if (this.cachedHostWeightMapString != null
+        && this.cachedHostWeightMapString.trim().equals(hostWeightMapString.trim())
+        && this.cachedHostWeightMap != null
+        && !this.cachedHostWeightMap.isEmpty()) {
+      return this.cachedHostWeightMap;
+    }
```
aaronchung-bitquill: Caching code might require locks

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
aaronchung-bitquill: Doh. I just realized i might actually be able to do this mockito instead. I'll be trying that.

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
aaronchung-bitquill: Oh thats a good one. Thanks for pointing that out. Will fix.

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,169 @@
...
+        counter++;
+      }
+    }
+
+    if (this.random == null) {
+      this.random = new Random();
+    }
+    int randomInt = this.random.nextInt(counter);
+
+    // Check random number is in host weigh range map
```
aaron-congo:
```suggestion
    // Check random number is in host weight range map
```

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -0,0 +1,169 @@
...
+    for (final HostSpec host : eligibleHosts) {
+      if (hostWeightRangeMap.containsKey(host.getHost())
+          && hostWeightRangeMap.get(host.getHost()).isInRange(randomInt)) {
+        return host;
```
aaron-congo: is there a chance that the key is removed in between the containsKey and isInRange call? If so we would get an NPE.
```suggestion
      NumberRange range = hostWeightRangeMap.get(host.getHost());
      if (range != null && range.isInRange(randomInt)) {
        return host;
```

aaronchung-bitquill: I think with the way its currently written, that shouldnt happen. But, this could change in the future and guarding against this sounds like a good idea. I'll take your suggestion!

PR #1541, opened 2025-09-26
refactor: decouple dependencies and clean up initialization logic

aaron-congo:
- before this refactor, the ConnectionWrapper and ConnectionPluginManager classes both depended on each other, and the PluginServiceImpl and ConnectionPluginManager classes also depended on each other. This PR removes the ConnectionPluginManager's dependencies on the ConnectionWrapper and PluginService
- also cleans up initialization logic by performing service container setup in the ServiceUtility class. This class can be used to either create service containers for client connections or for more limited ("minimal") monitor use

### DIFF
--- a/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
+++ b/docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
@@ -378,7 +378,7 @@ public static void main(String[] args) throws SQLException {
 public static void main(String[] args) throws SQLException {
 
     final Properties properties = new Properties();
-    properties.setProperty("wrapperPlugins", "awsSecretsManagers");
+    properties.setProperty("wrapperPlugins", "awsSecretsManager");
     properties.setProperty("secretsManagerSecretId", "secretId");
     properties.setProperty("secretsManagerRegion", "us-east-2");
 


wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
```diff
@@ -449,14 +476,7 @@ public static <T> T createInstance(
...
 
-      Class<?>[] argClasses = constructorArgClasses;
-      if (argClasses == null) {
-        argClasses = new Class<?>[constructorArgs.length];
-        for (int i = 0; i < constructorArgs.length; i++) {
-          argClasses[i] = constructorArgs[i].getClass();
-        }
-      }
-      final Constructor<?> constructor = classToInstantiate.getConstructor(argClasses);
+      final Constructor<?> constructor = classToInstantiate.getConstructor(constructorArgClasses);
```
sergiyvamz, line 479: I'd be careful with this change. There was a bug. Please double check:

https://github.com/aws/aws-advanced-jdbc-wrapper/commit/ca0b8fdf2f42bbea1e315c162cfb679b8430af66#diff-d0ef1b1788499382bf76f53bc7fc3cbf0fed1a23fc0896c10d2d7afa34202d0bR284-R289

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
aaron-congo: The reason I removed this was because my IDE was telling me that we never enter the if-block. If you look at the code immediately above this, we will have returned if constructorArgClasses was null, so argClasses will always be equal to constructorArgClasses here. It looks like the `constructorArgClasses == null` check above this was added after the commit that you linked. Does that make sense or am I missing something?

PR #1543, opened 2025-09-29
KMS Encryption Plugin

davecramer:
### Summary

Plugin to enable encrypting columns on the client


docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
```diff
@@ -0,0 +1,337 @@
...
+- **Metadata-Driven**: Configurable encryption based on table and column metadata
+- **Audit Logging**: Optional audit logging for encryption operations
+- **Minimal Performance Impact**: Efficient encryption with caching and optimized operations
+
+## Prerequisites
+
+- AWS KMS key with appropriate permissions
+- Database table to store encryption metadata
+- AWS credentials configured (via IAM roles, profiles, or environment variables)
+- **JSqlParser 4.5.x dependency** - Required for SQL parsing and analysis
```
alexey-amazon, line 18: we're in the SQL parsing business now!

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Setup
+
+### 1. Create Encryption Metadata Table
+
+First, create the required tables to store encryption metadata and keys:
+
+```sql
+-- Key storage table (must be created first due to foreign key)
+CREATE TABLE key_storage (
```
alexey-amazon, line 158: Do we generally rely on user connecting to a database which has this table? What if I have a user with multiple schemas available to me, where this table needs to be? Also of course, what is multiple tables with the same name exist in different schemas? Could there be a any confusion as to what metadata table is relevant to them?

alexey-amazon: should we have tables prefixed with  like "aws_" or configurable via properties?

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
```diff
@@ -0,0 +1,337 @@
...
+
+```java
+KeyManagementUtility keyManagementUtility = new KeyManagementUtility(
+    keyManager, metadataManager, dataSource, kmsClient);
+
+// Configure encryption for sensitive columns
+keyManagementUtility.initializeEncryptionForColumn("customers", "ssn", masterKeyArn);
+keyManagementUtility.initializeEncryptionForColumn("customers", "credit_card", masterKeyArn);
+keyManagementUtility.initializeEncryptionForColumn("customers", "phone", masterKeyArn);
+keyManagementUtility.initializeEncryptionForColumn("customers", "address", masterKeyArn);
```
alexey-amazon, line 194: It feels like to make sure there is no sneaky insert of unencrypted data by mistake the table should have a trigger which looks for some KMS-related prefixes in the values inserted and reject if it doesn't find them. Maybe we should recommend  customers setting something up. I believe KMS-encrypted strings have specific look to them

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
```diff
@@ -0,0 +1,337 @@
...
+
+```java
+props.setProperty("enableAuditLogging", "true");
+```
+
+Check the application logs for encryption-related messages.
+
+## Limitations
+
+- Currently supports string data types for encryption
```
alexey-amazon, line 321: I assume ORDER BY or LIKE are not much useful since server side cannot figure it out.

However, if I do group by encrypted column, will it work?

wrapper/build.gradle.kts
```diff
@@ -49,6 +50,7 @@ dependencies {
     optionalImplementation("io.opentelemetry:opentelemetry-api:1.52.0")
     optionalImplementation("io.opentelemetry:opentelemetry-sdk:1.52.0")
     optionalImplementation("io.opentelemetry:opentelemetry-sdk-metrics:1.52.0")
+    optionalImplementation("com.github.jsqlparser:jsqlparser:4.5") // JSqlParser SQL parser (Java 8 compatible)
```
alexey-amazon, line 53: have we done a security review / audit for this dependency?

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
davecramer: No choice... that is the only way to tell if a column requires encryption (at the moment)

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
davecramer: sure, implementation detail... could be in an aws schema

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
davecramer: nope, won't work. You can't do any sorting, indexing etc on an encrypted column. You would need another column with say the last 4 digits if the ssn to do equaliity

wrapper/build.gradle.kts
davecramer: no, but I may replace it with my own.

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
alexey-amazon: If the same value is re-written to the encrypted column, is it going to have a different cyphertext?

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
alexey-amazon: This will cover all possible paths how the data might end up inserted into the table, such as via PL/SQL function or copied from another table...

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
davecramer: It will have a different cypher text. The encryption is non-determistic.

docs/using-the-jdbc-driver/using-plugins/UsingTheKmsEncryptionPlugin.md
davecramer: This is certainly a valid concern. Since this will work with MySQL as well we may have to figure something out.

sergiyvamz (approved):

PR #1546, opened 2025-10-01
fix: isValidTimeout potentially rounded down to 0

karenc-bq:
### Summary

`isValidTimeout` could potentially rounded down, causing the monitoring connections to send a isValid request without any timeout.


wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorImpl.java
```diff
@@ -325,7 +325,8 @@ boolean checkConnectionStatus() {
       // Some drivers, like MySQL Connector/J, execute isValid() in a double of specified timeout time.
       final int validTimeout = (int) TimeUnit.NANOSECONDS.toSeconds(
           this.failureDetectionIntervalNano - THREAD_SLEEP_NANO) / 2;
-      return this.monitoringConn.isValid(validTimeout);
+      // validTimeout could get rounded down to 0.
+      return this.monitoringConn.isValid(Math.max(1, validTimeout));
```
sergiyvamz: Does it make sense to define a const MIN_VALIDITY_CHECK_TIMEOUT_SEC?

PR #1553, opened 2025-10-09
docs: document clusterId requirement when using non-standard RDS URL

aaron-congo:


docs/using-the-jdbc-driver/UsingTheJdbcDriver.md
```diff
@@ -72,26 +72,27 @@ logging.level.software.amazon.jdbc=trace
 ## AWS Advanced JDBC Driver Parameters
 These parameters are applicable to any instance of the AWS JDBC Driver.
 
-| Parameter                                         | Value     | Required | Description                                                                                                                                                                                                                                                                                                                                          | Default Value  |
```
aaron-congo, line 75: I added the same row for cluster ID here and in the docs for failover1, not sure if we need it in all places but probably doesn't hurt since its important

PR #1560, opened 2025-10-15
fix: update closedExplicitly flag

karenc-bq:
### Summary

The closedExplicitly flag is never set. Update failover plugin to update the flag when the closed method is called.


wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -206,6 +206,9 @@ public <T, E extends Exception> T execute(
     }
 
     if (canDirectExecute(methodName)) {
+      if (JdbcMethod.CONNECTION_CLOSE.methodName.equals(methodName)) {
```
sergiyvamz: close or abort?

PR #1565, opened 2025-10-16
chore: remove mutable keys from HostSpec#hashCode and document immutability requirements

aaron-congo:


wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
```diff
@@ -212,7 +212,7 @@ public String toString() {
 
   @Override
   public int hashCode() {
-    return Objects.hash(this.host, this.port, this.availability, this.role, this.weight, this.lastUpdateTime);
```
aaron-congo, line 215: availability, weight, and update time are all mutable - we should avoid using mutable fields in the hashcode implementation

sergiyvamz: There're many parts of the code that relies on this particular implementation. I'd start with reviewing every place where HostSpec is compared (including sorting). There're few cases: 1) comparing entire HostSpec including mutable fields 2) comparing based on host and port 3) comparing based on immutable fields. 

For now it's hard to make conclusion whether this change is breaking change or not.

aaron-congo: Note that I'm changing the hashCode implementation here, but am not changing the equals method. I believe the equals method is what is used in comparison and sorting, not the hashCode. I think the hashCode is only used in hash-based collections, and hashCodes should not have mutable fields, otherwise entries can get lost in the collection. Looking online, I think its fine if hashCode uses an immutable subset of all fields used in the equals implementation.

sergiyvamz (approved):

PR #1581, opened 2025-10-27
feat: pooled flag for connect

sergiyvamz:
### Summary

- breaking change: `ConnectionProvider.connect()` changed
- add call context
- define a new call context flag `isPooledConnection`

### Additional Reviewers

@karenc-bq 
@aaron-congo 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/ConnectionInfo.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ConnectionInfo.java
@@ -16,13 +16,26 @@
 
 package software.amazon.jdbc;
 
-import org.checkerframework.checker.nullness.qual.Nullable;
+import java.sql.Connection;
 
-public interface PluginManagerService {
+public class ConnectionInfo {
+  private final Connection connection;
+  private final boolean isPooled;
 
-  void setInTransaction(boolean inTransaction);
+  public ConnectionInfo(Connection connection, boolean isPooled) {
+    this.connection = connection;
+    this.isPooled = isPooled;
+  }
 
-  void setIsPooledConnection(@Nullable Boolean pooledConnection);
+  public ConnectionInfo(Connection connection) {
+    this(connection, false);
+  }
 
-  void resetCallContext();
+  public Connection getConnection() {
+    return connection;
+  }
+
+  public boolean isPooled() {
+    return isPooled;
+  }
 }
--- a/wrapper/src/main/java/software/amazon/jdbc/ConnectionProvider.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ConnectionProvider.java
@@ -78,10 +78,10 @@ public interface ConnectionProvider {
    * @param targetDriverDialect the target driver dialect
    * @param hostSpec the HostSpec containing the host-port information for the host to connect to
    * @param props    the Properties to use for the connection
-   * @return {@link Connection} resulting from the given connection information
+   * @return {@link ConnectionInfo} resulting from the given connection information
    * @throws SQLException if an error occurs
    */
-  Connection connect(
+  @NonNull ConnectionInfo connect(
       @NonNull String protocol,
       @NonNull Dialect dialect,
       @NonNull TargetDriverDialect targetDriverDialect,
--- a/wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
@@ -215,6 +215,7 @@ public class PartialPluginService implements PluginService, CanReleaseResources,
     return this.connectionProviderManager.getDefaultProvider();
   }
 
+  @Deprecated
   public boolean isPooledConnectionProvider(HostSpec host, Properties props) {
     final ConnectionProvider connectionProvider =
         this.connectionProviderManager.getConnectionProvider(this.driverProtocol, host, props);
@@ -650,6 +651,26 @@ public class PartialPluginService implements PluginService, CanReleaseResources,
     } catch (SQLException e) {
       return false;
     }
+  }
+
+  @Override
+  public Boolean isPooledConnection() {
+    // This service implementation doesn't support call context.
+    throw new UnsupportedOperationException(
+        Messages.get("PartialPluginService.unexpectedMethodCall", new Object[] {"getSessionStateService"}));
+  }
+
+  @Override
+  public void setIsPooledConnection(Boolean pooledConnection) {
+    // This service implementation doesn't support call context.
+    // Do nothing.
+  }
+
+  @Override
+  public void resetCallContext() {
+    // This service implementation doesn't support call context.
+    throw new UnsupportedOperationException(
+        Messages.get("PartialPluginService.unexpectedMethodCall", new Object[] {"getSessionStateService"}));
   }
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/PluginService.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/PluginService.java
@@ -233,6 +233,7 @@ public interface PluginService extends ExceptionHandler, Wrapper {
 
   ConnectionProvider getDefaultConnectionProvider();
 
+  @Deprecated
   boolean isPooledConnectionProvider(HostSpec host, Properties props);
 
   String getDriverProtocol();
@@ -248,4 +249,19 @@ public interface PluginService extends ExceptionHandler, Wrapper {
   <T> T getPlugin(final Class<T> pluginClazz);
 
   boolean isPluginInUse(final Class<? extends ConnectionPlugin> pluginClazz);
+
+  // JDBC call context functions
+
+  /**
+  * Retrieves details about the most recent {@link PluginService#connect} or
+  * {@link PluginService#forceConnect} calls. Specifically indicates whether the
+  * returned connection was obtained from a connection pool or newly created.
+  *
+  * <p>Note: The {@link ConnectionPlugin} must process or store this information during
+  * the current JDBC call, as these details will be reset before the next JDBC call
+  * is processed, or another {@link PluginService#connect} or {@link PluginService#forceConnect}
+  * is made.
+  *
+  */
+  @Nullable Boolean isPooledConnection();
 }
--- a/wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
@@ -91,6 +91,9 @@ public class PluginServiceImpl implements PluginService, CanReleaseResources,
 
   protected final ReentrantLock connectionSwitchLock = new ReentrantLock();
 
+  // JDBC call context members
+  protected Boolean pooledConnection = null;
+
   public PluginServiceImpl(
       @NonNull final FullServicesContainer servicesContainer,
       @NonNull final Properties props,
@@ -243,6 +246,7 @@ public class PluginServiceImpl implements PluginService, CanReleaseResources,
     return this.connectionProviderManager.getDefaultProvider();
   }
 
+  @Deprecated
   public boolean isPooledConnectionProvider(HostSpec host, Properties props) {
     final ConnectionProvider connectionProvider =
         this.connectionProviderManager.getConnectionProvider(this.driverProtocol, host, props);
@@ -787,6 +791,21 @@ public class PluginServiceImpl implements PluginService, CanReleaseResources,
     } catch (SQLException e) {
       return false;
     }
+  }
+
+  @Override
+  public Boolean isPooledConnection() {
+    return this.pooledConnection;
+  }
+
+  @Override
+  public void setIsPooledConnection(Boolean pooledConnection) {
+    this.pooledConnection = pooledConnection;
+  }
+
+  @Override
+  public void resetCallContext() {
+    this.pooledConnection = null;
   }
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
@@ -31,6 +31,7 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.ConnectionInfo;
 import software.amazon.jdbc.ConnectionPlugin;
 import software.amazon.jdbc.ConnectionProvider;
 import software.amazon.jdbc.ConnectionProviderManager;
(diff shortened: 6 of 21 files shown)


wrapper/src/main/java/software/amazon/jdbc/ConnectionProvider.java
```diff
@@ -78,10 +78,10 @@ HostSpec getHostSpecByStrategy(
    * @param targetDriverDialect the target driver dialect
    * @param hostSpec the HostSpec containing the host-port information for the host to connect to
    * @param props    the Properties to use for the connection
-   * @return {@link Connection} resulting from the given connection information
+   * @return {@link ConnectionInfo} resulting from the given connection information
    * @throws SQLException if an error occurs
    */
-  Connection connect(
+  ConnectionInfo connect(
```
aaron-congo: Can we add @NonNull here? I don't think it would make sense for this method to return null. It should either succeed and return a non-null value, or throw a SQLException if the connect attempt failed

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -204,14 +205,17 @@ private Connection connectInternal(
       }
     }
 
-    this.connProviderManager.initConnection(conn, driverProtocol, hostSpec, props);
+    if (connectionInfo != null) {
```
aaron-congo: See my comment above, if possible I think `connect` should always return a non-null value. Then we can remove the null checks here.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
```diff
@@ -652,6 +653,24 @@ public boolean isPluginInUse(final Class<? extends ConnectionPlugin> pluginClazz
     }
   }
 
+  @Override
+  public Boolean isPooledConnection() {
+    // This service implementation doesn't support call context.
+    return null;
```
aaron-congo: The other methods that we don't want called in this class throw an UnsupportedOperationException in this class, what do you think of doing the same here? Or just returning false? Then the method could return a boolean value, which would be nice because we wouldn't have to do null checks or deal with the ambiguity of what a null value means. It looks like we currently only use this in the ReadWriteSplittingPlugin, and both places we use it assign a false value if it is null anyways. For example, in this snippet:
```
this.isReaderConnFromInternalPool =
            this.pluginService.isPooledConnection() != null && this.pluginService.isPooledConnection();
```
this.isReaderConnFromInternalPool will be set to false if isPooledConnection returns null

wrapper/src/main/java/software/amazon/jdbc/ConnectionInfo.java
```diff
@@ -0,0 +1,37 @@
...
+
+package software.amazon.jdbc;
+
+import java.sql.Connection;
+
+public class ConnectionInfo {
+  private final Connection connection;
+  private final boolean isPooled;
+
+  public ConnectionInfo(Connection connection, boolean isPooled) {
```
karenc-bq, line 25: Can we overload the constructor with a default value for isPooled
```java
public ConnectionInfo(Connection connection) { this(connection, false); }
```

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
```diff
@@ -217,6 +217,8 @@ public static <T> T executeWithPlugins(
         context.setAttribute("jdbcCall", jdbcMethod.methodName);
       }
 
+      connectionWrapper.getServicesContainer().getPluginService().resetCallContext();
```
aaron-congo: Why do we reset the isPooledConnection flag here? If I understand correctly, if we are using a pooled connection, this means that the isPooled flag will be cleared at the first jdbc call, even though the connection is still pooled.

wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
```diff
@@ -789,6 +793,19 @@ public boolean isPluginInUse(final Class<? extends ConnectionPlugin> pluginClazz
     }
   }
 
+  @Override
+  public Boolean isPooledConnection() {
+    return this.pooledConnection;
+  }
+
+  public void setPooledConnection(Boolean pooledConnection) {
```
karenc-bq: nit: can we rename this to `setIsPooledConnection` for clarity

wrapper/src/main/java/software/amazon/jdbc/PluginService.java
```diff
@@ -248,4 +249,12 @@ Connection forceConnect(
   <T> T getPlugin(final Class<T> pluginClazz);
 
   boolean isPluginInUse(final Class<? extends ConnectionPlugin> pluginClazz);
+
+  // JDBC call context functions
+
+  Boolean isPooledConnection();
```
aaron-congo: If we leave this as `Boolean` instead of `boolean` (see my other comments), can we please add the Nullable annotation here so that the compiler will give a warning if we try to use it as a regular `boolean`?

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -204,14 +205,17 @@ private Connection connectInternal(
...
+      this.pluginService.setPooledConnection(connectionInfo.isPooled());
+      this.connProviderManager.initConnection(connectionInfo.getConnection(), driverProtocol, hostSpec, props);
+    }
 
     this.pluginService.setAvailability(hostSpec.asAliases(), HostAvailability.AVAILABLE);
```
aaron-congo: Does it make sense to set the host as available if connectionInfo is null?

wrapper/src/test/java/software/amazon/jdbc/ConnectionPluginManagerTests.java
```diff
@@ -43,6 +43,7 @@
 import java.util.logging.Logger;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
```
karenc-bq: Is this an unused import

wrapper/src/test/java/software/amazon/jdbc/HikariPooledConnectionProviderTest.java
```diff
@@ -141,7 +141,9 @@ void testConnectWithDefaultMapping() throws SQLException {
     Properties props = new Properties();
     props.setProperty(PropertyDefinition.USER.name, user1);
     props.setProperty(PropertyDefinition.PASSWORD.name, password);
-    try (Connection conn = provider.connect(protocol, mockDialect, mockTargetDriverDialect, mockHostSpec, props)) {
+    ConnectionInfo connectionInfo =
+        provider.connect(protocol, mockDialect, mockTargetDriverDialect, mockHostSpec, props);
+    try (Connection conn = connectionInfo.getConnection()) {
```
karenc-bq: Curious, does it make sense to make `ConnectionInfo` implement `AutoCloseable`?
```java
try (ConnectionInfo connectionInfo = provider.connect(protocol, mockDialect, mockTargetDriverDialect, mockHostSpec, props);
Connection conn = connectionInfo.getConnection()) {
  // do stuff
}
If I understood correctly the connection is being opened in this line 
```java
ConnectionInfo connectionInfo =
        provider.connect(protocol, mockDialect, mockTargetDriverDialect, mockHostSpec, props);
```
so this line could throw exceptions, and should be wrapped in try-catch.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
sergiyvamz: I wanted the flag has tree values. True and false values when it's set up by Default plugin and null when the flag not yet applicable (for instance when the driver executes a SQL query). For simplicity, RW Splitting plugin treats null value as a regular (not pooled) connection. I really believe that getting null in this case is impossible but it's better to check this null case.

`PartialPluginService` is designed to be multithread safe and, ideally, to be stateless. That's why it's safer to return null stating it's inconclusive about the nature of connection. Raising an exception may be a showstopper so I prefer to avoid it.

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
sergiyvamz: If a plugin calls `pluginService.connect()` or `forceConnect()` then `isPooledConnection` flag will be overridden so it may no longer correspond to an initial connection. That's why the flag is available right after opening a new connection and it should be handled (stored) by plugins immediately. 

My idea was to have a call context and collect various flags and supplementary objects (like SQL query analysis results) to be available among all plugins during processing a particular JDBC call. Such call context should be reset at the beginning of a next call.

wrapper/src/test/java/software/amazon/jdbc/HikariPooledConnectionProviderTest.java
sergiyvamz: I reworked the test code and use try-finally to control connections.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
aaron-congo: The PartialPluginService currently throws UnsupportedOperationExceptions from 13 methods. How are we deciding when to throw this exception vs when to do nothing or return null?

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
aaron-congo: Few concerns with this:
1. I see your point about clearing this flag because it is ambiguous whether it applies to an initial connection or a connect/forceConnect connection. We still have this problem with the current code, except that its now limited to the duration when this is set. So I think the problem is not as bad as it could be, but it is still a problem, since plugins still won't be able to tell which connection it applies to.
2. If I had not reviewed this PR and I wanted to use isPooledConnection(), I would expect the flag to be set after connecting and not be cleared. For future users, it is not obvious that the flag is only set for a short duration
3. what if future code/plugins want to know whether a connection is pooled outside the short duration of time when this is set?

An idea I have that I think would address these points would be to change the return type of the connect/forceConnect pipeline from Connection to ConnectionInfo. By keeping the isPooled flag tied to the Connection, it is clear that the flag applies to the connection that it sits with in ConnectionInfo. The isPooled information would also continue to be available instead of only being available in a short snippet of time.

wrapper/src/main/java/software/amazon/jdbc/ConnectionInfo.java
```diff
@@ -0,0 +1,41 @@
...
+    this.connection = connection;
+    this.isPooled = isPooled;
+  }
+
+  public ConnectionInfo(Connection connection) {
+    this(connection, false);
+  }
+
+  public Connection getConnection() {
```
aaron-congo, line 34:
```suggestion
  private final @NonNull Connection connection;
  private final boolean isPooled;

  public ConnectionInfo(@NonNull Connection connection, boolean isPooled) {
    this.connection = connection;
    this.isPooled = isPooled;
  }

  public ConnectionInfo(@NonNull Connection connection) {
    this(connection, false);
  }

  public @NonNull Connection getConnection() {
```

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -204,14 +205,17 @@ private Connection connectInternal(
       }
     }
 
-    this.connProviderManager.initConnection(conn, driverProtocol, hostSpec, props);
+    this.pluginService.setIsPooledConnection(connectionInfo.isPooled());
+    this.connProviderManager.initConnection(connectionInfo.getConnection(), driverProtocol, hostSpec, props);
 
-    this.pluginService.setAvailability(hostSpec.asAliases(), HostAvailability.AVAILABLE);
+    if (connectionInfo.getConnection() != null) {
```
aaron-congo, line 211: I think we can mark this method with a NonNull return value, and then we don't have to make this check

wrapper/src/main/java/software/amazon/jdbc/ConnectionInfo.java
sergiyvamz: Unfortunately connection could be null. I recall few cases when MariaDb driver returns null.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
sergiyvamz: Unfortunately connection could be null.

wrapper/src/main/java/software/amazon/jdbc/PartialPluginService.java
sergiyvamz: There's no strict rule. `UnsupportedOperationExceptions` is a protection from calling method that are not supported. It's for developers to catch such illegal calls and fix the code.

I added raise an `UnsupportedOperationExceptions` exception for `isPooledConnection()` and `resetCallContext()`. `setIsPooledConnection()` is used in `DefaultConnectionPlugin` and the call legit so method implementation should not raise any exceptions.

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
sergiyvamz: Changing the return type of the connect/forceConnect pipeline from `Connection` to `ConnectionInfo` was my original idea. After applying this change I ended with massive changes in pipelines and plugins. When I got more than 60 file changed I reconsider this approach in favour of changing `ConnectionProvider` interface only.

1, 3. Plugin need to collect the flag (if needed) and make sure it corresponds to a particular connection. `PluginService` provides this flag as part of call context and a plugin needs to take care of flag usage.
2. I added javadoc to the method explaining its usage.

aaron-congo: Yeah, I get how it works now, it just seems strange and non-intuitive to me. When the read-write plugin opens a new reader connection on setReadOnly, it is easy for the r/w plugin to know that the flag refers to the new connection it is opening, but other plugins that want to use the flag won't know that. I think the ConnectionInfo approach is better. I know the change would touch many files, but we are already changing a lot for version 3 and won't be able to make such big changes after. But I'll let you decide what you want to do, just let me know so I know you've seen this and then I'll give you an approval

sergiyvamz: > but other plugins that want to use the flag won't know that. 

Do you mean the case when that connection is set as a current connection then it's not possible to say if it's pooled one?

aaron-congo: I don't think I understand your question, but I was talking about the scenario when setReadOnly(true) is called so we are switching from a writer to a reader. The r/w plugin calls `plugin.connect()` to create the reader connection. Once the reader connection is established in the default plugin, the isPooled flag is set, but it will not be set as the current connection until we return back up the plugin chain to the r/w plugin. In this time window it isn't necessarily clear for other plugins whether the isPooled flag refers to the current writer connection or the reader connection that is about to be set as the current connection.

sergiyvamz: The flag refers to the recent `connect()` or `forceConnect()` call. It doesn't refer to the current connection. Other plugins don't have context about RW Splitting plugin and whether this new connection is reader or writer connection, or whether it becomes a new current connection.

PR #1583, opened 2025-10-28
feat: add support of RDS Proxy custom endpoints

sergiyvamz:
### Summary

add support of RDS Proxy custom endpoints

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@sophia-bq 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
@@ -73,6 +73,20 @@ public class RdsUtils {
   // Governmental endpoints
   // https://aws.amazon.com/compliance/fips/#FIPS_Endpoints_by_Service
   // https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/Region.html
+  //
+  //
+  // Aurora Global Database
+  // https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.Aurora_Fea_Regions_DB-eng.Feature.GlobalDatabase.html
+  // Global Database Endpoint: <gdb-name>.global-<xyz>.global.rds.amazonaws.com
+  // Example: test-global-db-name.global-123456789012.global.rds.amazonaws.com
+  //
+  //
+  // RDS Proxy
+  // RDS Proxy Endpoint: <proxy-name>.proxy-<xyz>.<aws-region>.rds.amazonaws.com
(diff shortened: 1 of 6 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
```diff
@@ -372,6 +425,8 @@ public RdsUrlType identifyRdsType(final String host) {
       return RdsUrlType.RDS_AURORA_LIMITLESS_DB_SHARD_GROUP;
     } else if (isRdsProxyDns(host)) {
       return RdsUrlType.RDS_PROXY;
+    } else if (isRdsProxyEndpointDns(host)) {
+      return RdsUrlType.RDS_PROXY_ENDPOINT;
```
karenc-bq, line 429: Curious why do we need these? Do the existing `isRdsProxyDns` and `RDS_PROXY` not cover the same cases? Could we remove one of these?

sergiyvamz: Some of our plugins are not compatible with RDS Proxy so it's important to properly recognize all type of urls related to RDS Proxy. we can reuse the same RDS_PROXY but I think it's better to be precise.

sophia-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/util/RdsUtils.java
```diff
@@ -73,6 +73,20 @@ public class RdsUtils {
...
+  // Global Database Endpoint: <gdb-name>.global-<xyz>.global.rds.amazonaws.com
+  // Example: test-global-db-name.global-123456789012.global.rds.amazonaws.com
+  //
+  //
+  // RDS Proxy
+  // RDS Proxy Endpoint: <proxy-name>.proxy-<xyz>.<aws-region>.rds.amazonaws.com
+  // Example: test-rds-proxy-name.proxy-123456789012.us-east-2.rds.amazonaws.com
+  //
+  // RDS Proxy Custom Endpoint: <custom-endpoint-name>.endpoint.proxy-<xyz>.<aws-region>.rds.amazonaws.com
+  // Example: test-custom-endpoint-name.proxy-123456789012.us-east-2.rds.amazonaws.com
```
karenc-bq:
```suggestion
  // Example: test-custom-endpoint-name.endpoint.proxy-123456789012.us-east-2.rds.amazonaws.com
```

karenc-bq (approved):

PR #1586, opened 2025-10-30
feat: srw

sophia-bq:
### Summary

introduces the Simple Read/Write Splitting Plugin (srw)

### Description

an alternative to the existing Read/Write Splitting Plugin that connects directly to specified read and write endpoints


docs/using-the-jdbc-driver/CompatibilityCrossPlugins.md
```diff
@@ -88,9 +92,11 @@
...
+| [srw](./using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md)                      | <span style="color:yellow;font-size:15px">&check;</span> | <span style="color:yellow;font-size:15px">&check;</span> | <span style="color:yellow;font-size:15px">&check;</span> | <span style="color:yellow;font-size:15px">&check;</span>                              |
 
 <br>
 
-| Plugin codes / Plugin codes                      | [limitless](./using-plugins/UsingTheLimitlessConnectionPlugin.md) | [bg](./using-plugins/UsingTheBlueGreenPlugin.md) | 
-|--------------------------------------------------|-------------------------------------------------------------------|--------------------------------------------------|
-| [bg](./using-plugins/UsingTheBlueGreenPlugin.md) | <span style="color:red;font-size:20px">&cross;</span>             |                                                  | 
+| Plugin codes / Plugin codes                                      | [limitless](./using-plugins/UsingTheLimitlessConnectionPlugin.md) | [bg](./using-plugins/UsingTheBlueGreenPlugin.md)         |
```
sergiyvamz, line 99: `srw` is not compatible with `readWriteSplitting` plugin.

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
+# Simple Read/Write Splitting Plugin
+
+The simple read/write splitting plugin adds functionality to switch between endpoints via calls to the `Connection#setReadOnly` method. Based off the values provided in the properties, upon calling `setReadOnly(true)`, the plugin will connect to the given endpoint for read operations. When `setReadOnly(false)` is called, the plugin will connect to the given endpoint for write operations. Future calls to `setReadOnly` will switch between the established writer and reader connections according to the boolean argument you supply to the `setReadOnly` method.
+
+## Loading the Simple Read/Write Splitting Plugin
+
+The simple read/write splitting plugin is not loaded by default. To load the plugin, include it in the `wrapperPlugins` connection parameter. If you would like to load the simple read/write splitting plugin alongside the failover and host monitoring plugins, the simple read/write splitting plugin must be listed before these plugins in the plugin chain. If it is not, failover exceptions will not be properly processed by the plugin. See the example below to properly load the read/write splitting plugin with these plugins.
```
sergiyvamz: > the simple read/write splitting plugin must be listed before these plugins in the plugin chain


That's a bit outdated statement. The driver sort plugins automatically so plugin order isn't so important. However, if user opt out with auto sorting then certainly `srw` should go before `failover` and `efm`.

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you would like to use the simple read/write splitting plugin without the failover plugin, make sure you have the `srw` plugin in the `wrapperPlugins` property, and that the failover plugin is not part of it.
+```
+final Properties properties = new Properties();
+properties.setProperty(PropertyDefinition.PLUGINS.name, "srw");
+properties.setProperty("srwWriteEndpoint", "test-db.cluster-XYZ.us-east-2.rds.amazonaws.com");
+properties.setProperty("srwReadEndpoint", "test-db.cluster-ro-XYZ.us-east-2.rds.amazonaws.com");
+```
+
+## Federated Authentication Plugin Parameters
```
sergiyvamz: Federated? Should it be SRW Plugin Parameters?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+| `srwWriteEndpoint`        | String  |   Yes    | The endpoint to connect to when `setReadOnly(false)` is called.                               | `null`        | `<cluster-name>.cluster-<XYZ>.<region>.rds.amazonaws.com`    |
+| `srwReadEndpoint`         | String  |    No    | The endpoint to connect to when `setReadOnly(true)` is called.                                | `null`        | `<cluster-name>.cluster-ro-<XYZ>.<region>.rds.amazonaws.com` |
+| `verifyNewSrwConnections` | Boolean |    No    | Enables role-verification for new connections made by the simple read/write splitting plugin. | `true`        | `true`, `false`                                              |
+
+### Limitations of `verifyNewSrwConnections`
+
+The property `verifyNewSrwConnections` is enabled by default. This means that when new connections are made with the simple read/write splitting plugin, a query is sent to the new connection to verify its role. If a write connection connects to a reader, an error is thrown. If a read connection connects to the writer, a warning is logged and the connection is not stored for reuse.
+
+#### Non-Aurora clusters
```
sergiyvamz: Non-Aurora and non RDS Cluster?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+The verification step uses an Aurora specific query to determine the role of the connection, if the endpoint is not to an Aurora database, disable `verifyNewSrwConnections` by setting it to `false` in the properties.
+
+#### Autocommit
+If `verifyNewSrwConnections` is enabled and autocommit is set to false when the simple read/write splitting plugin makes a new connection, unexpected errors such as `Cannot change transaction read-only property in the middle of a transaction.` will occur. As part of verification, the plugin executes a query against a new connection, which when autocommit is set to false will open a transaction.
```
sergiyvamz: Can we rephrase it? It's hard to understand when the occurs and what to do to fix it.

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+If `verifyNewSrwConnections` is enabled and autocommit is set to false when the simple read/write splitting plugin makes a new connection, unexpected errors such as `Cannot change transaction read-only property in the middle of a transaction.` will occur. As part of verification, the plugin executes a query against a new connection, which when autocommit is set to false will open a transaction. 
+If autocommit is essential to a workflow, either establish both internal reader and writer connections before setting autocommit to false or disable `verifyNewSrwConnections` by setting it to `false` in the properties.
+
+## Supplying the connection string
+
+When using the simple read/write splitting plugin against Aurora clusters, you do not have to supply multiple instance URLs in the connection string. Instead, supply just the URL for the initial instance to which you're connecting. You must also include either the failover plugin or the Aurora host list plugin in your plugin chain so that the driver knows to query Aurora for its topology. See the section on [loading the simple read/write splitting plugin](#loading-the-simple-readwrite-splitting-plugin) for more info.
```
sergiyvamz: Aurora and RDS clusters?

sergiyvamz: What is `Aurora host list plugin` ?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+## Using the Simple Read/Write Splitting Plugin with RDS Proxy
+
+RDS Proxy provides connection pooling and management that significantly improves application scalability by reducing database connection overhead and enabling thousands of concurrent connections through
+connection multiplexing. Connecting exclusively through the proxy endpoint ensures consistent connection management, automatic failover handling, and centralized monitoring, while protecting the underlying database from connection exhaustion
+and providing a stable abstraction layer that remains consistent even when database topology changes. By providing the read/write endpoint and a read-only endpoint to the simple read/write splitting plugin, the AWS JDBC Driver will connect using 
+these endpoints any time setReadOnly is called. 
+
+To take full advantage of the benefits of RDS Proxy it is recommended to only connect through RDS Proxy endpoints. See [Using the AWS JDBC Driver with RDS Proxy](./../../../README.md#rds-proxy) for limitations.
+
+## Using the Simple Read/Write Splitting Plugin against non-Aurora clusters
```
sergiyvamz: non-Aurora and non-RDS Clusters?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+
+The simple read/write splitting plugin can be used to switch between any two endpoints. If the endpoints do not direct to an Aurora database, ensure the property `verifyNewSrwConnections` is set to `false`. See [Limitations of verifyNewSrwConnections](UsingTheSimpleReadWriteSplittingPlugin.md#non-aurora-clusters) for details.
```
sergiyvamz: the same, Aurora or RDS Cluster

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
...
+
+Verify plugin compatibility within your driver configuration using the [compatibility guide](../Compatibility.md).
+
+### Failover
+
+Immediately following a failover event, due to DNS caching, an Aurora cluster endpoint may connect to the previous writer, and the read-only endpoint may connect to the new primary instance. To avoid stale DNS connections, as well as errors/warnings from the verification logic we recommend including the [Aurora Initial Connection Strategy Plugin](./UsingTheAuroraInitialConnectionStrategyPlugin.md) in your configuration.
```
sergiyvamz: I think that using `initialConnection` plugin ruins `srw` plugin. I don't think they're compatible.

docs/using-the-jdbc-driver/CompatibilityCrossPlugins.md
sergiyvamz: and it's not compatible with `initialConnection`.

sergiyvamz: and not with `auroraStaleDns`

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,78 @@
+# Simple Read/Write Splitting Plugin
+
+The simple read/write splitting plugin adds functionality to switch between endpoints via calls to the `Connection#setReadOnly` method. Based off the values provided in the properties, upon calling `setReadOnly(true)`, the plugin will connect to the given endpoint for read operations. When `setReadOnly(false)` is called, the plugin will connect to the given endpoint for write operations. Future calls to `setReadOnly` will switch between the established writer and reader connections according to the boolean argument you supply to the `setReadOnly` method.
```
sergiyvamz: We need to mention that this plugin doesn't rely on cluster topology but just on provided endpoints and their DNS resolution.

examples/AWSDriverExample/src/main/java/software/amazon/SimpleReadWriteSplittingMySQLExample.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final String PASSWORD = "password";
+  private static final String SRW_WRITE_ENDPOINT = "test-db.cluster-XYZ.us-east-2.rds.amazonaws.com";
+  private static final String SRW_READ_ENDPOINT = "test-db.cluster-ro-XYZ.us-east-2.rds.amazonaws.com";
+
+  public static void main(String[] args) throws SQLException {
+
+    final Properties props = new Properties();
+
+    // Enable srw, failover, and efm2 plugins and set properties
+    props.setProperty(PropertyDefinition.PLUGINS.name, "srw,failover,efm2");
```
sergiyvamz: Lets use `failover2` to promote it.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -130,9 +130,6 @@ public void initHostProvider(
       final JdbcCallable<Void, SQLException> initHostProviderFunc) throws SQLException {
 
     this.hostListProviderService = hostListProviderService;
-    if (hostListProviderService.isStaticHostListProvider()) {
```
sergiyvamz: why?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,526 @@
...
+      throw new
+          RuntimeException(
+          Messages.get(
+              "SimpleReadWriteSplittingPlugin.missingRequiredConfigParameter",
+              new Object[] {SRW_WRITE_ENDPOINT.name}));
+    }
+    this.pluginService = pluginService;
+    this.properties = properties;
+    this.writeEndpoint = writeEndpoint;
+    this.readEndpoint = SRW_READ_ENDPOINT.getString(properties);
```
sergiyvamz: Does it mean that SRW_READ_ENDPOINT is optional parameter?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,526 @@
...
+        && (!this.verifyNewConnections || this.pluginService.getHostRole(currentConnection) == HostRole.WRITER)) {
+      setWriterConnection(currentConnection, currentHost);
+    } else if (isReadEndpoint(currentHost) && !currentConnection.equals(this.readerConnection)
+        && (!this.verifyNewConnections || this.pluginService.getHostRole(currentConnection) == HostRole.READER)) {
+      setReaderConnection(currentConnection, currentHost);
+    }
+  }
+
+  private boolean isWriteEndpoint(final @NonNull HostSpec hostSpec) {
+    return Objects.equals(hostSpec.getHost(), this.writeEndpoint);
```
sergiyvamz: We need to use getHostAndPort() and verify equalsIgnoreCase()

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,526 @@
...
+  }
+
+  private void getNewWriterConnection() throws SQLException {
+    if (this.writeEndpointHostSpec == null) {
+      this.writeEndpointHostSpec = createHostSpec(this.writeEndpoint, HostRole.WRITER);
+    }
+    final Connection conn = this.pluginService.connect(this.writeEndpointHostSpec, this.properties, this);
+
+    if (this.verifyNewConnections && this.pluginService.getHostRole(conn) != HostRole.WRITER) {
+      logAndThrowException(
```
sergiyvamz, line 567: Need retry logic as discussed

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,526 @@
...
+          Messages.get("SimpleReadWriteSplittingPlugin.writerEndpointConnectedToReader",
+              new Object[]{this.writeEndpoint}));
+    }
+
+    setWriterConnection(conn, writeEndpointHostSpec);
+    switchCurrentConnectionTo(this.writerConnection, writeEndpointHostSpec);
+  }
+
+  private void getNewReaderConnection() throws SQLException {
```
sergiyvamz: Need retry logic as discussed

wrapper/src/test/java/integration/container/tests/SimpleReadWriteSplittingTest.java
```diff
@@ -0,0 +1,192 @@
...
+    TestEnvironmentFeatures.BLUE_GREEN_DEPLOYMENT,
+    TestEnvironmentFeatures.RUN_DB_METRICS_ONLY})
+@EnableOnNumOfInstances(min = 2)
+@MakeSureFirstInstanceWriter
+@Order(15)
+
+
+public class SimpleReadWriteSplittingTest extends ReadWriteSplittingTests {
+  String pluginCode = "srw";
+  String pluginCodesWithFailover = "failover,efm2,srw,initialConnection";
```
sergiyvamz: failover2

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
sophia-bq: I've updated it so that SRW_READ_ENDPOINT is required

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
sophia-bq: Are we expecting that the port is included in the endpoint?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -346,35 +346,38 @@ void initHostProvider(
 
   @Override
   public void notifyNodeListChanged(final Map<String, EnumSet<NodeChangeOptions>> changes) {
+    try {
+      if (!this.enableFailoverSetting) {
```
karenc-bq: Curious, if we don't have failover enabled do we still want to call `this.staleDnsHelper.notifyNodeListChanged(changes);`?
Do we want to move this outside of the try-finally

sophia-bq: Good catch, I think moving the check for failover enabled outside of the try catch would be better

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+The Simple Read/Write Splitting Plugin is not loaded by default. To load the plugin, include it in the `wrapperPlugins` connection parameter. 
+
+```
+final Properties properties = new Properties();
+properties.setProperty(PropertyDefinition.PLUGINS.name, "srw,failover2,efm2");
+properties.setProperty("srwWriteEndpoint", "test-db.cluster-XYZ.us-east-2.rds.amazonaws.com");
+properties.setProperty("srwReadEndpoint", "test-db.cluster-ro-XYZ.us-east-2.rds.amazonaws.com");
+```
+
+If you would like to use the Simple Read/Write Splitting Plugin without the failover plugin, make sure you have the `srw` plugin in the `wrapperPlugins` property, and that the failover plugin is not part of it.
```
aaron-congo: I know we use a similar blurb to this for the docs for the RW plugin but I can't remember why we mentioned this for this specific combination of plugins. Should we consider removing it? It seems self-explanatory to me.

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+|------------------------------|:-------:|:--------:|:----------------------------------------------------------------------------------------------------------------------------------------------|---------------|--------------------------------------------------------------|
+| `srwWriteEndpoint`           | String  |   Yes    | The endpoint to connect to when `setReadOnly(false)` is called.                                                                               | `null`        | `<cluster-name>.cluster-<XYZ>.<region>.rds.amazonaws.com`    |
+| `srwReadEndpoint`            | String  |   Yes    | The endpoint to connect to when `setReadOnly(true)` is called.                                                                                | `null`        | `<cluster-name>.cluster-ro-<XYZ>.<region>.rds.amazonaws.com` |
+| `verifyNewSrwConnections`    | Boolean |    No    | Enables role verification for new connections made by the Simple Read/Write Splitting Plugin.                                                 | `true`        | `true`, `false`                                              |
```
aaron-congo:
```suggestion
| `verifyNewSrwConnections`    | Boolean |    No    | Enables writer/reader role verification for new connections made by the Simple Read/Write Splitting Plugin.                                                 | `true`        | `true`, `false`                                              |
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+If you would like to use the Simple Read/Write Splitting Plugin without the failover plugin, make sure you have the `srw` plugin in the `wrapperPlugins` property, and that the failover plugin is not part of it.
+```
+final Properties properties = new Properties();
+properties.setProperty(PropertyDefinition.PLUGINS.name, "srw");
+properties.setProperty("srwWriteEndpoint", "test-db.cluster-XYZ.us-east-2.rds.amazonaws.com");
+properties.setProperty("srwReadEndpoint", "test-db.cluster-ro-XYZ.us-east-2.rds.amazonaws.com");
+```
+
+## Simple Read/Write Splitting Plugin Parameters
+| Parameter                    |  Value  | Required | Description                                                                                                                                   | Default Value | Example Value                                                |
```
aaron-congo:
```suggestion
| Parameter                    |  Value  | Required | Description                                                                                                                                   | Default Value | Example Values                                                |
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
aaron-congo: What is the difference between `verifyNewSrwConnections` and `verifyOpenedConnectionType`? It sounds like the first enables verification of both readers and writers when setReadOnly is called, but the second enables verification for only "writers" or only "readers"? If the first is set but the second is not, does it verify both?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+
+## How the Simple Read/Write Splitting Plugin Verifies Connections
+
+The property `verifyNewSrwConnections` is enabled by default. This means that when new connections are made with the Simple Read/Write Splitting Plugin, a query is sent to the new connection to verify its role. If the connection cannot be verified as having the correct role—that is, a write connection is not connected to a writer, or a read connection is not connected to a reader—the plugin will retry the connection up to the time limit of `srwConnectRetryTimeoutMs`. 
+
+The values of `srwConnectRetryTimeoutMs` and `srwConnectRetryIntervalMs` control the timing and aggressiveness of the plugin's retries.
+
+Additionally, to consistently ensure the role of connections made with the plugin, the plugin also provides role verification for the initial connection. When connecting with an RDS writer cluster or reader cluster endpoint, the plugin will retry the initial connection up to `srwConnectRetryTimeoutMs` until it has verified connection to the intended role of the endpoint.
```
aaron-congo:
```suggestion
Additionally, to consistently ensure the role of connections made with the plugin, the plugin also provides role verification for the initial connection. When connecting with an RDS writer cluster or reader cluster endpoint, the plugin will retry the initial connection up to `srwConnectRetryTimeoutMs` until it has verified the intended role of the endpoint.
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+If it is unable to return a verified initial connection, it will log a message and continue with the normal workflow of the other plugins.
+When connecting with custom endpoints and other non-standard URLs, role verification on the initial connection can also be triggered by providing the expected role through the `verifyOpenedConnectionType` parameter. Set this to `writer` or `reader` accordingly.
```
aaron-congo: Can we consider changing the name from verifyOpenedConnectionType to verifyInitialConnectionType? I think it would be a bit clearer

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
aaron-congo: Can we add "More details below" to the descriptions for the bottom 4 parameters?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+## How the Simple Read/Write Splitting Plugin Verifies Connections
+
+The property `verifyNewSrwConnections` is enabled by default. This means that when new connections are made with the Simple Read/Write Splitting Plugin, a query is sent to the new connection to verify its role. If the connection cannot be verified as having the correct role—that is, a write connection is not connected to a writer, or a read connection is not connected to a reader—the plugin will retry the connection up to the time limit of `srwConnectRetryTimeoutMs`. 
+
+The values of `srwConnectRetryTimeoutMs` and `srwConnectRetryIntervalMs` control the timing and aggressiveness of the plugin's retries.
+
+Additionally, to consistently ensure the role of connections made with the plugin, the plugin also provides role verification for the initial connection. When connecting with an RDS writer cluster or reader cluster endpoint, the plugin will retry the initial connection up to `srwConnectRetryTimeoutMs` until it has verified connection to the intended role of the endpoint.
+If it is unable to return a verified initial connection, it will log a message and continue with the normal workflow of the other plugins.
```
aaron-congo, line 37: Is this referring to the case when we successfully check the role but it does not align with what we expected? Or is this the case when we cannot determine the role we are connected to?

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+When connecting with custom endpoints and other non-standard URLs, role verification on the initial connection can also be triggered by providing the expected role through the `verifyOpenedConnectionType` parameter. Set this to `writer` or `reader` accordingly.
+
+## Limitations When Verifying Connections
+
+#### Non-RDS clusters
+The verification step determines the role of the connection by executing a query against it. If the endpoint is not part of an Aurora or RDS cluster, disable `verifyNewSrwConnections` by setting it to `false` in the properties.
```
aaron-congo:
```suggestion
The verification step determines the role of the connection by executing a query against it. If the endpoint is not part of an Aurora or RDS cluster, the plugin will not be able to verify the role, so `verifyNewSrwConnections` must be set to `false`.
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+- as part of setReadOnly, a new connection is opened
+- that connection's role is verified
+
+This is a result of the plugin executing the role-verification query against a new connection, and when autocommit is false, this opens a transaction. 
+
+If autocommit is essential to a workflow, either ensure the plugin has connected to the desired target connection of the setReadOnly query before setting autocommit to false or disable `verifyNewSrwConnections`. Examples of the former can be found in the [Simple Read/Write Splitting Examples](UsingTheSimpleReadWriteSplittingPlugin.md#examples).
+
+## Using the Simple Read/Write Splitting Plugin with RDS Proxy
+
+RDS Proxy provides connection pooling and management that significantly improve application scalability by reducing database connection overhead and enabling thousands of concurrent connections through
```
aaron-congo:
```suggestion
RDS Proxy provides connection pooling and management that significantly improves application scalability by reducing database connection overhead and enabling thousands of concurrent connections through
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+
+### General plugin limitations
+
+When a Statement or ResultSet is created, it is internally bound to the database connection established at that moment. There is no standard JDBC functionality to change the internal connection used by Statement or ResultSet objects. Consequently, even if the read/write plugin switches the internal connection, any Statements/ResultSets created before this will continue using the old database connection. This bypasses the desired functionality provided by the plugin. To prevent these scenarios, an exception will be thrown if your code uses any Statements/ResultSets created before a change in internal connection. To solve this problem, please ensure you create new Statement/ResultSet objects after switching between the writer/reader.
+
+Verify plugin compatibility within your driver configuration using the [compatibility guide](../Compatibility.md).
+
+### Failover
+
+Immediately following a failover event, due to DNS caching, an RDS cluster endpoint may connect to the previous writer, and the read-only endpoint may connect to the new primary instance.
```
aaron-congo:
```suggestion
Immediately following a failover event, due to DNS caching, an RDS cluster endpoint may connect to the previous writer, and the read-only endpoint may connect to the new writer instance. 
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,98 @@
...
+
+Verify plugin compatibility within your driver configuration using the [compatibility guide](../Compatibility.md).
+
+### Failover
+
+Immediately following a failover event, due to DNS caching, an RDS cluster endpoint may connect to the previous writer, and the read-only endpoint may connect to the new primary instance. 
+
+To avoid stale DNS connections, enable `verifyNewSrwConnections`, as this will retry the connection. Service for Aurora clusters is typically restored in less than 60 seconds, and often less than 30 seconds. RDS Proxy endpoints to Aurora databases can update in as little as 3 seconds. Depending on your configuration and cluster availability `srwConnectRetryTimeoutMs` and `srwConnectRetryIntervalMs` may be set to customize the timing of the retries.
```
aaron-congo:
```suggestion
To avoid stale DNS connections, enable `verifyNewSrwConnections`, as this will retry the connection until the role has been verified. Service for Aurora clusters is typically restored in less than 60 seconds, and often less than 30 seconds. RDS Proxy endpoints to Aurora databases can update in as little as 3 seconds. Depending on your configuration and cluster availability `srwConnectRetryTimeoutMs` and `srwConnectRetryIntervalMs` may be set to customize the timing of the retries.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -183,6 +183,7 @@ public void notifyNodeListChanged(final Map<String, EnumSet<NodeChangeOptions>>
         LOGGER.finest(() -> Messages.get("AuroraStaleDnsHelper.reset"));
         this.writerHostSpec = null;
         this.writerHostAddress = null;
+        return;
```
aaron-congo, line 187: I think this doesn't do anything so it can probably be removed

wrapper/src/test/java/integration/container/tests/SimpleReadWriteSplittingTest.java
```diff
@@ -0,0 +1,205 @@
...
+@EnableOnTestFeature(TestEnvironmentFeatures.FAILOVER_SUPPORTED)
+@DisableOnTestFeature({
+    TestEnvironmentFeatures.PERFORMANCE,
+    TestEnvironmentFeatures.RUN_HIBERNATE_TESTS_ONLY,
+    TestEnvironmentFeatures.RUN_AUTOSCALING_TESTS_ONLY,
+    TestEnvironmentFeatures.BLUE_GREEN_DEPLOYMENT,
+    TestEnvironmentFeatures.RUN_DB_METRICS_ONLY})
+@EnableOnNumOfInstances(min = 2)
+@MakeSureFirstInstanceWriter
+@Order(15)
```
aaron-congo: This is the same order as Failover2Test, is it fine to have both use "15"?

wrapper/src/test/java/integration/container/tests/SimpleReadWriteSplittingTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+  @Override
+  protected Properties getPropsWithFailover() {
+    return getSrwProps(false, pluginCodesWithFailover);
+  }
+
+  @Override
+  protected boolean connectedToCorrectReaderInstance(String readerConnectionId, String currentConnectionId) {
+    // On conn.setReadOnly(true), the SimpleReadWriteSplittingPlugin ensures connection to the reader endpoint.
+    // If connected to a reader instance originally, the connection will change: return true no matter values.
```
aaron-congo: A bit confused about this comment, are you meaning to say that we can just return true here because the srw plugin already verifies that we are connected to a reader automatically?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+  public static final AwsWrapperProperty SRW_WRITE_ENDPOINT =
+      new AwsWrapperProperty(
+          "srwWriteEndpoint",
+          null,
+          "The read-write/cluster endpoint that should be used to connect to the writer.");
+  public static final AwsWrapperProperty VERIFY_NEW_SRW_CONNECTIONS =
+      new AwsWrapperProperty(
+          "verifyNewSrwConnections",
+          "true",
+          "Enables role-verification for new connections made by the Simple Read Write Splitting Plugin.",
```
aaron-congo:
```suggestion
          "Enables role verification for new connections made by the Simple Read Write Splitting Plugin.",
```

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+
+    final RdsUrlType type = this.rdsUtils.identifyRdsType(hostSpec.getHost());
+
+    if (type == RdsUrlType.RDS_WRITER_CLUSTER
+        || type == RdsUrlType.RDS_GLOBAL_WRITER_CLUSTER
+        || this.verifyOpenedConnectionType == AuroraInitialConnectionStrategyPlugin.VerifyOpenedConnectionType.WRITER) {
+      Connection writerCandidateConn = this.getVerifiedConnection(props, hostSpec, HostRole.WRITER, connectFunc);
+      if (writerCandidateConn == null) {
+        // Can't get writer connection. Continue with a normal workflow.
+        return connectFunc.call();
```
aaron-congo: should we set the initial connection host spec in this case as well or no?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+      return writerCandidateConn;
+    }
+
+    if (type == RdsUrlType.RDS_READER_CLUSTER
+        || this.verifyOpenedConnectionType == AuroraInitialConnectionStrategyPlugin.VerifyOpenedConnectionType.READER) {
+      Connection readerCandidateConn =
+          this.getVerifiedConnection(props, hostSpec, HostRole.READER, connectFunc);
+      if (readerCandidateConn == null) {
+        // Can't get a reader connection. Continue with a normal workflow.
+        return connectFunc.call();
```
aaron-congo: same question as above

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+  private final boolean verifyNewConnections;
+  private volatile boolean inReadWriteSplit = false;
+  private HostListProviderService hostListProviderService;
+  private Connection writerConnection;
+  private Connection readerConnection;
+  private final String writeEndpoint;
+  private final String readEndpoint;
+  private HostSpec readEndpointHostSpec;
+  private HostSpec writeEndpointHostSpec;
+  private final AuroraInitialConnectionStrategyPlugin.VerifyOpenedConnectionType verifyOpenedConnectionType;
```
aaron-congo: Can we move VerifyOpenedConnectionType to its own separate class since it is now shared? The VERIFY_OPENED_CONNECTION_TYPE property can be moved there as well

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+      try {
+        if (hostSpec == null) {
+          if (connectFunc == null) {
+            // Unable to connect to verify role.
+            break;
+          }
+          // No HostSpec provided, still verify role.
+          candidateConn = connectFunc.call();
+        } else {
+          candidateConn = this.pluginService.connect(hostSpec, props, this);
```
aaron-congo, line 282: if hostSpec is not null and connectFunc is not null, should we use connectFunc instead of pluginService.connect?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,678 @@
...
+        // ignore
+      }
+    }
+  }
+
+  private void delay() {
+    try {
+      TimeUnit.MILLISECONDS.sleep(this.connectRetryIntervalMs);
+    } catch (InterruptedException ex) {
+      // ignore
```
aaron-congo:
```suggestion
      Thread.currentThread().interrupt()
```

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
sophia-bq: I agree. Changed the name :)

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
sophia-bq: Regardless of why we cannot verify the role, if a verified connection cannot be established within the timeout it continues with the normal workflow.

wrapper/src/test/java/integration/container/tests/SimpleReadWriteSplittingTest.java
sophia-bq: Updated it to 23

wrapper/src/test/java/integration/container/tests/SimpleReadWriteSplittingTest.java
sophia-bq: I've updated the test to better accomplish its intentions - this was only used for a reader instance initial connection and then calling setReadOnly(true)

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
sophia-bq: In the case that there are multiple nodes impacted and `changes` contains the writer promoted to a reader and then a reader promoted to the writer:

in the first iteration of the for loop this.writerHostSpec is set to null
in the second iteration of the for loop on line 181, this.writerHostSpec.getUrl() throws an exception

after the iteration where we reset there is nothing more to be reset and any further iterations of the for loop will throw exceptions - this is why I added the return

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
sergiyvamz: It could be included.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -392,8 +370,6 @@ private HostSpec getReader(final Properties props, final @Nullable String awsReg
         } else {
           return this.pluginService.getHostSpecByStrategy(HostRole.READER, strategy);
         }
-      } catch (UnsupportedOperationException ex) {
```
sergiyvamz, line 395: why?

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,672 @@
...
+
+    this.inReadWriteSplit = true;
+    if (!isConnectionUsable(this.writerConnection)) {
+      getNewWriterConnection();
+    } else {
+      switchCurrentConnectionTo(this.writerConnection, this.writeEndpointHostSpec);
+    }
+  }
+
+  private void getNewWriterConnection() throws SQLException {
```
sergiyvamz: Can we rename the method? `getXXX` has a strong association with a classic getter method that return a value.

sergiyvamz:
Please check the following PR and adopt similar logic for `this.isWriterConnFromInternalPool` and `this.isReaderConnFromInternalPool` flags.
https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1581

wrapper/src/main/java/software/amazon/jdbc/plugin/srw/SimpleReadWriteSplittingPluginFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package software.amazon.jdbc.plugin.srw;
+
+import java.util.Properties;
+import software.amazon.jdbc.ConnectionPlugin;
+import software.amazon.jdbc.ConnectionPluginFactory;
+import software.amazon.jdbc.PluginService;
+
+public class SimpleReadWriteSplittingPluginFactory implements ConnectionPluginFactory {
```
sergiyvamz: Can we use our new factory interface?
https://github.com/aws/aws-advanced-jdbc-wrapper/blob/main-3.x/wrapper/src/main/java/software/amazon/jdbc/ServicesContainerPluginFactory.java

docs/using-the-jdbc-driver/using-plugins/UsingTheSimpleReadWriteSplittingPlugin.md
aaron-congo: Okay gotcha. So the main thing the role verification does is log a message when the role is incorrect? I'm curious if users might expect or want stricter verification, eg throwing an exception. Could be a configurable setting whether to throw an exception or just log. Just a thought though, maybe this has already been considered

wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
aaron-congo: ah my bad, didn't see the for loop above

wrapper/src/main/java/software/amazon/jdbc/plugin/VerifyConnectionType.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.plugin;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum VerifyConnectionType {
+  WRITER,
```
aaron-congo: Since the enum values for this are the same as for the HostRole enum, should we consider merging them together? Just an idea

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
sophia-bq: I tested this out and since an UnsupportedOperationException is not a SQLException, if one is thrown above it will not be caught. I believe removing these lines should not change behaviour

PR #1588, opened 2025-10-31
Refactor: dialects and host list providers to reduce code duplication

aaron-congo:
### Summary
- Introduces TopologyUtils classes to define querying logic common to different database engine deployments 
  - logic common to all deployments -> TopologyUtils
  - logic common to Multi-AZ deployments -> MultiAzTopologyUtils
  - logic common to Aurora deployments -> AuroraTopologyUtils
  - logic common to Global deployments -> GlobalAuroraTopologyUtils
- cleans up code in the dialect and hostlistprovider packages

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/TopologyUtils.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/TopologyUtils.java
@@ -124,7 +124,6 @@ public abstract class TopologyUtils {
 
     int writerCount = writers.size();
     if (writerCount == 0) {
-      LOGGER.warning(() -> Messages.get("TopologyUtils.invalidTopology"));
       return null;
     } else if (writerCount == 1) {
       hosts.add(writers.get(0));


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/AuroraGlobalDbHostListProvider.java
```diff
@@ -1,114 +0,0 @@
...
-import software.amazon.jdbc.HostListProviderService;
-import software.amazon.jdbc.HostSpec;
-import software.amazon.jdbc.HostSpecBuilder;
-import software.amazon.jdbc.PropertyDefinition;
-import software.amazon.jdbc.util.ConnectionUrlParser;
-import software.amazon.jdbc.util.FullServicesContainer;
-import software.amazon.jdbc.util.RdsUtils;
-import software.amazon.jdbc.util.StringUtils;
-
-public class AuroraGlobalDbHostListProvider extends AuroraHostListProvider {
```
aaron-congo, line 38: Renamed this to GlobalAuroraHostListProvider to be consistent with GlobalAuroraMysqlDialect

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsMultiAzDbClusterListProvider.java
```diff
@@ -1,208 +0,0 @@
-/*
```
aaron-congo, line 1: The logic in this class has been moved to TopologyUtils/MultiAzTopologyUtils

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/AuroraGlobalDbMonitoringHostListProvider.java
```diff
@@ -1,111 +0,0 @@
-/*
```
aaron-congo, line 1: I renamed this to MonitoringGlobalAuroraHostListProvider to be consistent with GlobalAuroraMysqlDialect

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/GlobalDbClusterTopologyMonitorImpl.java
```diff
@@ -1,111 +0,0 @@
-/*
```
aaron-congo, line 1: I renamed this to GlobalAuroraTopologyMonitor

wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
```diff
@@ -33,79 +33,58 @@
 
 public class MysqlDialect implements Dialect {
 
+  protected static final String VERSION_QUERY = "SHOW VARIABLES LIKE 'version_comment'";
+  protected static final String HOST_ALIAS_QUERY = "SELECT CONCAT(@@hostname, ':', @@port)";
+
+  protected static final DialectUtils dialectUtils = new DialectUtils();
```
sergiyvamz: Why static?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/TopologyUtils.java
```diff
@@ -0,0 +1,236 @@
...
+    for (HostSpec host : allHosts) {
+      if (HostRole.WRITER == host.getRole()) {
+        writers.add(host);
+      } else {
+        hosts.add(host);
+      }
+    }
+
+    int writerCount = writers.size();
+    if (writerCount == 0) {
```
sergiyvamz, line 126: We might need to review this condition. It seems that during failover no-writer topology is quite possible.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/GlobalAuroraTopologyMonitor.java
```diff
@@ -0,0 +1,78 @@
...
+        instanceTemplate,
+        refreshRateNano,
+        highRefreshRateNano);
+
+    this.instanceTemplatesByRegion = instanceTemplatesByRegion;
+    this.topologyUtils = topologyUtils;
+  }
+
+  @Override
+  protected HostSpec getinstanceTemplate(String instanceId, Connection connection) throws SQLException {
```
sergiyvamz: getInstanceTemplate (capital I)

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorServiceImpl.java
```diff
@@ -64,7 +63,6 @@ public class MonitorServiceImpl implements MonitorService, EventSubscriber {
         TimeUnit.MINUTES.toNanos(15), TimeUnit.MINUTES.toNanos(3), recreateOnError);
 
     suppliers.put(ClusterTopologyMonitorImpl.class, () -> new CacheContainer(defaultSettings, Topology.class));
```
sergiyvamz, line 65: A side question. Should we use implementation class name or an interface class name?

wrapper/src/main/java/software/amazon/jdbc/dialect/MysqlDialect.java
aaron-congo: I wasn't sure whether to do static or not here. It is a stateless class so it should be fine if multiple threads use it, and if its static we don't need to make multiple instances of it. But I'm fine with making it non-static too.

wrapper/src/main/java/software/amazon/jdbc/util/monitoring/MonitorServiceImpl.java
aaron-congo: I think implementation class, because we may at some point have multiple implementation classes for the same interface with each requiring different cache settings

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/TopologyUtils.java
aaron-congo: Sure, what did you have in mind? We could query the connection for the writer ID if there are no writers in the host list. But if the problem scenario we are considering is failover, I'm not sure if that would be reliable during failover or not.

PR #1591, opened 2025-11-04
test: enable IAM for multi-az integration tests

karenc-bq:
### Summary

IAM is now supported for RDS Multi-AZ clusters, should enable IAM testing for such deployments

### DIFF
--- a/wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
+++ b/wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
@@ -112,8 +112,8 @@ public class TestEnvironmentProvider implements TestTemplateInvocationContextPro
               // Multi-AZ clusters supports only 3 instances
               continue;
             }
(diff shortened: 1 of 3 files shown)


wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
```diff
@@ -112,8 +112,8 @@ public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContex
               // Multi-AZ clusters supports only 3 instances
               continue;
             }
-            if (deployment == DatabaseEngineDeployment.RDS_MULTI_AZ_INSTANCE && numOfInstances != 1) {
-              // Multi-AZ Instances supports only 1 instance
+            if (deployment == DatabaseEngineDeployment.RDS_MULTI_AZ_INSTANCE && numOfInstances > 2) {
```
sergiyvamz, line 115: That's not correct. RDS Instance is always 1 instance. Standby reader, if it's presented/deployed, isn't visible through topology or any other ways.

PR #1599, opened 2025-11-15
fix outdated topology during B/G switchover caused wrong connection failover

sergiyvamz:
### Summary

fix outdated topology during BG switchover caused wrong connection failover

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@sophia-bq 


docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
```diff
@@ -66,15 +66,15 @@ Verify plugin compatibility within your driver configuration using the [compatib
...
+| `bgSwitchoverTimeoutMs`           | Integer |                              No                              | Maximum duration (in milliseconds) allowed for switchover completion. If the switchover process stalls or exceeds this timeframe, the driver will automatically assume completion and resume normal operations.                                                                                                                                                                                                                                         | `180000`                 | `180000`      |
+| ~~`bgSuspendNewBlueConnections`~~ | Boolean |                              No                              | **Deprecated. Adding this parameter causes no changes in plugin behaviour.** <br/><br/>Enables Blue/Green Deployment switchover to suspend new blue connection requests while the switchover process is in progress.                                                                                                                                                                                                                                    | `false`                  | `false`       |
```
karenc-bq:
```suggestion
| ~~`bgSuspendNewBlueConnections`~~ | Boolean |                              No                              | **Deprecated. New blue connection requests will always be suspended during switchover for versions >=v2.6.7.** <br/><br/>Enables Blue/Green Deployment switchover to suspend new blue connection requests while the switchover process is in progress.                                                                                                                                                                                                                                    | `false`                  | `false`       |
```

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -450,12 +458,64 @@ public void monitor() throws Exception {
       this.monitoringConnection.set(null);
       this.closeConnection(conn);
 
+      this.servicesContainer.getEventPublisher().unsubscribe(
+          this, Collections.singleton(MonitorResetEvent.class));
+
+
```
karenc-bq:
```suggestion
```

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -450,12 +458,64 @@ public void monitor() throws Exception {
...
   }
 
+  protected void reset() {
+    LOGGER.finest(
+        String.format("Reset: clusterId=%s, host=%s", this.clusterId, this.initialHostSpec.getHost()));
+
+    this.nodeThreadsStop.set(true);
+    this.shutdownNodeExecutorService();
+    this.nodeThreadsStop.set(false);
+    this.submittedNodes.clear();
```
karenc-bq, line 472: Does it make sense to clear the submittedNodes in `this.shutdownNodeExecutorService();`? So we never forget to call it like we did in `stop()` and in the `finally` block in `monitor()`?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -267,6 +270,8 @@ public void close() {
     this.closeConnection(this.monitoringConnection.get());
     this.closeConnection(this.nodeThreadsWriterConnection.get());
     this.closeConnection(this.nodeThreadsReaderConnection.get());
```
karenc-bq: Do we also need to set these atomic references back to null here?

karenc-bq: Or does it make sense to have `closeConnection()` accept a `AtomicReference<Connection>` as a parameter and handle it internally

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
```diff
@@ -73,6 +73,8 @@ public class BlueGreenConnectionPlugin extends AbstractConnectionPlugin {
     PropertyDefinition.registerPluginProperties(BlueGreenConnectionPlugin.class);
   }
 
+  protected static final String BG_SKIP_ROUTING_IN_FORCE_CONNECT = "3a864d24-568f-4b55-a227-6f649ae3021a";
```
karenc-bq, line 76: Why can't we use `pluginToSkip` parameter for forceConnect and have to rely on this parameter?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -509,6 +521,25 @@ protected void updateDnsFlags(BlueGreenRole role, BlueGreenInterimStatus interim
...
+          .map(Entry::getKey)
+          .collect(Collectors.toSet());
+
+      // A special case for efm related monitors.
+      // TODO: remove this code as soon as efm1 monitors are migrated to MonitorService
+      HostMonitorThreadContainer.resetMonitors(blueEndpoints);
+
+      // Notify all monitors to reset if needed.
+      this.servicesContainer.getEventPublisher().publish(new MonitorResetEvent(this.clusterId, blueEndpoints));
+      this.storeMonitorResetTime(eventName);
```
karenc-bq, line 534: Can we also remove the `getInterimStatusHash` method low, it is no longer in use

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -1031,28 +1112,35 @@ protected void logSwitchoverFinalSummary() {
       return;
     }
 
+    final int maxEventNameLength = this.phaseTimeNano.keySet().stream()
+        .map(x -> x.length() + 5)
+        .max(Comparator.comparingInt(x -> x))
+        .orElse(31);
```
karenc-bq: What are the 5 and the 31?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
sergiyvamz: That's different kind of skip. It helps to configure a hostlist provider to follow BG routes or not.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
sergiyvamz: I added constants that explain their usage. It's just for pretty logging.

docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
```diff
@@ -66,15 +66,15 @@ Verify plugin compatibility within your driver configuration using the [compatib
...
+| `bgSwitchoverTimeoutMs`           | Integer |                              No                              | Maximum duration (in milliseconds) allowed for switchover completion. If the switchover process stalls or exceeds this timeframe, the driver will automatically assume completion and resume normal operations.                                                                                                                                                                                                                                         | `180000`                 | `180000`      |
+| ~~`bgSuspendNewBlueConnections`~~ | Boolean |                              No                              | **Deprecated. New blue connection requests are always be suspended during switchover for versions v2.6.7+.** <br/><br/>Enables Blue/Green Deployment switchover to suspend new blue connection requests while the switchover process is in progress.                                                                                                                                                                                                    | `false`                  | `false`       |
```
aaron-congo:
```suggestion
| ~~`bgSuspendNewBlueConnections`~~ | Boolean |                              No                              | **Deprecated. New blue connection requests are always suspended during switchover for versions v2.6.7+.** <br/><br/>Enables Blue/Green Deployment switchover to suspend new blue connection requests while the switchover process is in progress.                                                                                                                                                                                                    | `false`                  | `false`       |
```

docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
```diff
@@ -108,30 +108,33 @@ properties.setProperty("blue-green-monitoring-socketTimeout", "10000");
...
-5. Initiate the Blue/Green Deployment switchover through the AWS Console, CLI, or RDS API.
-6. Monitor the process until the switchover completes successfully or rolls back. This may take several minutes.
-7. Review the switchover summary in the application logs. This requires setting the log level to `FINE` for either the entire package `software.amazon.jdbc.plugin.bluegreen` or specifically for the class `software.amazon.jdbc.plugin.bluegreen.BlueGreenStatusProvider`.
-8. Update your application by deactivating the `bg` plugin through its removal from your application configuration. Redeploy your application afterward. Note that an active Blue/Green plugin produces no adverse effects once the switchover has been completed.
-9. Delete the Blue/Green Deployment through the appropriate AWS interface.
-10. The sequence of steps 8 and 9 is flexible and can be executed in either order based on your preference.
+2. If you're planning to connect with a non-admin user, add necessary permissions to allow the plugin to access Blue/Green metadata. Make sure that permissions are granted on **both** blue and green clusters.
```
aaron-congo:
```suggestion
2. If you're planning to connect with a non-admin user, add the necessary permissions to allow the plugin to access Blue/Green metadata. Make sure that permissions are granted on **both** blue and green clusters. 
```

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -450,12 +455,61 @@ public void monitor() throws Exception {
...
+
       LOGGER.finest(() -> Messages.get(
           "ClusterTopologyMonitorImpl.stopMonitoringThread",
           new Object[]{this.initialHostSpec.getHost()}));
     }
   }
 
+  protected void reset() {
+    LOGGER.finest(
+        String.format("Reset: clusterId=%s, host=%s", this.clusterId, this.initialHostSpec.getHost()));
```
aaron-congo: Should we be using Messages.get instead? Or does it not really matter?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -450,12 +455,61 @@ public void monitor() throws Exception {
...
+    this.nodeThreadsStop.set(false);
+    this.submittedNodes.clear();
+    this.createNodeExecutorService();
+
+    this.nodeThreadsWriterConnection.set(null);
+    this.nodeThreadsWriterHostSpec.set(null);
+    this.nodeThreadsReaderConnection.set(null);
+    this.nodeThreadsLatestTopology.set(null);
+
+    try {
```
aaron-congo: is the try block necessary? Doesn't look like any of the statements below will throw any exceptions

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -450,12 +455,61 @@ public void monitor() throws Exception {
...
+      this.highRefreshRateEndTimeNano = 0;
+      this.requestToUpdateTopology.set(false);
+      this.ignoreNewTopologyRequestsEndTimeNano.set(-1);
+    } finally {
+      this.closeConnection(this.monitoringConnection);
+    }
+
+    this.clearTopologyCache();
+
+    // It breaks a waiting/sleeping cycles in monitoring thread
```
aaron-congo:
```suggestion
    // This breaks any waiting/sleeping cycles in the monitoring thread
```

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
aaron-congo: Just for my own understanding, can you clarify how this skip works vs pluginToSkip? I don't understand your statement above

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
```diff
@@ -132,6 +153,18 @@ public Connection connect(
       final boolean isInitialConnection,
       final JdbcCallable<Connection, SQLException> connectFunc)
       throws SQLException {
+    return this.connectInternal(
+        driverProtocol, hostSpec, props, isInitialConnection, false, connectFunc);
+  }
+
+  protected Connection connectInternal(
+      final String driverProtocol,
```
aaron-congo: It doesn't look like this is used, can we remove it from the parameter list?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
aaron-congo: Also, not sure if this is related, but on one of my branches I added an "isDialectConfirmed" method to PluginService. If we used that, would we be able to get rid of this parameter? It looks like part of the purpose of the parameter may be related to checking whether the dialect has been confirmed or not

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -85,6 +88,7 @@ public class BlueGreenStatusProvider {
       "bgSwitchoverTimeoutMs", "180000", // 3min
       "Blue/Green Deployment switchover timeout (in msec).");
 
+  // TODO: deprecate it
```
aaron-congo: Should this comment be removed now that the docs have been updated, or is this a reminder for the future to remove the field?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -509,55 +521,26 @@ protected void updateDnsFlags(BlueGreenRole role, BlueGreenInterimStatus interim
...
-    result = this.getValueHash(result,
-        interimStatus.currentIpAddressesByHostMap == null
-            ? ""
-            : interimStatus.currentIpAddressesByHostMap.entrySet().stream()
-                .map(x -> x.getKey() + x.getValue())
-                .sorted(Comparator.comparing(x -> x))
-                .collect(Collectors.joining(",")));
-    return result;
+  protected void resetMonitors(final AtomicBoolean monitorResetCompleted, final String eventName) {
+    if (monitorResetCompleted.compareAndSet(false, true)) {
```
aaron-congo, line 521: Here we are setting monitorResetCompleted before the reset is actually performed below, should we rename to monitorResetRequested, or check the value here but update it after the block below is finished?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -901,7 +875,60 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
...
+                : this.hostSpecBuilder.host(greenHost).port(interimStatus.port).build();
+
+            Optional<String> greenIp = this.hostIpAddresses.get(greenHost);
+            HostSpec greenHostSpecWithIp = greenIp == null || !greenIp.isPresent()
+                ? this.hostSpecBuilder.host(greenHost).build()
+                : (interimStatus == null
+                    ? this.hostSpecBuilder.host(greenIp.get()).build()
+                    : this.hostSpecBuilder.host(greenIp.get()).port(interimStatus.port).build());
+
+            // Check whether green host is already been connected with blue (no-prefixes) IAM host name.
```
aaron-congo:
```suggestion
            // Check if the green host has connected to the blue (no-prefixes) IAM host name.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -901,7 +875,60 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
...
+            List<HostSpec> iamHosts;
+            if (this.isAlreadySuccessfullyConnected(greenHost, greenHost)) {
```
aaron-congo: I don't understand this code to clearly so just wanted to double check that it makes sense to pass in greenHost twice here? The comment above mentions checking if green has been connected to blue so I expected one of these to be a blue host

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -901,7 +875,60 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
...
+              // Green node has already changed its name, and it's not a new blue node (no prefixes).
```
aaron-congo: A bit confused, when green nodes change their name they end up with no prefixes, but here you're saying the new name is not a no-prefix URL?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -901,7 +875,60 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
...
+              iamHosts = Collections.singletonList(blueHostSpec);
+            } else {
+              // Green node isn't yet changed its name, so we need to try both possible IAM host options.
```
aaron-congo:
```suggestion
              // Green node has not changed its name, so we need to try both possible IAM host options.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -901,7 +875,60 @@ protected void createPostRouting(List<ConnectRouting> connectRouting) {
...
+                  this.getHostAndPort(greenHost, interimStatus.port),
+                  x.getValue(),
+                  greenHostSpecWithIp,
+                  iamHosts,
+                  isGreenHostInstance ? (iamHost) -> this.registerIamHost(greenHost, iamHost) : null));
+            }
+          });
+
+    } else if (!this.greenDnsRemoved) {
+      // Green topology is already changed.
```
aaron-congo:
```suggestion
      // Green topology has already changed.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/HostMonitorImpl.java
```diff
@@ -178,8 +179,8 @@ public void run() {
           }
 
           if (!this.activeContexts.isEmpty()
-              || this.monitoringConn == null
-              || this.monitoringConn.isClosed()) {
+              || this.monitoringConn.get() == null
+              || this.monitoringConn.get().isClosed()) {
```
aaron-congo: I know the odds of this are low but I think that monitoringConn could become null between the first and second line, which would produce an NPE, should we introduce a local variable above to store the result of monitoringConn.get()?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/HostMonitorImpl.java
```diff
@@ -319,7 +314,7 @@ ConnectionStatus checkConnectionStatus(final long shortestFailureDetectionInterv
 
     long startNano = this.getCurrentTimeNano();
     try {
-      if (this.monitoringConn == null || this.monitoringConn.isClosed()) {
+      if (this.monitoringConn.get() == null || this.monitoringConn.get().isClosed()) {
```
aaron-congo: Same as above

wrapper/src/main/java/software/amazon/jdbc/plugin/efm/HostMonitorImpl.java
```diff
@@ -336,15 +331,15 @@ ConnectionStatus checkConnectionStatus(final long shortestFailureDetectionInterv
...
+        this.monitoringConn.set(this.pluginService.forceConnect(this.hostSpec, monitoringConnProperties));
+        LOGGER.finest(() -> "Opened monitoring connection: " + this.monitoringConn.get());
         return new ConnectionStatus(true, this.getCurrentTimeNano() - startNano);
       }
 
       startNano = this.getCurrentTimeNano();
       // Some drivers, like MySQL Connector/J, execute isValid() in a double of specified timeout time.
       // TODO: fix me. Need to find a better solution to double timeout issue.
-      final boolean isValid = this.monitoringConn.isValid(
+      final boolean isValid = this.monitoringConn.get().isValid(
```
aaron-congo: Similarly, do we need a local variable to guard against the case where monitoringConn.get returns null here?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorImpl.java
```diff
@@ -300,7 +317,7 @@ boolean checkConnectionStatus() {
     }
 
     try {
-      if (this.monitoringConn == null || this.monitoringConn.isClosed()) {
+      if (this.monitoringConn.get() == null || this.monitoringConn.get().isClosed()) {
```
aaron-congo: same as above, do we need a local variable here in case monitoringConn becomes null between the first and second statement?

wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorImpl.java
```diff
@@ -315,17 +332,17 @@ boolean checkConnectionStatus() {
...
+        LOGGER.finest(() -> "Opened monitoring connection: " + this.monitoringConn.get());
         return true;
       }
 
       // Some drivers, like MySQL Connector/J, execute isValid() in a double of specified timeout time.
       final int validTimeout = (int) TimeUnit.NANOSECONDS.toSeconds(
           this.failureDetectionIntervalNano - THREAD_SLEEP_NANO) / 2;
       // validTimeout could get rounded down to 0.
-      return this.monitoringConn.isValid(Math.max(MIN_VALIDITY_CHECK_TIMEOUT_SEC, validTimeout));
+      return this.monitoringConn.get().isValid(Math.max(MIN_VALIDITY_CHECK_TIMEOUT_SEC, validTimeout));
```
aaron-congo: same as above, do we need a local variable here?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover2/FailoverConnectionPlugin.java
```diff
@@ -19,7 +19,9 @@
 import static software.amazon.jdbc.plugin.failover.FailoverMode.STRICT_READER;
 
 import java.sql.Connection;
+import java.sql.ResultSet;
```
aaron-congo: Can we remove these unused imports?

wrapper/src/main/java/software/amazon/jdbc/plugin/strategy/fastestresponse/NodeResponseTimeMonitor.java
```diff
@@ -122,7 +128,7 @@ public void monitor() {
               break;
             }
             long startTime = this.getCurrentTime();
-            if (this.pluginService.getTargetDriverDialect().ping(this.monitoringConn)) {
+            if (this.pluginService.getTargetDriverDialect().ping(this.monitoringConn.get())) {
```
aaron-congo: same as above, do we need a local variable here in case monitoringConn becomes null between the check above and the usage here?

wrapper/src/main/java/software/amazon/jdbc/plugin/strategy/fastestresponse/NodeResponseTimeMonitor.java
```diff
@@ -176,7 +176,7 @@ public void monitor() {
 
   private void openConnection() {
     try {
-      if (this.monitoringConn == null || this.monitoringConn.isClosed()) {
+      if (this.monitoringConn.get() == null || this.monitoringConn.get().isClosed()) {
```
aaron-congo: same as above, do we need a local variable here?

wrapper/src/main/java/software/amazon/jdbc/util/events/Event.java
```diff
@@ -20,4 +20,5 @@
  * A marker interface for events that need to be communicated between different components.
```
aaron-congo:
```suggestion
 * An interface for events that need to be communicated between different components.
```

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -42,6 +43,7 @@
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
```
karenc-bq: Is this the correct annotation import?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
sergiyvamz: `monitorResetCompleted` is a flag that helps to reset monitor just once. I think it doesn't matter where to set it up, before actual reset or after.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
sergiyvamz: `pluginToSkip` is used to skip calling the current plugin when the plugin initiates opening a new connection. For instance, `failover2` plugin needs to open a new connection and calls `connect()` on `PluginService`. This call goes through plugin chain and eventually calls `connect()` method on the same plugin instance. That can make `connect()` processing  challenging since the plugin (probably) needs to skip it and just to pass it through. In other words, the plugin opens a new connection and doesn't want to be called again for this call.

For BG plugin, the case is completely different. After BG plugin is initialized, there're 3 topology monitors get started. 
- Topology monitor that's associated to the client application. This monitor connects to blue cluster and eventually reconnects to green cluster. In other words all `forceConnect()` calls from this monitor instance should follow B/G routing.
- Topology monitor that's opened specifically for the purpose of monitoring blue cluster. This monitor should always stay with blue cluster (that eventually becomes an old blue cluster). Thus, this monitor should not follow B/G routing when `forceConnect()` is called from this monitor.
- Topology monitors for green cluster monitoring. This monitor is always connected to green cluster that eventually becomes a new blue cluster. This monitor should not follow B/G routing while calling `forceConnect()`.

To sum up, there're 3 instances of topology monitor (which instantiated from the same monitor class). The only way to change their behaviour is to use different monitoring connection properties. The property mentioned above helps to skip B/G routing in `forceConnect()` method.

sergiyvamz: It seems that `isDialectConfirmed` is a different case and is not related to B/G routing.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
```diff
@@ -124,6 +128,23 @@ public Set<String> getSubscribedMethods() {
...
+  @Override
+  public Connection forceConnect(
+      final String driverProtocol,
+      final HostSpec hostSpec,
+      final Properties props,
+      final boolean isInitialConnection,
+      final JdbcCallable<Connection, SQLException> connectFunc)
+      throws SQLException {
+
+    if (props.containsKey(BG_SKIP_ROUTING_IN_FORCE_CONNECT)) {
```
aaron-congo, line 140: What if props contains the key but the value is set to false?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
aaron-congo: Okay gotcha, I see the difference in usage, but I believe we can still use pluginToSkip, which I would prefer if possible because it feels less hacky. If I understand correctly, the first topology monitor you mention is ClusterTopologyMonitorImpl and the second/third you mention are BlueGreenStatusMonitor's. Looking at the code, this property is only set for the BlueGreenStatusMonitor's and is only used to skip the BG plugin. The ClusterTopologyMonitor doesn't use this property. So would it work to just pass the BG plugin as the pluginToSkip in all the BlueGreenStatusMonitor forceConnect calls? I believe this would have the same effect as the current implementation: with both implementations, the BG plugin will be skipped when BlueGreenStatusMonitor calls forceConnect but not when ClusterTopologyMonitorImpl calls forceConnect. The BG plugin can be grabbed using PluginService#getPlugin(BlueGreenConnectionPlugin.class) so that it can be passed as pluginToSkip

sergiyvamz: All 3 monitors are ClusterTopologyMonitorImpl.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenConnectionPlugin.java
sergiyvamz: It's internal flag and we can control the logic behind it. The current contract is a simple property presence. The value is ignored.

aaron-congo (approved):

PR #1609, opened 2025-11-22
feat: failover read only connections

sergiyvamz:
### Summary

Automatically trigger writer failover when connections are incorrectly established as read-only connections

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@sophia-bq 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/exceptions/AbstractPgExceptionHandler.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/exceptions/AbstractPgExceptionHandler.java
@@ -23,6 +23,9 @@ import software.amazon.jdbc.targetdriverdialect.TargetDriverDialect;
 import software.amazon.jdbc.util.StringUtils;
 
 public abstract class AbstractPgExceptionHandler implements ExceptionHandler {
+
+  protected static final String READ_ONLY_CONNECTION_SQLSTATE = "25006";
+
   public abstract List<String> getNetworkErrors();
 
   public abstract List<String> getAccessErrors();
@@ -99,7 +102,7 @@ public abstract class AbstractPgExceptionHandler implements ExceptionHandler {
   @Override
   public boolean isReadOnlyConnectionException(
       final @Nullable String sqlState, final @Nullable Integer errorCode) {
-    return "25006".equals(sqlState);
+    return READ_ONLY_CONNECTION_SQLSTATE.equals(sqlState);
   }
 
   @Override
--- a/wrapper/src/main/java/software/amazon/jdbc/exceptions/MySQLExceptionHandler.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/exceptions/MySQLExceptionHandler.java
@@ -17,6 +17,9 @@
 package software.amazon.jdbc.exceptions;
 
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.jdbc.targetdriverdialect.TargetDriverDialect;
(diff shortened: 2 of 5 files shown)


wrapper/src/main/java/software/amazon/jdbc/exceptions/AbstractPgExceptionHandler.java
```diff
@@ -95,4 +95,36 @@ public boolean isLoginException(final String sqlState) {
...
+      final Throwable throwable, @Nullable TargetDriverDialect targetDriverDialect) {
+
+    Throwable exception = throwable;
+
+    while (exception != null) {
+      String sqlState = null;
+      Integer errorCode = null;
+      if (exception instanceof SQLException) {
+        sqlState = ((SQLException) exception).getSQLState();
+        errorCode = ((SQLException) exception).getErrorCode();
```
sophia-bq, line 119: Is there a reason for gathering the error code? Only the state is checked

wrapper/src/main/java/software/amazon/jdbc/exceptions/AbstractPgExceptionHandler.java
```diff
@@ -95,4 +95,36 @@ public boolean isLoginException(final String sqlState) {
     }
     return getAccessErrors().contains(sqlState);
   }
+
+  @Override
+  public boolean isReadOnlyConnectionException(
+      final @Nullable String sqlState, final @Nullable Integer errorCode) {
+    return "25006".equals(sqlState);
```
sophia-bq: Perhaps "25006" would be better defined in a constant with a clear name. Otherwise a comment here that captures the sql state's meaning something like "read only sql transaction error" would be nice

wrapper/src/main/java/software/amazon/jdbc/exceptions/MySQLExceptionHandler.java
```diff
@@ -103,6 +103,38 @@ public boolean isLoginException(final String sqlState) {
     return SQLSTATE_ACCESS_ERROR.equals(sqlState);
   }
 
+  @Override
+  public boolean isReadOnlyConnectionException(
+      final @Nullable String sqlState, final @Nullable Integer errorCode) {
+    return "HY000".equals(sqlState) && errorCode != null && (errorCode == 1290 || errorCode == 1836);
```
sophia-bq: Similar comment here - a bit of explanation of these codes might be nice

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/FailoverConnectionPlugin.java
```diff
@@ -890,7 +890,12 @@ protected boolean shouldExceptionTriggerConnectionSwitch(final Throwable t) {
       return false;
     }
 
-    return this.pluginService.isNetworkException(t, this.pluginService.getTargetDriverDialect());
+    if (this.pluginService.isNetworkException(t, this.pluginService.getTargetDriverDialect())) {
+      return true;
+    }
+
+    return this.failoverMode == FailoverMode.STRICT_WRITER
```
sophia-bq, line 899: Nit: a comment explain this section of the failover plugins could improve readability

sophia-bq: Also should there be unit tests added for this change?

sophia-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/exceptions/AbstractPgExceptionHandler.java
sergiyvamz: Mysql exception handler uses errorCode.

PR #1616, opened 2025-11-26
fix: wait timeout for custom endpoint info

aaron-congo:
### Description

- Before these changes, the custom endpoint plugin would throw an exception after 5 minutes when the custom endpoint info expired. This occurred because the main thread waits 5 seconds for the info to be replaced, but the monitor thread waits ~30s between iterations
- This PR fixes this problem by introducing a synchronizing lock object and changing the monitor thread's wait to use `Object.wait` instead of `TimeUnit.sleep`. Now, when a request for custom endpoint comes in but the info does not exist or is expired, the monitoring thread will be woken up via `Object.notifyAll`

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
@@ -218,9 +218,11 @@ public class CustomEndpointMonitorImpl extends AbstractMonitor implements Custom
 
   protected void sleep(long durationNano) throws InterruptedException {
     long endNano = System.nanoTime() + durationNano;
+    // Choose the minimum between 500ms and the durationNano passed in, in case durationNano is less than 500ms.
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
```diff
@@ -213,8 +216,24 @@ public void monitor() {
     }
   }
 
+  protected void sleep(long durationNs) throws InterruptedException {
+    synchronized (this.refreshRequired) {
+      this.refreshRequired.wait(TimeUnit.NANOSECONDS.toMillis(durationNs));
```
sergiyvamz: This `wait()` may return earlier than specified time when the value of `this.refreshRequired` hasn't changed. You need a loop to continue waiting. `ClusterTopologyMonitorImpl` has good implementation of `delay()` with a loop.

aaron-congo: Fixed now

PR #1621, opened 2025-12-01
fix error with "too many connections"

sergiyvamz:
slow down node monitoring thread for non-transient exceptions

### Summary

Handle various transient and non-transient exceptions for topology monitor. Inspired by
https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1620

### Additional Reviewers

@karenc-bq 
@aaron-congo 
@sophia-bq 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/exceptions/PgExceptionHandler.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/exceptions/PgExceptionHandler.java
@@ -23,7 +23,6 @@ public class PgExceptionHandler extends AbstractPgExceptionHandler {
 
   // The following SQL States for Postgresql are considered as "communication" errors
   private static final List<String> NETWORK_ERRORS = Arrays.asList(
-      "53", // insufficient resources
       "57P01", // admin shutdown
       "57P02", // crash shutdown
       "57P03", // cannot connect now


wrapper/src/main/java/software/amazon/jdbc/exceptions/MySQLExceptionHandler.java
```diff
@@ -71,7 +71,8 @@ public boolean isNetworkException(final String sqlState) {
       return false;
     }
 
-    return sqlState.startsWith("08");
+    // 08004 -  "Server rejected the connection" or "Connection refused by server"
+    return sqlState.startsWith("08") && !"08004".equals(sqlState);
```
aaron-congo, line 75: I think the result of this method helps us determine whether we should try failover, do we want to skip failover if the state is 08004? 

Also, does PgExceptionHandler need a similar change?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -1037,5 +1058,11 @@ private void readerThreadFetchTopology(final Connection connection, final @Nulla
         LOGGER.fine(Utils.logTopology(hosts));
       }
     }
+
+    private long calculateBackoffWithJitter(int attempt) {
+      long backoff = INITIAL_BACKOFF_MS * Math.round(Math.pow(2, Math.min(attempt, 6)));
+      backoff = Math.min(backoff, MAX_BACKOFF_MS);
+      return Math.round(backoff * (0.5 + random.nextDouble() * 0.5));
```
aaron-congo, line 1087: for my own knowledge, what is the benefit of using a random factor in this calculation?

wrapper/src/main/java/software/amazon/jdbc/exceptions/MySQLExceptionHandler.java
sergiyvamz: Good call. For PG error 53xxx are related to insufficient resources like dick full, max number of connections, etc. I removed this code from the list of network related errors.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
sergiyvamz: It gives some variety in delaying a next action and helps to reduce race conditions. It's a well known techniques to repeat (re-broadcast) data over shared media (like cable or air). If two senders send data at the same time, transmission fails most probably (since both data pieces "corrupt" each other). If re-send occur by both senders after same delay time, the chance of another collision increases. Jitter helps to low such probability.

aaron-congo (approved):

PR #1624, opened 2025-12-02
bug: possible leaked connections

sergiyvamz:

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
@@ -253,11 +253,10 @@ public class ClusterTopologyMonitorImpl extends AbstractMonitor implements Clust
   public void stop() {
     this.nodeThreadsStop.set(true);
 
+    this.closeNodeMonitors();
     this.closeConnection(this.nodeThreadsWriterConnection);
     this.closeConnection(this.nodeThreadsReaderConnection);
     this.closeConnection(this.monitoringConnection);
-
-    this.closeNodeMonitors();
 
     // It breaks a waiting/sleeping cycles in monitoring thread
     synchronized (this.requestToUpdateTopology) {


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -513,6 +519,10 @@ protected void shutdownNodeExecutorService() {
           return;
         }
 
+        // Close atomic reference connections before shutdown
```
aaron-congo: Can we move this to the monitor() finally block?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -557,7 +567,7 @@ protected List<HostSpec> openAnyConnectionAndUpdateTopology() {
     boolean writerVerifiedByThisThread = false;
     if (this.monitoringConnection.get() == null) {
 
-      Connection conn;
+      Connection conn = null;
```
aaron-congo: nit: my IDE tells me setting it to null here is redundant

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -572,6 +582,9 @@ protected List<HostSpec> openAnyConnectionAndUpdateTopology() {
             "ClusterTopologyMonitorImpl.openedMonitoringConnection",
             new Object[]{this.initialHostSpec.getHost()}));
 
+        // Clear temp reference to prevent double-close
+        conn = null;
```
aaron-congo: This variable isn't used anywhere after this assignment, so setting it to null doesn't have an effect

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
sergiyvamz: `shutdownNodeExecutorService()` is already in `monitor()` finally block.

aaron-congo: yeah I noticed that, as a nit, can we rename this to `closeNodeMonitors` or `stopNodeMonitors`? when I read `shutdownNodeExecutorService` I expect that we only interact with the executor service itself, and not any of the related variables like nodeThreadsWriterConnection

aaron-congo: Also, we are performing this before the node monitors have fully stopped. Is there a chance that between closing the connection and the node monitors fully stopping, a new nodeThreadsWriterConnection is opened? Should we move these closeConnection calls to after nodeExecutorService.shutdown()?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitorImpl.java
```diff
@@ -503,7 +505,20 @@ public void processEvent(Event event) {
     }
   }
 
-  protected void shutdownNodeExecutorService() {
+  protected void nodeThreadConnectionCleanUp() {
+    if (this.monitoringConnection.get() != this.nodeThreadsWriterConnection.get()) {
+      this.closeConnection(this.nodeThreadsWriterConnection);
+    } else {
+      this.nodeThreadsWriterConnection.set(null);
+    }
```
karenc-bq, line 513: Do we want to set reader/writer connection to null after closing the connection?
Does it make sense to move this outside of the else statement

sergiyvamz: Connection sets null in both case. At line 512 or inside `closeConnection()`.

PR #1625, opened 2025-12-02
remove finalizer and use LazyCleaner implementation from pgjdbc

davecramer:
### Summary

see $subject

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
@@ -76,7 +76,7 @@ public class ConnectionWrapper implements Connection, CanReleaseResources {
 
     this.pluginManager = servicesContainer.getConnectionPluginManager();
     this.pluginService = servicesContainer.getPluginService();
-    this.cleanable = LazyCleaner.getInstance().register(this, new CleanupAction(openConnectionStacktrace, pluginService));
+    this.cleanable = LazyCleaner.getInstance().register(this, new CleanupAction(openConnectionStacktrace, pluginManager, pluginService));
     this.hostListProviderService = servicesContainer.getHostListProviderService();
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/LazyCleaner.java
```diff
@@ -0,0 +1,212 @@
...
+    watchedCount++;
+
+    if (!threadRunning) {
+      threadRunning = startThread();
+    }
+    return node;
+  }
+
+  private boolean startThread() {
+    Thread thread = threadFactory.newThread(new Runnable() {
```
sergiyvamz, line 118: Should the thread be a daemon thread?

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
```diff
@@ -126,9 +130,6 @@ protected void init(final Properties props) throws SQLException {
 
   public void releaseResources() {
     this.pluginManager.releaseResources();
-    if (this.pluginService instanceof CanReleaseResources) {
```
sergiyvamz, line 129: Should we clean up pluginManager similar way?

wrapper/src/main/java/software/amazon/jdbc/util/LazyCleaner.java
davecramer: Actually we do not use a daemon thread in pgjdbc

wrapper/src/main/java/software/amazon/jdbc/wrapper/ConnectionWrapper.java
davecramer: we would have to pass pluginManager into LazyCleaner, but sure

davecramer:
closing in favour of #1641

PR #1626, opened 2025-12-03
fix: add expiration time for cached reader connections and refactor BG plugin

karenc-bq:
### Summary

Fix
- add expiration time for cached reader connections

Chore
- refactor bg plugin

### Description

If connection pool isn't used, the read/write splitting plugin caches the reader connection forever, all read operations will be redirected to a single instance when setReadOnly(true) is called. This may impede performance.
This PR provides a connection parameter allowing user applications to redirect read operations to other instances after certain expiration time.

This PR also reuses existing methods in the BG plugin to reduce code duplication

### DIFF
--- a/wrapper/build.gradle.kts
+++ b/wrapper/build.gradle.kts
@@ -283,8 +283,21 @@ tasks.shadowJar {
 
     archiveBaseName.set("aws-advanced-jdbc-wrapper")
     archiveClassifier.set("bundle-federated-auth")
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 
-    mergeServiceFiles("META-INF")
+    from("${layout.buildDirectory.get()}/META-INF/services/") {
+        into("META-INF/services/")
+    }
+
+    mergeServiceFiles()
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
```diff
@@ -546,16 +546,15 @@ protected void openConnection() {
 
     this.openConnectionFuture = openConnectionExecutorService.submit(() -> {
 
-      HostSpec connectionHostSpecCopy = this.connectionHostSpec.get();
-      String connectedIpAddressCopy = this.connectedIpAddress.get();
-
-      if (connectionHostSpecCopy == null) {
+      if (this.connectionHostSpec == null) {
```
karenc-bq:
```suggestion
      if (this.connectionHostSpec.get() == null) {
```

sergiyvamz (approved):

PR #1627, opened 2025-12-04
[2.6.x] chore: update tests and docs to use MariaDB 2.x 

aaron-congo:
MariaDB 3.x should not be used against Aurora because it uses pipelining, which is not compatible with Aurora. This aligns with their documentation (see [here](https://mariadb.com/docs/connectors/mariadb-connector-j/about-mariadb-connector-j#aurora)). Additionally, these properties should also be set, because they also use pipelining.

```java
props.setProperty("usePipelineAuth", "false");
props.setProperty("useBatchMultiSend", "false");
```

This PR updates docs to clarify that MariaDB 2.x should be used with these properties. It also updates the tests to use MariaDB 2.x.

Relates to #1596 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointMonitorImpl.java
@@ -20,6 +20,7 @@ import static software.amazon.jdbc.plugin.customendpoint.MemberListType.STATIC_L
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


sergiyvamz:
Closing this in favor of two docs-only PRs: #2145 (`main`) and #2146 (`main-2.x`).

Thanks for chasing the 2.7.13 fix, @aaron-congo, and apologies for the long silence, @luneo7. Here is where we landed after re-checking the evidence.

**MariaDB Connector/J 3.x does work against Aurora in our own test matrix.** The nightly Aurora MySQL suite on `main` runs against 3.5.10 in four sharded `mysql-aurora-mariadb-driver` jobs, all green on 2026-09-19. On `main-2.x`, `test-all-mysql-aurora` does not set `test-no-mariadb-driver`, so it covers the MariaDB driver too, and it was green at 3.5.6 (run 20276803447). The 2.7.12 downgrade on this branch is what turned the standard container suite red.

**Pipelining is not in play against Aurora MySQL.** Tracing 3.5.10: `ConnectionHelper.initializeClientCapabilities` returns `capabilities & serverCapabilities`, and `STMT_BULK_OPERATIONS` is a MariaDB-server capability that Aurora MySQL never advertises. Both `BaseContext.permitPipeline()` and `canUseBulk` in `ClientPreparedStatement` require that bit, so neither pipelining nor `COM_STMT_BULK` is used there.

**There is a 3.x switch after all.** `disablePipeline=true` exists from Connector/J 3.1 onward (absent in 3.0.10, present in 3.1.1 and 3.5.10). It is missing from MariaDB's published option reference, which is probably why #1596 concluded that no 3.x equivalent existed.

**The code half would be a breaking change on this branch.** Renaming `TargetDriverDialectCodes.MARIADB_CONNECTOR_J_VER_3` to `_VER_2` changes the accepted value of the user-facing `wrapperTargetDriverDialect` property, and `TargetDriverDialectManager.getDialect` throws `SQLException(unknownDialectCode)` for an unrecognized code, so anyone setting `mariadb-connector-j-3` today would break going from 2.6.8 to 2.6.9. Per Maintenance.md the 2.x line takes bug fixes and security patches only. Since a single `MariadbTargetDriverDialect` serves both majors unchanged, the dialect code is only a label.

The docs half is preserved and corrected. #1984 already shipped it to `main`, but it kept `usePipelineAuth` and `useBatchMultiSend`, both removed in Connector/J 3.0 and silently ignored there, so the published advice currently has no effect. #2145 replaces it with version-scoped guidance, and #2146 brings the same section to the 2.6.x line.

Happy to reopen if there is a concrete 3.x failure against Aurora that we have not reproduced. An operation and an error message would be enough to go on.

PR #1633, opened 2025-12-09
refactor: use the ServicesContainer PluginFactory method by default

aaron-congo:
Removes the old PluginFactory method that accepts a PluginService argument


wrapper/src/main/java/software/amazon/jdbc/PluginFactory.java
```diff
@@ -20,14 +20,10 @@
...
- * A factory for plugins that utilizes a {@link FullServicesContainer}. This interface extends
- * {@link ConnectionPluginFactory} to provide additional flexibility in plugin instantiation while maintaining backward
- * compatibility.
- *
- * <p>Implementations of this interface can access all services in the {@link FullServicesContainer} when creating
- * connection plugins, rather than being limited to just the {@link PluginService}</p>
+ * Interface for plugin factories. This class implements ways to initialize a plugin.
  */
-public interface ServicesContainerPluginFactory extends ConnectionPluginFactory {
+public interface PluginFactory {
```
sergiyvamz: What's the reason to change interface name?

aaron-congo: Fixed now

PR #1636, opened 2025-12-10
refactor: use monitoring RdsHostListProvider by default

aaron-congo:
This PR removes the old non-monitoring RdsHostListProvider so that the default HostListProvider is a monitoring one.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorService.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorService.java
@@ -30,10 +30,7 @@ public interface HostMonitorService {
   HostMonitorConnectionContext startMonitoring(
       Connection connectionToAbort,
       HostSpec hostSpec,
-      Properties properties,
-      int failureDetectionTimeMillis,
-      int failureDetectionIntervalMillis,
-      int failureDetectionCount) throws SQLException;
+      Properties properties) throws SQLException;
 
   /**
    * Stop monitoring for a connection represented by the given {@link HostMonitorConnectionContext}.
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorServiceImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorServiceImpl.java
@@ -16,6 +16,10 @@
 
 package software.amazon.jdbc.plugin.efm2;
 
+import static software.amazon.jdbc.plugin.efm2.HostMonitoringConnectionPlugin.FAILURE_DETECTION_COUNT;
+import static software.amazon.jdbc.plugin.efm2.HostMonitoringConnectionPlugin.FAILURE_DETECTION_INTERVAL;
+import static software.amazon.jdbc.plugin.efm2.HostMonitoringConnectionPlugin.FAILURE_DETECTION_TIME;
+
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Properties;
@@ -54,13 +58,22 @@ public class HostMonitorServiceImpl implements HostMonitorService {
   protected final MonitorService coreMonitorService;
   protected final TelemetryFactory telemetryFactory;
   protected final TelemetryCounter abortedConnectionsCounter;
+  protected final int failureDetectionTimeMillis;
+  protected final int failureDetectionIntervalMillis;
+  protected final int failureDetectionCount;
 
-  public HostMonitorServiceImpl(final @NonNull FullServicesContainer serviceContainer, Properties props) {
+  protected HostMonitorKey monitorKey;
+
+  public HostMonitorServiceImpl(final @NonNull FullServicesContainer serviceContainer, final Properties props) {
     this.serviceContainer = serviceContainer;
     this.coreMonitorService = serviceContainer.getMonitorService();
     this.pluginService = serviceContainer.getPluginService();
     this.telemetryFactory = serviceContainer.getTelemetryFactory();
     this.abortedConnectionsCounter = telemetryFactory.createCounter("efm2.connections.aborted");
+
+    this.failureDetectionTimeMillis = FAILURE_DETECTION_TIME.getInteger(props);
+    this.failureDetectionIntervalMillis = FAILURE_DETECTION_INTERVAL.getInteger(props);
+    this.failureDetectionCount = FAILURE_DETECTION_COUNT.getInteger(props);
 
     this.coreMonitorService.registerMonitorTypeIfAbsent(
         HostMonitorImpl.class,
@@ -78,21 +91,10 @@ public class HostMonitorServiceImpl implements HostMonitorService {
   public HostMonitorConnectionContext startMonitoring(
       final Connection connectionToAbort,
       final HostSpec hostSpec,
-      final Properties properties,
-      final int failureDetectionTimeMillis,
-      final int failureDetectionIntervalMillis,
-      final int failureDetectionCount) throws SQLException {
-
-    final HostMonitor monitor = this.getMonitor(
-        hostSpec,
-        properties,
-        failureDetectionTimeMillis,
-        failureDetectionIntervalMillis,
-        failureDetectionCount);
-
+      final Properties properties) throws SQLException {
+    final HostMonitor monitor = this.getMonitor(hostSpec, properties);
     final HostMonitorConnectionContext context = new HostMonitorConnectionContext(connectionToAbort);
     monitor.startMonitoring(context);
-
     return context;
   }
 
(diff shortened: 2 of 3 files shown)


wrapper/src/test/java/integration/container/tests/FailoverTest.java
```diff
@@ -571,8 +571,6 @@ public void test_failFromWriter() throws SQLException {
   @TestTemplate
   @EnableOnTestFeature(TestEnvironmentFeatures.NETWORK_OUTAGES_ENABLED)
   @EnableOnNumOfInstances(min = 2)
-  // Multi-AZ tests already simulate this in other tests instead of sending server failover requests.
-  @EnableOnDatabaseEngineDeployment(DatabaseEngineDeployment.AURORA)
```
aaron-congo, line 575: This test passes on multi-AZ now

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -157,31 +147,59 @@ protected void initSettings() throws SQLException {
...
-    init();
-
+  protected FetchTopologyResult getTopology(final boolean forceUpdate) throws SQLException {
     final List<HostSpec> storedHosts = this.getStoredTopology();
     if (storedHosts == null || forceUpdate) {
       // We need to re-fetch topology.
-      if (conn == null) {
-        // We cannot fetch the latest topology since we do not have access to a connection, so we return the original
-        // hosts parsed from the connection string.
+      if (!this.pluginService.isDialectConfirmed()) {
```
aaron-congo, line 217: This is needed because without it, cluster topology monitors are created before the db dialect has been updated. This causes errors because the queries that the topology monitors execute are not aligned with the actual database dialect

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -157,31 +147,59 @@ protected void initSettings() throws SQLException {
...
+            this.refreshRateNano,
+            this.highRefreshRateNano));
+  }
+
+  protected List<HostSpec> queryForTopology() throws SQLException {
+    ClusterTopologyMonitor monitor = this.servicesContainer.getMonitorService()
+        .get(ClusterTopologyMonitorImpl.class, this.clusterId);
+    if (monitor == null) {
+      monitor = this.initMonitor();
+    }
```
sergiyvamz: Can we simplify it to this?

`ClusterTopologyMonitor monitor = this.initMonitor();`

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ConnectionStringHostListProvider.java
```diff
@@ -51,26 +51,21 @@ public class ConnectionStringHostListProvider implements StaticHostListProvider
...
-      final @NonNull ConnectionUrlParser connectionUrlParser) {
+      final @NonNull ConnectionUrlParser connectionUrlParser) throws SQLException {
 
     this.isSingleWriterConnectionString = SINGLE_WRITER_CONNECTION_STRING.getBoolean(properties);
     this.initialUrl = initialUrl;
     this.connectionUrlParser = connectionUrlParser;
     this.hostListProviderService = hostListProviderService;
-  }
 
-  private void init() throws SQLException {
```
sergiyvamz, line 70: Why postponed initialization isn't good?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/DynamicHostListProvider.java
```diff
@@ -16,7 +16,13 @@
...
 
-// A marker interface for providers that can fetch a host list reflecting the current database topology.
+import software.amazon.jdbc.HostSpec;
+
+// An interface for providers that can fetch a host list reflecting the current database topology.
 // Examples include providers for Aurora or Multi-AZ clusters, where the cluster topology, status, and instance roles
 // change over time.
-public interface DynamicHostListProvider extends HostListProvider { }
+public interface DynamicHostListProvider extends HostListProvider {
+  TopologyUtils getTopologyUtils();
```
sergiyvamz: Does it make sense to replace these two methods with a single one? StaticHostListProvider can also implement this method.

`List<HostSpec> queryForTopology(conn, initialHostSpec)`

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
sergiyvamz: ClusterTopologyMonitor monitor = this.initMonitor();

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/monitoring/ClusterTopologyMonitor.java
```diff
@@ -16,11 +16,9 @@
 
 package software.amazon.jdbc.hostlistprovider.monitoring;
```
sergiyvamz: It seems that monitoring package makes not much sense now. Should we relocate remaining classes to the upper package?

wrapper/src/main/java/software/amazon/jdbc/plugin/failover/ClusterAwareWriterFailoverHandler.java
```diff
@@ -71,6 +78,7 @@ public ClusterAwareWriterFailoverHandler(
       final Properties initialConnectionProps) {
     this.servicesContainer = servicesContainer;
     this.pluginService = servicesContainer.getPluginService();
+    this.hostListProvider = this.pluginService.getHostListProvider();
```
sergiyvamz: We use HLP from the main connection and it's used by handlers in separate threads. Is it ok to use plugin service and related HLP which are not multi-thread safe?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -228,43 +234,12 @@ public void clear() {
...
-
-  @Override
-  public List<HostSpec> refresh(final Connection connection) throws SQLException {
-    init();
-    final Connection currentConnection = connection != null
-        ? connection
-        : this.hostListProviderService.getCurrentConnection();
-
-    final FetchTopologyResult results = getTopology(currentConnection, false);
+    final FetchTopologyResult results = getTopology(false);
```
sergiyvamz: `getTopology(false)` always called with `false`. It seems we can rid of this parameter.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -157,31 +147,59 @@ protected void initSettings() throws SQLException {
...
+      if (!this.pluginService.isDialectConfirmed()) {
+        // We need to confirm the dialect before creating a topology monitor so that it uses the correct SQL queries.
+        // We will return the original hosts parsed from the connections string until the dialect has been confirmed.
         return new FetchTopologyResult(false, this.initialHostList);
       }
 
-      final List<HostSpec> hosts = this.queryForTopology(conn);
+      final List<HostSpec> hosts = this.queryForTopology();
       if (!Utils.isNullOrEmpty(hosts)) {
         this.servicesContainer.getStorageService().set(this.clusterId, new Topology(hosts));
```
sergiyvamz: Topology monitor should update topology cache by itself.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/DynamicHostListProvider.java
aaron-congo: Good idea. Does it make sense to have a queryForTopology method in StaticHostListProvider if it doesn't actually query anything though?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ConnectionStringHostListProvider.java
aaron-congo: What are the benefits of postponed initialization? Hypothetically it could save us a small amount of time if we instantiate a host list provider and then skip initializing it because of a dialect update. However, in my testing this does not seem to happen. The original host list provider still gets initialized because refreshHostList is called before updateDialect.

Downsides of using postponed initialization here is complexity. When bugs occur in the connection initialization code it becomes hard to determine when/where the host list provider is initialized. Our code first gets a supplier for a HostListProvider. Then at some later point it creates the HostListProvider but its not initialized yet. Then we have to determine which of the 9 methods in RdsHostListProvider was the one that initialized the host list provider. We also have to remember to call init() whenever we add a public method to the class.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -157,33 +166,64 @@ protected void initSettings() throws SQLException {
     this.rdsUrlType = rdsHelper.identifyRdsType(this.initialHostSpec.getHost());
   }
 
+  protected ClusterTopologyMonitor getOrCreateMonitor() throws SQLException {
+    ClusterTopologyMonitor monitor = this.servicesContainer.getMonitorService()
```
sergiyvamz: `monitor` is unused.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/RdsHostListProvider.java
```diff
@@ -157,33 +166,64 @@ protected void initSettings() throws SQLException {
...
+            this.highRefreshRateNano));
+  }
+
+  @Override
+  public List<HostSpec> getCurrentTopology(Connection conn, HostSpec initialHostSpec) throws SQLException {
+    init();
+    return this.topologyUtils.queryForTopology(conn, initialHostSpec, this.instanceTemplate);
+  }
+
+  protected List<HostSpec> getFreshTopology(boolean shouldVerifyWriter, long timeoutMs) throws SQLException {
```
sergiyvamz: boolean shouldVerifyWriter, long timeoutMs - not in use

aaron-congo: Good catch, thanks

sergiyvamz (approved):

PR #1639, opened 2025-12-13
fix: restart BG monitors when B/G switchover is completed

sergiyvamz:
### Summary

Restart BG monitors and associated HostList providers when B/G switchover is completed

### Additional Reviewers

@aaron-congo 
@sophia-bq 


wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -1099,6 +1094,19 @@ protected void resetContextWhenCompleted() {
 
     if (switchoverCompleted && hasActiveSwitchoverPhases) {
       LOGGER.finest(Messages.get("bgd.resetContext"));
+
+      LOGGER.info("Stopping before");
```
aaron-congo: Was this for testing purposes? Should we remove this now?

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java
```diff
@@ -1099,6 +1094,19 @@ protected void resetContextWhenCompleted() {
...
+      if (this.monitors[BlueGreenRole.SOURCE.getValue()] != null) {
+        LOGGER.info("Stopping SOURCE");
+        this.monitors[BlueGreenRole.SOURCE.getValue()].setStop(true);
```
aaron-congo: I know the odds of this are low but we will get a NPE if `this.monitors[BlueGreenRole.SOURCE.getValue()]` is changed to null between the `if` condition check and the call to `setStop`, should we use a local variable just in case?

PR #1640, opened 2025-12-15
perf: cache efm2 monitor key for better performance

aaron-congo:
- fixes a performance issue caused by frequent calculation of the EFM2 monitor key

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorServiceImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorServiceImpl.java
@@ -161,9 +161,9 @@ public class HostMonitorServiceImpl implements HostMonitorService {
             servicesContainer,
             hostSpec,
             properties,
-            failureDetectionTimeMillis,
-            failureDetectionIntervalMillis,
-            failureDetectionCount,
+            this.failureDetectionTimeMillis,
+            this.failureDetectionIntervalMillis,
+            this.failureDetectionCount,
             this.abortedConnectionsCounter));
   }
 


wrapper/src/main/java/software/amazon/jdbc/plugin/efm2/HostMonitorServiceImpl.java
```diff
@@ -131,37 +133,55 @@ public void releaseResources() {
...
-        failureDetectionTimeMillis,
-        failureDetectionIntervalMillis,
-        failureDetectionCount,
-        hostSpec.getUrl());
+      final Properties properties) throws SQLException {
+    String hostUrl = hostSpec.getUrl();
+    if (this.monitorKey == null || !hostUrl.equals(this.monitorKey.getUrl())) {
+      // The URL being monitored has changed, so we need to recalculate the monitor key.
+      this.monitorKey = new HostMonitorKey(
+          hostUrl,
```
davecramer, line 146: Why are we storing the URL 2x here ? Looking at the code, the only thing we are interested in is the port and the host, seems to me that we could store those and just compare them?

aaron-congo: The first time is so that we can check the passed in URL against the URL of the current key. These URLs could be different because of failover, read/write splitting, or the initial connection plugin selecting a new host. If a new URL appears, we need a new monitor. 

The second value is the cache key, which consists of the URL as well as the efm2 settings. We include the settings in the key so that users can have multiple monitors for the same URL but with different settings. This may be useful if some of their connections require aggressive failure detection while other connections are fine with more relaxed settings.

PR #1655, opened 2025-12-22
feat: configurable instance substitution and role verification in AuroraInitialConnectionStrategyPlugin

sergiyvamz:


wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -127,6 +140,17 @@ public Connection connect(
       return connectFunc.call();
     }
 
+    if (type == RdsUrlType.RDS_CUSTOM_CLUSTER) {
+      Connection candidateConn =
+          this.getVerifiedConnection(props, isInitialConnection, connectFunc);
+      if (candidateConn == null) {
+        // Can't get a connection. Continue with a normal workflow.
+        LOGGER.finest("Continue with normal workflow.");
```
sophia-bq: Should this be in the messages file? Also should this be logged for writer clusters as well? I believe its missing before line 160

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -346,6 +364,89 @@ private Connection getVerifiedReaderConnection(
...
+
+      candidateConn = null;
+      candidateHost = null;
+
+      try {
+        candidateHost = this.getCandidateHost(props, this.verifyOpenedConnectionType, null);
+
+        if (candidateHost == null || this.rdsUtils.isRdsClusterDns(candidateHost.getHost())) {
+
+          // Reader is not found. It seems that topology is outdated.
```
sophia-bq: The host is not necessarily a reader

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -346,6 +364,89 @@ private Connection getVerifiedReaderConnection(
...
+          candidateConn = connectFunc.call();
+          this.pluginService.forceRefreshHostList();
+          candidateHost = this.pluginService.identifyConnection(candidateConn);
```
sophia-bq: Is there no verification of this candidateHost and connection? Couldn't it be worthwhile after forceRefresh is called?

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -57,16 +57,17 @@ public WeightedRandomHostSelector(final Random random) {
 
   public HostSpec getHost(
       @NonNull List<HostSpec> hosts,
-      @NonNull HostRole role,
+      @Nullable HostRole role,
       @Nullable Properties props) throws SQLException {
 
     final Map<String, Integer> hostWeightMap =
-        this.getHostWeightPairMap(WEIGHTED_RANDOM_HOST_WEIGHT_PAIRS.getString(props));
+        this.getHostWeightPairMap(props == null ? null : WEIGHTED_RANDOM_HOST_WEIGHT_PAIRS.getString(props));
```
aaron-congo: Looking at the code, if props is null or getString returns null or an empty string, it seems we will just end up throwing the "unable to get host" exception at the very end of this method. So I think this parameter is required, in which case we should validate it at the beginning of this method and throw an exception if validation fails (eg if props is null or getString returns a null/empty string). Then the parameter passed to getHostWeightPairMap can be marked @NonNull

wrapper/src/main/java/software/amazon/jdbc/plugin/customendpoint/CustomEndpointInfo.java
```diff
@@ -78,7 +78,7 @@ public static CustomEndpointInfo fromDBClusterEndpoint(DBClusterEndpoint respons
     final List<String> members;
     final MemberListType memberListType;
 
-    if (responseEndpointInfo.hasStaticMembers()) {
+    if (responseEndpointInfo.hasStaticMembers() && !responseEndpointInfo.staticMembers().isEmpty()) {
```
aaron-congo, line 81: should we consider throwing an error if the info indicates it has static members but the member list is empty? This seems like a problematic scenario and it might be helpful to have a specific error message pop up if it is encountered

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -76,13 +85,19 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
       new AwsWrapperProperty(
           "verifyOpenedConnectionType",
           null,
-          "Force to verify an opened connection to be either a writer or a reader.");
+          "Force to verify an opened connection to be either a writer or a reader.",
```
aaron-congo:
```suggestion
          "Defines whether an opened connection should be verified to be a writer or reader, or if no role verification should be performed.",
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -76,13 +85,19 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
...
+          false,
+          new String[] {
+              "writer", "reader", "no"
```
aaron-congo: Nit: can we change to "none"? Will need to update docs as well
```suggestion
              "writer", "reader", "none"
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -127,10 +146,42 @@ public Connection connect(
       return connectFunc.call();
     }
 
+    if (isInitialConnection) {
+      if (type == RdsUrlType.RDS_WRITER_CLUSTER
+          && this.verifyOpenedConnectionType == HostRoleVerification.READER) {
+        throw new SQLException(
+            "Wrong configuration. When using an Aurora Cluster Writer Endpoint parameter "
```
aaron-congo: Can we move these messages to the messages file?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -127,10 +146,42 @@ public Connection connect(
...
+          && this.verifyOpenedConnectionType == HostRoleVerification.WRITER) {
+        throw new SQLException(
+            "Wrong configuration. When using an Aurora Cluster Reader Endpoint parameter "
+            + "'verifyOpenedConnectionType' can't be 'writer'.");
+      }
+    }
+
+    if (type == RdsUrlType.RDS_CUSTOM_CLUSTER) {
+      Connection candidateConn =
+          this.getVerifiedConnection(props, isInitialConnection, connectFunc);
```
aaron-congo: Since verification is not always performed inside these methods, can we take the "verified" substring out of "getVerifiedConnection", "getVerifiedWriterConnection", and "getVerifiedReaderConnection"?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -346,6 +409,91 @@ private Connection getVerifiedReaderConnection(
...
+          if (isInitialConnection) {
+            hostListProviderService.setInitialConnectionHostSpec(candidateHost);
+          }
+          return candidateConn;
+        }
+
+        candidateConn = this.pluginService.connect(candidateHost, props, this);
+
+        // Verify connection if requested
+        if (this.verifyOpenedConnectionType != null
```
aaron-congo: If I understand correctly, I think there is a bug here. When users set the "verifyOpenedConnectionType" property to "no",  this.verifyOpenedConnectionType will be set to NO_VERIFICATION and hostRole will be null here. In this scenario we will always enter this branch even though "no verification" was requested.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -277,11 +338,11 @@ private Connection getVerifiedReaderConnection(
 
           if (readerCandidate == null) {
             this.closeConnection(readerCandidateConn);
-            this.delay(retryDelayMs);
+            this.delay(this.retryDelayMs);
             continue;
           }
 
-          if (readerCandidate.getRole() != HostRole.READER) {
+          if (readerCandidate.getRole() != hostRole) {
```
aaron-congo: If the user set NO_VERIFICATION than hostRole will be null here and we will always enter this branch. It seems like this branch should only be entered if the role does not match the requested verification type.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -303,7 +364,9 @@ private Connection getVerifiedReaderConnection(
 
         readerCandidateConn = this.pluginService.connect(readerCandidate, props, this);
 
-        if (this.pluginService.getHostRole(readerCandidateConn) != HostRole.READER) {
+        if (hostRoleVerification != HostRoleVerification.NO_VERIFICATION
+            && this.pluginService.getHostRole(readerCandidateConn) != hostRole) {
```
aaron-congo: If hostRoleVerification is null, hostRole will be null and we will always enter this branch even though its intended for the scenario where the role does not match the requested verification type

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -17,11 +17,13 @@ To enable the Aurora Initial Connection Strategy Plugin, add `initialConnection`
...
+| `initialConnectionHostSelectorStrategy`           | String  |    No    | The strategy that will be used to select a host when opening a new connection. A host will be selected according to a host role specified in parameter `verifyOpenedConnectionType`. If host role isn't provided then any available host can be selected by the strategy.<br><br> For more information on the available reader selection strategies, see this [table](../ReaderSelectionStrategies.md).                                                                                 | `leastConnections` | `random`                                                                                                                                            |
```
aaron-congo:
```suggestion
| `initialConnectionHostSelectorStrategy`           | String  |    No    | The strategy that will be used to select a host when opening a new connection. A host will be selected according to the host role specified in parameter `verifyOpenedConnectionType`. If `verifyOpenedConnectionType` isn't provided then any available host will be selected by the strategy.<br><br> For more information on the available host selection strategies, see this [table](../ReaderSelectionStrategies.md).                                                                                 | `leastConnections` | `random`                                                                                                                                            |
```

aaron-congo: We might need to also rename "ReaderSelectionStrategies.md" to "HostSelectionStrategies.md" since they might be used when the user does not care what the host role is.

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -17,11 +17,13 @@ To enable the Aurora Initial Connection Strategy Plugin, add `initialConnection`
...
+| `verifyOpenedConnectionType`                      | String  |    No    | Set up a desired host role used by the strategy and verify opened connection host role.<br><br>When using cluster writer endpoint or global cluster endpoint possible values are `writer` and `no`. When using cluster reader endpoint possible values are `reader` and `no`. In other cases possible values are `reader`,`writer` and `no`. The value `no` defines no host role verification. Explicitly set parameter value overrides default value.                                  | `reader`           | `writer` when using cluster writer endpoint or global cluster endpoint.<br><br>`reader` when using cluster reader endpoint.<br><br>Otherwise: `no`. |
```
aaron-congo:
```suggestion
| `verifyOpenedConnectionType`                      | String  |    No    | Defines whether an opened connection should be verified to be a writer or reader, or if no role verification should be performed.<br><br>When using cluster writer endpoint or global cluster endpoint possible values are `writer` and `no`. When using cluster reader endpoint possible values are `reader` and `no`. In other cases possible values are `reader`,`writer` and `no`. The value `no` defines no host role verification. Explicitly setting the parameter overrides the default value.                                  | `reader`           | `writer` when using cluster writer endpoint or global cluster endpoint.<br><br>`reader` when using cluster reader endpoint.<br><br>Otherwise: `no`. |
```

aaron-congo: Should "writer" be allowed if using custom endpoint URLs? I don't think it makes much sense since custom endpoints either have to be of type "any" or "reader". Should we add: "When using custom cluster endpoint possible values are `reader` and `no`."

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -1,42 +1,62 @@
...
+The Aurora Initial Connection Strategy Plugin allows users to configure their initial connection strategy, and it can
+also be used to obtain a connection more reliably if DNS is updating by replacing an out of date endpoint. When the
+Aurora Initial Connection Strategy Plugin attempts to make a connection, it may retry the connection attempt if there is
+a failure. Users are able to configure how often to retry a connection and the maximum allowed time to obtain a
+connection using the connection parameters.
 
-This plugin also helps retrieve connections more reliably. When a user connects to a cluster endpoint, the actual instance for a new connection is resolved by DNS. During failover, the cluster elects another instance to be the writer. While DNS is updating, which can take up to 40-60 seconds, if a user tries to connect to the cluster endpoint, they may be connecting to an old node. This plugin helps by replacing the out of date endpoint if DNS is updating.
+When this plugin is enabled, if the initial connection is to a reader cluster endpoint, the connected reader host will
```
sergiyvamz: We use strategy for other endpoints either (like custom cluster endpoints)

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -1,42 +1,62 @@
...
...can improve the chances of connecting to the correct role. It also can be used in conjunction with `initialConnectionHostSelectorStrategy` to load balance instances according to a host selector strategy when multiple instances with the matching role exist.<br><br>For writer cluster endpoints or global writer endpoints, valid values are `writer` and `none`. For reader endpoints possible values are `reader` and `none`. For custom endpoints with the custom endpoint plugin enabled, possible values are `any`, `reader`, or `none`. For custom endpoints without the custom endpoint plugin enabled, the only possible value is `none`. If the value is set to `none`, the initial URL will not be replaced. Explicitly setting the parameter overrides the default behavior. | `reader`           | `writer` when using writer cluster endpoints or global writer cluster endpoints.<br><br>`reader` when using reader cluster endpoints.<br><br>`any` when using custom cluster endpoints and the custom endpoint plugin is enabled. Note that the host selected will still be limited to the instances in your custom endpoint. If the custom endpoint plugin is not enabled, `none` will be used.<br><br>Otherwise: `none`. |
```
sergiyvamz:
```suggestion
| `instanceSubstitutionRole`                        | String  |    No    | Defines whether or not the initial connection URL should be replaced with an instance URL from the topology info when available, and if so, the role of the instance URL that should be selected. This property should only be set if using a URL that resolves to a cluster URL (global writer, writer, reader, or custom). This behavior can improve the chances of connecting to the correct role. It also can be used in conjunction with `initialConnectionHostSelectorStrategy` to load balance instances according to a host selector strategy when multiple instances with the matching role exist.<br><br>For writer cluster endpoints or global writer endpoints, valid values are `writer` and `none`. For reader endpoints possible values are `reader` and `none`. For custom cluster endpoints with the custom endpoint plugin enabled, possible values are `any`, `reader`, or `none`. For custom endpoints without the custom endpoint plugin enabled, the only possible value is `none`. If the value is set to `none`, the initial URL will not be replaced. Explicitly setting the parameter overrides the default behavior. | `reader`           | `writer` when using writer cluster endpoints or global writer cluster endpoints.<br><br>`reader` when using reader cluster endpoints.<br><br>`any` when using custom cluster endpoints and the custom endpoint plugin is enabled. Note that the host selected will still be limited to the instances in your custom endpoint. If the custom endpoint plugin is not enabled, `none` will be used.<br><br>Otherwise: `none`. |
```

sergiyvamz: For custom endpoints with the custom endpoint plugin enabled
->
For custom cluster endpoints with the custom endpoint plugin enabled

sergiyvamz:
Could you please add few examples how to configure plugin to 
docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -114,239 +151,232 @@ public void initHostProvider(
...
-        if (isInitialConnection) {
-          hostListProviderService.setInitialConnectionHostSpec(writerCandidate);
+        // Verification failed.
+        // We will try again, unless a reader was requested but the cluster has no readers, which is a special case.
+        this.pluginService.forceRefreshHostList();
+        List<HostSpec> allHosts = this.pluginService.getAllHosts();
+        if (roleToVerify == HostRole.READER && !Utils.isNullOrEmpty(allHosts) && !this.hasReaders(allHosts)) {
+          // A reader was requested but the cluster has no readers.
+          // Simulate the reader cluster endpoint logic and return the current (writer) connection.
+          if (RoleVerificationSetting.READER.name().toLowerCase().equals(this.verifyRolePropValue)) {
```
sergiyvamz: RoleVerificationSetting.READER.name().toLowerCase() is always the same.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -72,26 +82,53 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
...
   }
 
   public AuroraInitialConnectionStrategyPlugin(final PluginService pluginService, final Properties properties) {
     this.pluginService = pluginService;
-    this.verifyOpenedConnectionType =
-        HostRole.verifyConnectionTypeFromValue(VERIFY_OPENED_CONNECTION_TYPE.getString(properties));
+    this.retryDelayMs = OPEN_CONNECTION_RETRY_INTERVAL_MS.getInteger(properties);
+    this.openConnectionRetryTimeoutNano =
+        TimeUnit.MILLISECONDS.toNanos(OPEN_CONNECTION_RETRY_TIMEOUT_MS.getInteger(properties));
+    this.verifyRolePropValue = VERIFY_OPENED_CONNECTION_ROLE.getString(properties);
```
sergiyvamz: You can convert the value to lower case or upper case here if needed.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -114,239 +151,232 @@ public void initHostProvider(
...
 
-    final int retryDelayMs = OPEN_CONNECTION_RETRY_INTERVAL_MS.getInteger(props);
-
-    final long endTimeNano = this.getTime()
-        + TimeUnit.MILLISECONDS.toNanos(OPEN_CONNECTION_RETRY_TIMEOUT_MS.getInteger(props));
+    // This is not an initial connection, or INSTANCE_SUBSTITUTION_ROLE was not set.
+    // We will pick a strategy according to the default behavior.
+    if (urlType == RdsUrlType.RDS_WRITER_CLUSTER
+        || urlType == RdsUrlType.RDS_GLOBAL_WRITER_CLUSTER) {
+      return InstanceSubstitutionStrategy.SUBSTITUTE_WRITER;
```
sergiyvamz: SUBSTITUTE_WRITER -> SUBSTITUTE_WITH_WRITER ?

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -114,239 +151,232 @@ public void initHostProvider(
...
+    } else if (urlType == RdsUrlType.RDS_READER_CLUSTER) {
+      return InstanceSubstitutionStrategy.SUBSTITUTE_READER;
+    } else if (urlType == RdsUrlType.RDS_CUSTOM_CLUSTER
+        // If the custom endpoint plugin is not in use, we should not substitute, because the host list will include
+        // instances that are not part of the custom endpoint.
+        && this.pluginService.isPluginInUse(CustomEndpointPlugin.class)) {
```
sergiyvamz: I wouldn't check for that plugin.

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -114,239 +151,232 @@ public void initHostProvider(
...
+          "AuroraInitialConnectionStrategyPlugin.invalidSettingForEndpoint",
+          new Object[]{INSTANCE_SUBSTITUTION_ROLE.name, "any", "writer cluster, reader cluster, or global cluster"}));
+    }
 
-          // Reader is not found. It seems that topology is outdated.
-          readerCandidateConn = connectFunc.call();
-          this.pluginService.forceRefreshHostList();
-          readerCandidate = this.pluginService.identifyConnection(readerCandidateConn);
+    if (urlType == RdsUrlType.RDS_CUSTOM_CLUSTER
+        && !this.pluginService.isPluginInUse(CustomEndpointPlugin.class)) {
```
sergiyvamz: The same, I'd not check for plugin

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
aaron-congo: If I remove this, users connecting with the default plugins and custom cluster URLs may end up connected to an instance outside their custom cluster because the custom endpoint plugin is not in use and we will return SUBSTITUTE_ANY here. These users will see incorrect behaviour when they upgrade if this check is removed. But I also know this check adds a performance penalty so I'm not sure what to do. Thoughts?

aaron-congo (approved):

PR #1673, opened 2026-01-13
fix: RW plugin - update references when connection is not null but already closed.

jonathanl-bq:
### Summary
This change updates how the RW plugin retains reader and writer references if they are currently in use to better match the logic used in the Go wrapper. We prefer Go's approach because the JDBC code only sets the members to null if the tracked connection is not null and not already closed. For the case where they are not null but are already closed we don't update the references. This change is to cover this additional condition.

This PR is based on discussion with @karenc-bq, which led to the closure of https://github.com/aws/aws-advanced-go-wrapper/pull/319


karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/ReadWriteSplittingPlugin.java
```diff
@@ -562,28 +562,32 @@ void closeReaderConnectionIfIdle(CacheItem<Connection> readerCacheItem) {
...
+        // Do nothing.
       }
-    } catch (SQLException e) {
-      // Do nothing.
+      this.readerConnection = null;
+      this.readerHostSpec = null;
     }
   }
 
   void closeWriterConnectionIfIdle(final Connection internalConnection) {
```
sergiyvamz: This method is always called with `this.WriteConnection` as parameter. Does it make sense to make it similar to closeReaderConnectionIfIdle() method? Otherwise it looks suspicious that you close `internalConnection` and then do `this.writerConnection = null;`

PR #1676, opened 2026-01-17
feat: introduce gdbFailover plugin

sergiyvamz:
### Summary

- introduces `gdbFailover` plugin that allows specifying different failover modes depending on whether home region is a primary region or a secondary region. The plugin is applicable to Global Database. 
- introduces new configuration parameters `verifyInactiveClusterWriterEndpointConnectionType` and `inactiveClusterWriterEndpointSubstitutionRole` for `initialConnection` plugin.
- introduces new configuration parameter `skipInactiveWriterClusterEndpointCheck` when using failover plugins (`failover`, `failover2`, `gdbFailover`).


wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -43,6 +45,13 @@ public class AuroraStaleDnsHelper {
 
   private static final Logger LOGGER = Logger.getLogger(AuroraStaleDnsHelper.class.getName());
 
+  private static final int RETRIES = 3;
```
karenc-bq: This is unused

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -277,9 +301,29 @@ public Connection connect(
...
-    if (urlType == RdsUrlType.RDS_WRITER_CLUSTER
-        || urlType == RdsUrlType.RDS_GLOBAL_WRITER_CLUSTER) {
+    if (urlType == RdsUrlType.RDS_GLOBAL_WRITER_CLUSTER) {
       return InstanceSubstitutionStrategy.SUBSTITUTE_WITH_WRITER;
+    } else if (urlType == RdsUrlType.RDS_WRITER_CLUSTER) {
+      final HostSpec writer = Utils.getWriter(this.pluginService.getAllHosts());
+      if (writer != null && this.rdsUtils.isRdsInstance(writer.getHost())) {
+        final String writerRegion = this.rdsUtils.getRdsRegion(writer.getHost());
+        final String clusterRegion = this.rdsUtils.getRdsRegion(originalConnectHost);
+        if (!clusterRegion.equalsIgnoreCase(writerRegion)) {
```
karenc-bq: Is it possible to move these comparisions to a method in rdsUtils? Seeing similar code several times throughout this PR

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -38,14 +37,16 @@ To enable the Aurora Initial Connection Strategy Plugin, add `initialConnection`
...
...e, and if so, the role of the instance URL that should be selected. This parameter is applicable to Aurora Global Databases. Region-bound cluster writer endpoints are available in Global Databases. However, depending on the GDB primary region, they may be inactive and may act differently. The parameter is aimed at supporting such inactive cluster endpoints and helps configure the desired behavior for them. Valid values are `writer` and `none`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      | `writer`           | `writer`                                                                                                                                                                                                           |
```
aaron-congo:
```suggestion
| `inactiveClusterWriterEndpointSubstitutionRole`     | String  |    No    | Defines whether or not the inactive cluster writer endpoint in the initial connection URL should be replaced with a writer instance URL from the topology info when available. This parameter is applicable to Aurora Global Databases. Region-bound cluster writer endpoints are available in Global Databases. However, depending on the GDB primary region, they may be inactive and may act differently. The parameter is aimed at supporting such inactive cluster endpoints and helps configure the desired behavior for them. Valid values are `writer` and `none`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      | `writer`           | `writer`                                                                                                                                                                                                           |
```

aaron-congo: Just a thought, the config parameter name might be confusing since it seems to be asking for a role even though the writer endpoint should always be a writer. Since the only valid values are "writer" and "none", I think making this a boolean property called `enableInactiveClusterWriterEndpointSubstitution` would be a bit clearer

docs/using-the-jdbc-driver/using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md
```diff
@@ -38,14 +37,16 @@ To enable the Aurora Initial Connection Strategy Plugin, add `initialConnection`
...
...should be performed. This parameter is applicable to Aurora Global Databases. Region-bound cluster writer endpoints are available in Global Databases. However, depending on the GDB primary region, they may be inactive and may act differently. The parameter is aimed at supporting such inactive cluster endpoints and helps configure the desired behavior for them. Valid values are `writer` and `none`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | `writer`           | `writer`                                                                                                                                                                                                           |
```
aaron-congo: same

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
+# Global Database (GDB) Failover Plugin
+The AWS Advanced JDBC Wrapper uses the GDB Failover Plugin to provide minimal downtime in the event of a DB instance failure. The plugin is based on Failover Plugin v2 [Failover2 Plugin](./UsingTheFailover2Plugin.md) and unless explicitly stated otherwise, most of the information and suggestions for the Failover Plugin and Failover Plugin v2 are applicable to the GDB Failover Plugin.
+
+## Plugin Availability
+The plugin is available since version 3.1.0.
```
aaron-congo: Since it be 3.1.1 since our current version is 3.1.0 and it will be in the next release?
```suggestion
The plugin is available since version 3.1.1.
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
...r-writer` - During failover, the driver tries to connect to any available/accessible reader node in home region. If no reader is available, the driver will connect to a writer node (in any region).<br>- `strict-home-reader` - During failover, the driver tries to connect to any available reader node in home region. If no reader is available, the driver raises an error.<br>- `strict-out-of-home-reader`, `strict-any-reader`, `out-of-home-reader-or-writer`, `any-reader-or-writer` - other possible values with prioritizing different writer or reader nodes in home, out of home, or in any region.<br><br>If this parameter is omitted, default value depends on connection url. For Aurora writer cluster endpoint or Global Database endpoint, it's set to `strict-writer`. Otherwise, it's `home-reader-or-writer`.                                                                                                                                                                                                  | Default value depends on connection url. For Aurora writer cluster endpoint, it's set to `strict-writer`. Otherwise, it's `home-reader-or-writer`.                                                  |
```
aaron-congo: for activeHomeFailoverMode and inactiveHomeFailoverMode, the description column mentions the default is "strict-writer" for both writer cluster and global endpoints, but the default value column says its only "strict-writer for an aurora writer cluster endpoint. Also, we could consider removing the default value part in the description since its repeated in the default value column

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | `30000`                                                                                                                                                                                             |
```
aaron-congo: instead of "explained above" can we put "explained in [link to failover 2 docs]"

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
... was detected.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | `100`                                                                                                                                                                                               |
```
aaron-congo: same suggestion to replace "described earlier"

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+#### Example 1
+**Goal:** Provide a user application with a writer connection. The application is deployed in `us-west-1` region and connects to Global Database with `us-east-1`, `us-east-2` and `us-west-1` regions.
+
+**Solution:** Use the following configuration parameters to configure the driver:
+- `failoverHomeRegion=us-west-1`
+- `activeHomeFailoverMode=strict-writer`
+- `inactiveHomeFailoverMode=strict-writer`
+- `globalClusterInstanceHostPatterns=?.XYZ1.us-east-1.rds.amazonaws.com,?.XYZ2.us-east-2.rds.amazonaws.com,?.XYZ3.us-west-1.rds.amazonaws.com` (make sure you replace `XYZ1`, `XYZ2`, `XYZ3` with actual values that correspond to your database)
+- `wrapperDialect=global-aurora-mysql` (or `global-aurora-pg`)
+- `wrapperPlugins=initialConnection,gdbFailover,ef2`
```
aaron-congo:
```suggestion
- `wrapperPlugins=initialConnection,gdbFailover,efm2`
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+#### Example 2
+**Goal:** Provide a user application with a reader connection in `us-west-1`. The application is deployed in the `us-west-1` region and connects to a Global Database with `us-east-1`, `us-east-2` and `us-west-1` regions. If the GDB primary region switches over, prioritize reader connections from the `us-west-1` region.
+
+**Solution:** Use the following configuration parameters to configure the driver:
+- `failoverHomeRegion=us-west-1`
+- `activeHomeFailoverMode=strict-home-reader`
+- `inactiveHomeFailoverMode=strict-home-reader`
+- `globalClusterInstanceHostPatterns=?.XYZ1.us-east-1.rds.amazonaws.com,?.XYZ2.us-east-2.rds.amazonaws.com,?.XYZ3.us-west-1.rds.amazonaws.com` (make sure you replace `XYZ1`, `XYZ2`, `XYZ3` with actual values that correspond to your database)
+- `wrapperDialect=global-aurora-mysql` (or `global-aurora-pg`)
+- `wrapperPlugins=initialConnection,gdbFailover,ef2`
```
aaron-congo:
```suggestion
- `wrapperPlugins=initialConnection,gdbFailover,efm2`
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -92,6 +92,18 @@ public class AuroraInitialConnectionStrategyPlugin extends AbstractConnectionPlu
               "writer", "reader", "any", "none"
           });
 
+  public static final AwsWrapperProperty INACTIVE_CLUSTER_WRITER_SUBSTITUTION_ROLE =
+      new AwsWrapperProperty(
+          "inactiveClusterWriterEndpointSubstitutionRole",
+          "writer",
+          "Defines whether or not the inactive cluster writer endpoint in initial connection URL should"
+              + " be replaced with an instance URL from the"
+              + " topology info when available, and if so, the role of the instance URL that should be selected.",
```
aaron-congo:
```suggestion
          "Defines whether or not the inactive cluster writer endpoint in the initial connection URL should"
              + " be replaced with a writer instance URL from the topology info when available.",
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -277,9 +301,29 @@ public Connection connect(
...
+          InstanceSubstitutionStrategy inactiveClusterWriterStrategy = InstanceSubstitutionStrategy.fromPropertyValue(
+              INACTIVE_CLUSTER_WRITER_SUBSTITUTION_ROLE.getString(props));
+          return inactiveClusterWriterStrategy != null
+              ? inactiveClusterWriterStrategy
+              : InstanceSubstitutionStrategy.SUBSTITUTE_WITH_WRITER;
+        }
+        return InstanceSubstitutionStrategy.SUBSTITUTE_WITH_WRITER;
+      } else {
+        return InstanceSubstitutionStrategy.DO_NOT_SUBSTITUTE;
+      }
```
aaron-congo, line 309: Gets a bit hard to follow with so many nested ifs, can we invert some of these to make it easier to follow? Suggestion below:
```suggestion
    } else if (urlType == RdsUrlType.RDS_WRITER_CLUSTER) {
      final HostSpec writer = Utils.getWriter(this.pluginService.getAllHosts());
      if (writer == null || !this.rdsUtils.isRdsInstance(writer.getHost())) {
        return InstanceSubstitutionStrategy.DO_NOT_SUBSTITUTE;
      }

      final String writerRegion = this.rdsUtils.getRdsRegion(writer.getHost());
      final String clusterRegion = this.rdsUtils.getRdsRegion(originalConnectHost);
      if (clusterRegion.equalsIgnoreCase(writerRegion)) {
        return InstanceSubstitutionStrategy.SUBSTITUTE_WITH_WRITER;
      }

      // The cluster writer endpoint belongs to a different region than the current writer region.
      // It means that the cluster is Aurora Global Database and cluster writer endpoint is in secondary region.
      // In this case the cluster writer endpoint is in inactive state and doesn't represent the current writer
      // so any connection check should be skipped.
      // Continue with a normal workflow.
      InstanceSubstitutionStrategy inactiveClusterWriterStrategy = InstanceSubstitutionStrategy.fromPropertyValue(
          INACTIVE_CLUSTER_WRITER_SUBSTITUTION_ROLE.getString(props));
      return inactiveClusterWriterStrategy != null
          ? inactiveClusterWriterStrategy
          : InstanceSubstitutionStrategy.SUBSTITUTE_WITH_WRITER;
```

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -277,9 +301,29 @@ public Connection connect(
...
+    } else if (urlType == RdsUrlType.RDS_WRITER_CLUSTER) {
+      final HostSpec writer = Utils.getWriter(this.pluginService.getAllHosts());
+      if (writer != null && this.rdsUtils.isRdsInstance(writer.getHost())) {
+        final String writerRegion = this.rdsUtils.getRdsRegion(writer.getHost());
+        final String clusterRegion = this.rdsUtils.getRdsRegion(originalConnectHost);
+        if (!clusterRegion.equalsIgnoreCase(writerRegion)) {
+          // The cluster writer endpoint belongs to a different region than the current writer region.
+          // It means that the cluster is Aurora Global Database and cluster writer endpoint is in secondary region.
+          // In this case the cluster writer endpoint is in inactive state and doesn't represent the current writer
+          // so any connection check should be skipped.
```
aaron-congo: this comments mentions a connection check, but this code is about URL substitution instead of role verification

wrapper/src/main/java/software/amazon/jdbc/plugin/gdbfailover/GlobalDbFailoverConnectionPlugin.java
```diff
@@ -0,0 +1,520 @@
...
+      new AwsWrapperProperty(
+          "failoverHomeRegion", null,
+          "Set home region for failover.");
+
+  static {
+    PropertyDefinition.registerPluginProperties(GlobalDbFailoverConnectionPlugin.class);
+  }
+
+  // Inherited failoverMode member should not be used in this class.
+  // Use homeFailoverMode and outOfHomeFailoverMode instead.
```
aaron-congo:
```suggestion
  // Use activeHomeFailoverMode and inactiveHomeFailoverMode instead.
```

wrapper/src/main/java/software/amazon/jdbc/plugin/gdbfailover/GlobalDbFailoverConnectionPlugin.java
```diff
@@ -0,0 +1,520 @@
...
+
+
+  public GlobalDbFailoverConnectionPlugin(PluginService pluginService,
+      Properties properties) {
+    super(pluginService, properties);
+  }
+
+  @Override
+  protected void initFailoverMode() throws SQLException {
+    if (this.rdsUrlType == null) {
```
aaron-congo: to reduce nesting level
```suggestion
    if (this.rdsUrlType != null) {
      return;
```

wrapper/src/main/java/software/amazon/jdbc/plugin/gdbfailover/GlobalDbFailoverConnectionPlugin.java
```diff
@@ -0,0 +1,520 @@
...
+  protected void failoverToWriter(HostSpec writerCandidate) throws SQLException {
+    if (this.failoverWriterTriggeredCounter != null) {
+      this.failoverWriterTriggeredCounter.inc();
+    }
+    Connection writerCandidateConn = null;
+    try {
+      final List<HostSpec> allowedHosts = this.pluginService.getHosts();
+      if (!Utils.containsHostAndPort(allowedHosts, writerCandidate.getHostAndPort())) {
+        if (this.failoverWriterFailedCounter != null) {
+          this.failoverWriterFailedCounter.inc();
```
aaron-congo, line 313: I think these 3 inner inc() calls should be removed since the catch block will increase it a second time even though it should only be increased once

wrapper/src/main/java/software/amazon/jdbc/plugin/gdbfailover/GlobalDbFailoverConnectionPlugin.java
```diff
@@ -0,0 +1,520 @@
...
+      this.pluginService.refreshHostList();
+      List<HostSpec> updatedAllowedHosts = allowedHosts.get();
+      // Make a copy of hosts and set their availability.
+      updatedAllowedHosts = updatedAllowedHosts.stream()
+          .map(x -> this.pluginService.getHostSpecBuilder()
+              .copyFrom(x)
+              .availability(HostAvailability.AVAILABLE)
+              .build())
+          .collect(Collectors.toList());
+      final Set<HostSpec> remainingAllowedHosts = new HashSet<>(updatedAllowedHosts);
```
aaron-congo, line 438: Does this work?
```suggestion
      final Set<HostSpec> remainingAllowedHosts = updatedAllowedHosts.stream()
          .map(x -> this.pluginService.getHostSpecBuilder()
              .copyFrom(x)
              .availability(HostAvailability.AVAILABLE)
              .build())
          .collect(Collectors.toSet());
```

wrapper/src/main/java/software/amazon/jdbc/plugin/gdbfailover/GlobalDbFailoverConnectionPlugin.java
```diff
@@ -0,0 +1,520 @@
...
+
+      //LOGGER.finest(LogUtils.logTopology(updatedAllowedHosts, "Allowed hosts:"));
```
aaron-congo: Can be removed

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
+# Global Database (GDB) Failover Plugin
+The AWS Advanced JDBC Wrapper uses the GDB Failover Plugin to provide minimal downtime in the event of a DB instance failure. The plugin is based on Failover Plugin v2 [Failover2 Plugin](./UsingTheFailover2Plugin.md) and unless explicitly stated otherwise, most of the information and suggestions for the Failover Plugin and Failover Plugin v2 are applicable to the GDB Failover Plugin.
+
+## Plugin Availability
+The plugin is available since version 3.1.0.
+
+## Differences between the GDB Failover Plugin and the Failover Plugin v2
+
+The GDB Failover Plugin introduces a notion of home region and extends configuration parameters to allow setting different failover logic for `in-home` and `out-of-home` cases.
```
sophia-bq:
```suggestion
The GDB Failover Plugin introduces the notion of home region and extends configuration parameters to allow setting different failover logic for `in-home` and `out-of-home` cases.
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+
+Driver configuration should include a home region. The home region is defined as an AWS region name. When the primary region of the Global Database is the same as the specified home region, it's considered a `home` case. When GDB switches over to another region and the primary region is not the same as the specified home region, the driver considers it an `out of home` case.
```
sophia-bq:
```suggestion
Driver configuration for the GDB failover plugin should include a home region defined with an AWS region name. This introduces two cases:
1. **in-home**: when the primary region of the Global Database is the same as the specified home region
2. **out-of-home**: when GDB switches over to another region and the primary region is not the same as the specified home region
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+
+Users are allowed to specify different failover logic individually for in-home and out-of-home cases. For example, when a user application needs a writer connection, it makes sense to define the failover mode to follow the writer (`strict-writer`). When in-region failover occurs, the driver reconnects to a new writer node and continues serving the user application with a write connection. Some applications may choose not to follow a writer node when cross-region failover occurs. In this case, application logic may decide to turn to inactive mode and stop performing writes rather than being connected to a writer node in another region and lose performance due to increased connection latency. Such reasoning is up to the user application logic and such concerns are valid. 
+
+Another common example is to configure the driver to serve a user application with a reader connection and prioritize reader connections from the home region. One of the reasons for such a decision is to have connection latency as small as possible.
```
sophia-bq:
```suggestion
Users are allowed to specify different failover logic individually for in-home and out-of-home cases. The following are two examples where in-home and out-of-home cases impact the failover logic.

**Example 1**
When a user application needs a writer connection, it makes sense to define the failover mode to follow the writer (`strict-writer`). However, some applications may choose not to follow a writer node when cross-region failover occurs.

- in-home: when in-region failover occurs, the driver reconnects to a new writer node and continues serving the user application with a write connection
- out-of-home: application turns to inactive mode and stops performing writes, this prioritizes performance due to reduced connection latency over being connected to a writer node in another region

**Example 2**
When a user application needs a reader connection, prioritize reader connections in the home region to reduce connection latency.

- in-home: the primary region of the Global Database is the same as the specified home region, connect to any such reader
- out-of-home: connect to readers that are in the specified home region, these readers are no longer part of the primary region of the Global Database
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+
+The mentioned scenarios can be configured with the GDB Failover plugin. [Failover2 Plugin](./UsingTheFailover2Plugin.md) and [Failover Plugin](./UsingTheFailoverPlugin.md) don't support the home region and can't be used for the cases mentioned above.
+
+## Using the GDB Failover Plugin
+The GDB Failover Plugin will not be enabled by default if the [`wrapperPlugins`](../UsingTheJdbcDriver.md#connection-plugin-manager-parameters) value is not specified. If you would like to override the default plugins, you can explicitly include the GDB Failover Plugin in your list of plugins by adding the plugin code `gdbFailover` to the [`wrapperPlugins`](../UsingTheJdbcDriver.md#aws-advanced-jdbc-wrapper-parameters) value, or by adding it to the current [driver profile](../UsingTheJdbcDriver.md#connection-plugin-manager-parameters). After you load the plugin, the failover feature will be enabled.
+
+Please refer to the [failover configuration guide](../FailoverConfigurationGuide.md) for tips to keep in mind when using the failover plugin.
+
+> [!WARNING]
+> Do not use the `gdbFailover`, `failover` and/or `failover2` plugins (or their combination) at the same time for the same connection!
```
sophia-bq, line 38: We should update this warning in the other Failover plugins' docs

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
...address), the configuration parameter is mandatory.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |                                                                                                                                                                                                     |
```
sophia-bq:
```suggestion
| `failoverHomeRegion`                     |  String  | If connecting using an IP address, a custom domain URL, Global Database endpoint or other endpoint with no region: Yes<br><br>Otherwise: No | Defines a home region.<br><br>Examples: `us-west-2`, `us-east-1`. <br><br>If this parameter is omitted, the value is parsed from connection url. For regional cluster endpoints and instance endpoints, it's set to the region of the provided endpoint. If the provided endpoint has no region (for example, a Global Database endpoint or IP address) the configuration parameter is mandatory.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |                                                                                                                                                                                                     |
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
sophia-bq: This is mostly a reformatting to make it easier to read. I tried to keep the content mostly the same. It would also be great if we gave the connection parameters for these examples or links to where there would be an example of the settings for these

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
sophia-bq: Wow! A lot of new possible values! Might it make sense to have these options have their own table with descriptions for each?

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+**Solution:** Use the following configuration parameters to configure the driver:
+- `failoverHomeRegion=us-west-1`
+- `activeHomeFailoverMode=strict-writer`
+- `inactiveHomeFailoverMode=strict-writer`
+- `globalClusterInstanceHostPatterns=?.XYZ1.us-east-1.rds.amazonaws.com,?.XYZ2.us-east-2.rds.amazonaws.com,?.XYZ3.us-west-1.rds.amazonaws.com` (make sure you replace `XYZ1`, `XYZ2`, `XYZ3` with actual values that correspond to your database)
+- `wrapperDialect=global-aurora-mysql` (or `global-aurora-pg`)
+- `wrapperPlugins=initialConnection,gdbFailover,ef2`
+- use Global Database endpoint in your connection string
+
+#### Example 2
```
sophia-bq: Great let's link this example to when we first bring it up in line 15

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+- `wrapperPlugins=initialConnection,gdbFailover,ef2`
+- use cluster reader endpoint in region `us-west-1` in your connection string (`<cluster-name>.cluster-ro-XYZ3.us-west-1.rds.amazonaws.com`)
+
+#### Example 3
+**Goal:** Provide a user application with a writer connection when the primary region is the closest to the application deployed region. When the primary region switches over to another region and network latency becomes unacceptable, the application deployment in region `us-west-1` becomes inactive and lets applications in other regions act and process business transactions. There are application deployments in other regions. The Global Database has `us-east-1`, `us-east-2` and `us-west-1` regions. 
+
+**Solution:** Use the following configuration parameters to configure the driver for application deployment in the `us-west-1` region. Other application deployments need similar driver configuration with different home regions.
+- `failoverHomeRegion=us-west-1`
+- `activeHomeFailoverMode=strict-writer`
+- `inactiveHomeFailoverMode=strict-any-reader` (or `strict-home-reader`)
```
sophia-bq: Let's keep it simpler and give this example for only strict-any-reader as that is what we mention in the explanation 
```suggestion
- `inactiveHomeFailoverMode=strict-any-reader`
```

docs/using-the-jdbc-driver/using-plugins/UsingTheGdbFailoverPlugin.md
```diff
@@ -0,0 +1,97 @@
...
+**Solution:** Use the following configuration parameters to configure the driver:
+- `failoverHomeRegion=us-west-1`
+- `activeHomeFailoverMode=strict-home-reader`
+- `inactiveHomeFailoverMode=strict-home-reader`
+- `globalClusterInstanceHostPatterns=?.XYZ1.us-east-1.rds.amazonaws.com,?.XYZ2.us-east-2.rds.amazonaws.com,?.XYZ3.us-west-1.rds.amazonaws.com` (make sure you replace `XYZ1`, `XYZ2`, `XYZ3` with actual values that correspond to your database)
+- `wrapperDialect=global-aurora-mysql` (or `global-aurora-pg`)
+- `wrapperPlugins=initialConnection,gdbFailover,ef2`
+- use cluster reader endpoint in region `us-west-1` in your connection string (`<cluster-name>.cluster-ro-XYZ3.us-west-1.rds.amazonaws.com`)
+
+#### Example 3
```
sophia-bq: Same comment as above let's link this to when we first bring it up earlier in the documentation

wrapper/src/test/java/integration/container/tests/GdbFailoverTest.java
```diff
@@ -0,0 +1,157 @@
...
+@EnableOnNumOfInstances(min = 2)
+@MakeSureFirstInstanceWriter
+@Order(16)
+public class GdbFailoverTest extends FailoverTest {
+
+  @TestTemplate
+  @EnableOnNumOfInstances(min = 2)
+  @EnableOnTestFeature(TestEnvironmentFeatures.NETWORK_OUTAGES_ENABLED)
+  @Override
+  public void test_readerFailover_readerOrWriter() throws SQLException {
```
sergiyvamz, line 62: .

wrapper/src/main/java/software/amazon/jdbc/util/RdsUrlType.java
```diff
@@ -43,4 +45,8 @@ public boolean isRds() {
   public boolean isRdsCluster() {
     return this.isRdsCluster;
   }
+
+  public boolean isHasRegion() {
```
karenc-bq: Can we rename this to just `hasRegion`

wrapper/src/main/java/software/amazon/jdbc/ConnectionPluginChainBuilder.java
```diff
@@ -75,6 +75,7 @@ public class ConnectionPluginChainBuilder {
           put("efm2", new software.amazon.jdbc.plugin.efm2.HostMonitoringConnectionPluginFactory());
           put("failover", new FailoverConnectionPluginFactory());
           put("failover2", new software.amazon.jdbc.plugin.failover2.FailoverConnectionPluginFactory());
+          put("gdbFailover", new software.amazon.jdbc.plugin.gdbfailover.GlobalDbFailoverConnectionPluginFactory());
```
karenc-bq: Why does GDB need the fully qualified class name

PR #1679, opened 2026-01-22
fix: IAM related plugins unable to parse the correct region from global cluster endpoints

karenc-bq:
### Summary

Plugins utilizing IAM authentication tokens weren't supported for global clusters due to not being able to fetch the correct primary cluster region.
This change uses the RDS SDK to determine primary cluster, and generate IAM auth token for that region.

### DIFF
--- a/docs/using-the-jdbc-driver/using-plugins/UsingTheIamAuthenticationPlugin.md
+++ b/docs/using-the-jdbc-driver/using-plugins/UsingTheIamAuthenticationPlugin.md
@@ -13,7 +13,7 @@ AWS Identity and Access Management (IAM) grants users access control across all 
 > Note: AWS Java SDK RDS may have transitive dependencies that are also required (ex. [AWS Java SDK Core](https://central.sonatype.com/artifact/software.amazon.awssdk/aws-core/)). If you are not using a package manager such as Maven or Gradle, please refer to Maven Central to determine these transitive dependencies.
 >
 > Since [AWS Java SDK RDS v2.x](https://central.sonatype.com/artifact/software.amazon.awssdk/rds) size is around 5.4Mb (22Mb including all RDS SDK dependencies), some users may experience difficulties using the plugin due to limited available disk size.
-> In such cases, the [AWS Java SDK RDS v2.x](https://central.sonatype.com/artifact/software.amazon.awssdk/rds) dependency may be replaced with just two dependencies which have a smaller footprint (around 300Kb in total):
+> In such cases, the [AWS Java SDK RDS v2.x](https://central.sonatype.com/artifact/software.amazon.awssdk/rds) dependency may be replaced with just two dependencies which have a smaller footprint (around 300Kb in total)<sup>1</sup>:
 > - [software.amazon.awssdk:http-client-spi](https://central.sonatype.com/artifact/software.amazon.awssdk/http-client-spi)
 > - [software.amazon.awssdk:auth](https://central.sonatype.com/artifact/software.amazon.awssdk/auth)
 >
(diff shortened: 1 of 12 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/GDBRegionUtils.java
```diff
@@ -0,0 +1,72 @@
...
+public class GDBRegionUtils extends RegionUtils {
+  private static final RdsClient rdsClient;
+  private static final Pattern GDB_CLUSTER_ARN_PATTERN =
+      Pattern.compile("^arn:aws:rds:(?<region>[^:\\n]*):[^:\\n]*:([^:/\\n]*[:/])?(.*)$");
+  private static final String REGION_GROUP = "region";
+
+  static {
+    try {
+      Class.forName("software.amazon.awssdk.services.rds.RdsClient");
+      rdsClient = RdsClient.builder().build();
```
sergiyvamz: oh no, not in static contrsuctor!

docs/using-the-jdbc-driver/using-plugins/UsingTheIamAuthenticationPlugin.md
```diff
@@ -57,3 +57,31 @@ IAM database authentication use is limited to certain database engines. For more
...
+    ]
+}
+
+```
+
+> [!NOTE]
+> [AWS Java SDK RDS v2.x](https://central.sonatype.com/artifact/software.amazon.awssdk/rds) is **required** when using this plugin with Global databases.
+
+---
+<sup>1</sup> Note: The smaller dependencies cannot be used with Global databases, which require the full [AWS Java SDK RDS v2.x](https://central.sonatype.com/artifact/software.amazon.awssdk/rds) dependency.
```
sergiyvamz, line 87: "smaller dependencies" doesn't sound clear

wrapper/src/main/java/software/amazon/jdbc/plugin/iam/IamAuthConnectionPlugin.java
```diff
@@ -76,7 +78,7 @@ public class IamAuthConnectionPlugin extends AbstractConnectionPlugin {
       "iamAccessTokenPropertyName", PropertyDefinition.PASSWORD.name,
       "Overrides default IAM access token property name");
 
-  protected static final RegionUtils regionUtils = new RegionUtils();
+  protected static RegionUtils regionUtils;
```
sergiyvamz: If user uses `awsProfile` and the values are different for different connections then this static util can use wrong profile and wrong AWS credentials.

wrapper/src/main/java/software/amazon/jdbc/plugin/federatedauth/FederatedAuthPlugin.java
```diff
@@ -177,13 +180,17 @@ private Connection connectInternal(final HostSpec hostSpec, final Properties pro
...
-    final String host = IamAuthUtils.getIamHost(IAM_HOST.getString(props), hostSpec);
+    final HostSpec host = IamAuthUtils.getIamHost(IAM_HOST.getString(props), hostSpec);
 
     final int port = IamAuthUtils.getIamPort(
         IAM_DEFAULT_PORT.getInteger(props),
         hostSpec,
         this.pluginService.getDialect().getDefaultPort());
 
+    final RdsUrlType type = rdsUtils.identifyRdsType(host.getHost());
+    if (regionUtils == null) {
```
sergiyvamz: I'm not sure whether regionUtils can be static. It seems it's a short live object and depends on provided host.

PR #1680, opened 2026-01-23
fix: custom cluster of type reader with excluded members incorrectly allowed writer host connections

aaron-congo:
Before these changes, a bug was occurring when connecting to custom clusters of type reader with excluded member lists. When using this configuration, the custom cluster URL will not connect to the writer, even if the writer is not an excluded member. `PluginService#getHosts` was filtering out the blocked hosts but failed to filter out the writer instance in this scenario. Consequently, users could end up connecting to the writer instance with our driver even though the writer was not a part of their custom cluster. These changes address this problem by filtering out hosts that do not match the required role of the custom cluster.

Interestingly, even though custom clusters of type reader with excluded members do not allow writer connections, custom clusters with static members always allow connections to any static member, regardless of role. More specifically, even if a custom cluster is of type reader, if its member list type is static and the writer is one of the static members, it will still allow connections to the writer.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
@@ -71,7 +71,6 @@ public class PluginServiceImpl implements PluginService, CanReleaseResources,
 
   protected static final CacheMap<String, HostAvailability> hostAvailabilityExpiringCache = new CacheMap<>();
 
-  protected final RdsUtils rdsUtils = new RdsUtils();
   protected final FullServicesContainer servicesContainer;
   protected final ConnectionPluginManager pluginManager;
   private final Properties props;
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/PluginServiceImpl.java
```diff
@@ -69,8 +70,9 @@ public class PluginServiceImpl implements PluginService, CanReleaseResources,
   protected static final int DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS = 5000;
 
   protected static final CacheMap<String, HostAvailability> hostAvailabilityExpiringCache = new CacheMap<>();
-  protected final FullServicesContainer servicesContainer;
 
+  protected final RdsUtils rdsUtils = new RdsUtils();
```
sergiyvamz: Do we need it?

aaron-congo: forgot to remove this, good catch, thanks

PR #1683, opened 2026-01-23
fix: memory spike in MySQL when creating statements

faviansamatha:
### Summary

Fix memory spike in MySQL with creating Statement handlers by caching prepared statements

### DIFF
--- a/examples/AWSDriverExample/build.gradle.kts
+++ b/examples/AWSDriverExample/build.gradle.kts
@@ -21,14 +21,14 @@ dependencies {
     implementation("software.amazon.awssdk:rds:2.33.5")
     implementation("software.amazon.awssdk:secretsmanager:2.33.5")
     implementation("software.amazon.awssdk:sts:2.33.5")
-    implementation("com.fasterxml.jackson.core:jackson-databind:2.19.0")
+    implementation("com.fasterxml.jackson.core:jackson-databind:2.21.0")
     implementation(project(":aws-advanced-jdbc-wrapper"))
     implementation("io.opentelemetry:opentelemetry-api:1.52.0")
     implementation("io.opentelemetry:opentelemetry-sdk:1.51.0")
     implementation("io.opentelemetry:opentelemetry-exporter-otlp:1.52.0")
     implementation("com.amazonaws:aws-xray-recorder-sdk-core:2.18.2")
     implementation("org.jsoup:jsoup:1.21.1")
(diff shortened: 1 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
```diff
@@ -387,14 +412,14 @@ protected void collectStatus() {
               new Object[] {this.role, BlueGreenPhase.NOT_CREATED}));
         } else {
           this.connection.set(null);
+          this.clearCheckStatusStatement();
```
sergiyvamz, line 414: It should go before closing connection.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
```diff
@@ -479,6 +504,7 @@ protected void collectStatus() {
               .port(statusInfo.port)
               .build());
           this.connectionHostSpecCorrect.set(true);
+          this.clearCheckStatusStatement();
```
sergiyvamz, line 508: It should go before closing connection.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
```diff
@@ -387,14 +412,14 @@ protected void collectStatus() {
...
           this.currentPhase = null;
           this.panicMode.set(true);
         }
         return;
       }
 
-      final Statement statement = conn.createStatement();
-      final ResultSet resultSet = statement.executeQuery(this.blueGreenDialect.getBlueGreenStatusQuery());
+      ResultSet resultSet = this.executeCheckStatusStatement(conn);
```
sergiyvamz: You need to close ResultSet. You can use tr-with-resource block.

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
faviansamatha: I think this one is in the right place, it's before the the connection.set(null)

PR #1695, opened 2026-01-29
chore: remove IP Address checking in staleDnsHelper

faviansamatha:
### Summary

Currently, staleDnsHelper helps check stale dns by checking the IP address of the cluster and writer host, but this method didn't really work too well. 

This PR modifies it by removing the IP address, and switching to the writer host if the initial connection is a reader.

### Solution
Capture the connection's role (isConnectedToReader) immediately after checking it, before DNS can change. Then use this boolean in the final condition to ensure we reconnect to the writer if the connection is to a reader, regardless of what DNS currently resolves to.

### DIFF
--- a/wrapper/src/test/java/integration/util/AuroraTestUtility.java
+++ b/wrapper/src/test/java/integration/util/AuroraTestUtility.java
@@ -1457,6 +1457,12 @@ public class AuroraTestUtility {
     if (DatabaseEngineDeployment.RDS_MULTI_AZ_CLUSTER.equals(deployment)) {
       // Old multi-AZ writers take 10-20min to go up after server failover, so we will simulate failover.
       simulateTemporaryFailure(executor, instanceId);
+
+      try {
+        TimeUnit.MILLISECONDS.sleep(1000);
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/staledns/AuroraStaleDnsHelper.java
```diff
@@ -147,22 +130,7 @@ public Connection getVerifiedConnection(
...
-    LOGGER.finest(() -> Messages.get("AuroraStaleDnsHelper.writerInetAddress",
-        new Object[]{this.writerHostAddress}));
-
-    if (this.writerHostAddress == null) {
-      return conn;
-    }
-
-    if (!writerHostAddress.equals(clusterInetAddress)) {
+    if (isConnectedToReader) {
       // DNS resolves a cluster endpoint to a wrong writer
```
sergiyvamz: I'm not sure the comment is correct.

sergiyvamz (approved):

sergiyvamz (approved):

PR #1696, opened 2026-01-29
Fix: Correctly set Host ID in createHost in TopologyUtils

jonathanl-bq:
### Summary

This PR stops the `hostId` in `createHost` in TopologyUtils from being overwritten with the `instanceName`.


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/TopologyUtils.java
```diff
@@ -177,7 +175,6 @@ public HostSpec createHost(
         .lastUpdateTime(lastUpdateTime)
         .build();
     hostSpec.addAlias(instanceName);
-    hostSpec.setHostId(instanceName);
```
sergiyvamz, line 180: I'd suggest to keep this line removed and rollback other changes in this PR.

aaron-congo: We want the hostId field set to instanceName though right? If we keep this line and remove the other changes the hostId will be set to instanceId instead of instanceName

sergiyvamz: We want to set hostId with provided value and add instanceName to set of aliases. For Aurora hostId and instanceName are the same but it's up to the caller to pass the right (the same) values.

aaron-congo: so the hostId should be set to instanceId and instanceName should only be added as an alias? The code on main branch sets the hostId to the instanceName, so we would be changing that if we make this change

sergiyvamz: That's correct and that's what we want to fix.

PR #1697, opened 2026-01-29
fix: consider topology verified if topology from all readers remains consistent

aaron-congo:
### Description
- addresses an issue with the ClusterTopologyMonitor, particularly in conjunction with the GdbFailoverPlugin. Before these changes, if the application lost connection to the writer instance for awhile but the server was not actually failing over, forceRefresh(shouldVerifyWriter=true) would timeout without the GdbFailover attempting failover. This would occur because forceRefresh(shouldVerifyWriter=true) required a writer connection before returning, and the GdbFailoverPlugin does not attempt failover until the primary writer region is identified. To fix this issue, forceRefresh(shouldVerify=true) now also returns if topology from all readers remains consistent over multiple queries.
- also fixes an issue where the wrong dialect was being used when connection to global mysql clusters

### DIFF
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -22,7 +22,7 @@ dependencies {
     jmhImplementation(project(":aws-advanced-jdbc-wrapper"))
     implementation("org.postgresql:postgresql:42.7.7")
     implementation("com.mysql:mysql-connector-j:9.4.0")
-    implementation("org.mariadb.jdbc:mariadb-java-client:3.5.6")
+    implementation("org.mariadb.jdbc:mariadb-java-client:3.5.7")
     implementation("com.zaxxer:HikariCP:4.0.3")
     implementation("org.checkerframework:checker-qual:3.49.5")
 
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -56,7 +56,7 @@ pluginManagement {
         id("com.github.spotbugs") version "6.1.+"
         id("com.diffplug.spotless") version "6.13.0" // 6.13.0 is the last version that is compatible with Java 8
         id("com.github.vlsi.gradle-extensions") version "1.+"
-        id("com.github.vlsi.stage-vote-release") version "1.+"
+        id("com.github.vlsi.stage-vote-release") version "3.+"
         id("com.github.vlsi.ide") version "3.+"
         id("me.champeau.jmh") version "0.+"
         id("org.checkerframework") version "0.+"
--- a/wrapper/build.gradle.kts
+++ b/wrapper/build.gradle.kts
@@ -53,7 +53,7 @@ dependencies {
     compileOnly("org.checkerframework:checker-qual:3.49.5")
     compileOnly("com.mysql:mysql-connector-j:9.4.0")
     compileOnly("org.postgresql:postgresql:42.7.7")
-    compileOnly("org.mariadb.jdbc:mariadb-java-client:3.5.6")
+    compileOnly("org.mariadb.jdbc:mariadb-java-client:3.5.7")
     compileOnly("org.osgi:org.osgi.core:6.0.0")
     compileOnly("org.jetbrains.kotlin:kotlin-stdlib:2.2.20")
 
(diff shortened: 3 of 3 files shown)


wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -375,13 +378,15 @@ public void monitor() throws Exception {
             }
           }
 
+          checkForStableReaderTopologies();
```
sergiyvamz, line 393: this

wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
```diff
@@ -228,7 +228,6 @@ public boolean equals(final Object obj) {
     return Objects.equals(this.host, spec.host)
         && this.port == spec.port
         && this.availability == spec.availability
-        && this.role == spec.role
-        && this.weight == spec.weight;
+        && this.role == spec.role;
```
sergiyvamz: That might cause side effects. Can we avoid changing this method?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -441,6 +458,55 @@ public void monitor() throws Exception {
...
   }
 
+  protected void checkForStableReaderTopologies() {
+    List<HostSpec> latestHosts = this.getStoredHosts();
+    if (Utils.isNullOrEmpty(latestHosts)) {
+      this.stableTopologiesStartNano = 0;
+      return;
+    }
+
+    Set<String> readerIds = latestHosts.stream().map(HostSpec::getHostId).collect(Collectors.toSet());
```
sergiyvamz: I think List<> would be better in this case. Set<> is good to check whether item in the set or not, not for iterating though items in the set.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -441,6 +458,55 @@ public void monitor() throws Exception {
...
+    for (String id : readerIds) {
+      Boolean completedOneCycle = this.completedOneCycle.get(id);
```
sergiyvamz: You can use getOrDefault() to simplify the statement.

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -793,6 +862,7 @@ public void run() {
                 }
               } catch (SQLException e) {
                 // Invalid connection, retry.
+                this.monitor.completedOneCycle.put(this.hostSpec.getHostId(), Boolean.TRUE);
```
sergiyvamz, line 870: Should we set false?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -767,6 +835,7 @@ public void run() {
                 // It might be some transient error. Let's try again.
                 //  If the error repeats, we will try again after a longer delay.
                 TimeUnit.MILLISECONDS.sleep(this.calculateBackoffWithJitter(this.connectionAttempts++));
+                this.monitor.completedOneCycle.put(this.hostSpec.getHostId(), Boolean.TRUE);
```
sergiyvamz, line 842: Should we set false?

wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
```diff
@@ -758,6 +825,7 @@ public void run() {
                 // It's a network issue that's expected during a cluster failover.
                 // We will try again on the next iteration.
                 TimeUnit.MILLISECONDS.sleep(100);
+                this.monitor.completedOneCycle.put(this.hostSpec.getHostId(), Boolean.TRUE);
```
sergiyvamz, line 831: Should we set false?

sergiyvamz (approved):

PR #1701, opened 2026-02-03
perf: optimize WeightedRandomHostSelector 

faviansamatha:
### Summary
perf: optimize WeightedRandomHostSelector

### Description
The previous implementation of WeightedRandomHostSelector that's used in limitless was not ideal. It did the following:

1. From a list of hosts, create a host-weight pair string
2. Create a host weight map pair property (could be cached, but caching also includes some string operations to check the cache)
3. Create a list of eligible hosts
4. Create a map of eligible hosts and a weight range
5. Get a random number and go through the map created in step 3 and get a host where its weight range is within that number. 

This means that we would have to go through the list 4-5 times.

When used in limitless, we also re-create this string from the same map of eligible hosts and their weights.

This PR optimizes the weightedRandom and limitless by doing the following:
- Remove dependency on `weightedRandomHostWeightPairs` property for WeightedRandomHostSelector.
  - WeightedRandomHostSelector will use the host' weights if property is not defined
  - If property is defined, it will use that static list.
- Remove limitless call to set this property and add validation to not allow users to set `weightedRandomHostWeightPairs`
  - Previously, if users set this property, then it will be overridden by limitless. In the current change, it will cause limitless not to use the weights from the hostlist which is incorrect.
- Use weighted reservoir sampling alg. in WeightedRandomHostSelector


wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -60,57 +58,83 @@ public HostSpec getHost(
...
+      if ((role == null || role.equals(host.getRole()))
+          && host.getAvailability().equals(HostAvailability.AVAILABLE)) {
+
+        long weight = host.getWeight();
+        if (weight < 1) {
+          weight = DEFAULT_WEIGHT;
+        }
+
+        totalWeight += weight;
+        if ((long) (this.random.nextDouble() * totalWeight) < weight) {
```
sergiyvamz: It picks up the first element always. On the first iteration totalWeight has the same value as weight. So this condition is always true.

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -60,57 +58,83 @@ public HostSpec getHost(
...
-        hostWeightRangeMap.put(host.getHost(), new NumberRange(rangeStart, rangeEnd));
-        counter = counter + hostWeight;
-      } else {
-        hostWeightRangeMap.put(host.getHost(), new NumberRange(counter, counter));
-        counter++;
+    HostSpec selected = null;
+    long totalWeight = 0;
+
+    for (HostSpec host : hosts) {
+      if ((role == null || role.equals(host.getRole()))
```
aaron-congo: nit: can invert to decrease level of nesting
```suggestion
      if ((role != null && !role.equals(host.getRole())) || !host.getAvailability().equals(HostAvailability.AVAILABLE)) {
      continue;
     }
```

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -60,57 +58,83 @@ public HostSpec getHost(
...
-    // Check random number is in host weight range map
-    for (final HostSpec host : eligibleHosts) {
-      NumberRange range = hostWeightRangeMap.get(host.getHost());
-      if (range != null && range.isInRange(randomInt)) {
-        return host;
+    HostSpec selected = null;
+    long totalWeight = 0;
+
+    for (HostSpec host : hosts) {
+      if ((role == null || role.equals(host.getRole()))
```
aaron-congo: nit: can invert to decrease nesting level

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
```diff
@@ -60,57 +58,83 @@ public HostSpec getHost(
...
+          && host.getAvailability().equals(HostAvailability.AVAILABLE)) {
+
+        Integer weight = hostWeightMap.get(host.getHost());
+        if (weight == null || weight < 1) {
+          continue;
```
aaron-congo: should we set the weight to the default here if it is not null but less than 1? Or do you think it makes more sense to skip this host?

wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
```diff
@@ -215,6 +215,7 @@ PartialPluginService.unexpectedMethodCall=Calling ''{0}()'' is not supported wit
 
 LimitlessConnectionPlugin.failedToConnectToHost=Failed to connect to host {0}.
 LimitlessConnectionPlugin.unsupportedDialectOrDatabase=Unsupported dialect ''{0}'' encountered. Please ensure JDBC connection parameters are correct, and refer to the documentation to ensure that the connecting database is compatible with the Limitless Connection Plugin.
+LimitlessConnectionPlugin.incompatibleWeightedRandomHostWeightPairsProperty=The Limitless Plugin is incompatible with the {0} property. The Limitless plugin uses a list of dynamic hosts and weights through the WeightedRandomHostSelector when choosing a host. Setting this property to a specific value will make the host list static.
```
aaron-congo: Might want to consider getting rid of the last sentence as users might mistakenly believe their static host list will be followed still instead of the dynamic weights, when in reality it is completely rejected and an exception is thrown

wrapper/src/main/java/software/amazon/jdbc/WeightedRandomHostSelector.java
faviansamatha: Good point, I think we shouldn't skip it as weighted round robin doesn't skip it as well if it's not in the host map and give it a default weight of 1. If a user tries to use this and they change an instance name or they don't put down all the hosts, then it should still count.

wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
faviansamatha: Good point!

sergiyvamz (approved):

aaronchung-bitquill (approved):
LGTM

PR #1730, opened 2026-02-17
chore: GDB docs improvements

sergiyvamz:
### Summary

- docs improvements
- **breaking change:** set `gdbRwRestrictWriterToHomeRegion` and `gdbRwRestrictReaderToHomeRegion` to true by default

### DIFF
--- a/docs/using-the-jdbc-driver/HostSelectionStrategies.md
+++ b/docs/using-the-jdbc-driver/HostSelectionStrategies.md
@@ -17,4 +17,4 @@ These strategies are applicable when using the following plugins:
 - [Read/Write Splitting Connection Plugin](./using-plugins/UsingTheReadWriteSplittingPlugin.md)
 - [GDB Read/Write Splitting Connection Plugin](./using-plugins/UsingTheGdbReadWriteSplittingPlugin.md)
 - [GDB Failover Connection Plugin](./using-plugins/UsingTheGdbFailoverPlugin.md)
-- [Failover Plugin](./using-plugins/UsingTheFailover2Plugin.md)
+- [Failover Plugin v2](./using-plugins/UsingTheFailover2Plugin.md)


docs/using-the-jdbc-driver/using-plugins/UsingTheGdbReadWriteSplittingPlugin.md
```diff
@@ -37,11 +37,11 @@ Verify plugin compatibility within your driver configuration using the [compatib
...
+| `readerHostSelectorStrategy`      | String  |                                                                     No                                                                      | The name of the strategy that should be used to select a new reader host. For more information on the available reader selection strategies, see this [table](../HostSelectionStrategies.md).                                                                                                                                                                                                          | `random`                                                                                                                         |
```
karenc-bq, line 40: Should we also add some backlinks in the HostSelectionStrategies file?
At the top we can add a section like this:
```
These strategies are applicable when using the following plugins:
- [Aurora Initial Connection Strategy Plugin](./using-plugins/UsingTheAuroraInitialConnectionStrategyPlugin.md)
- [Read/Write Splitting Connection Plugin](./using-plugins/UsingTheReadWriteSplittingPlugin.md)
- [GDB Read/Write Splitting Connection Plugin](./using-plugins/UsingTheGdbReadWriteSplittingPlugin.md)
- [GDB Failover Connection Plugin](./using-plugins/UsingTheGdbFailoverPlugin.md)
```

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/GdbReadWriteSplittingPlugin.java
```diff
@@ -51,13 +51,13 @@ public class GdbReadWriteSplittingPlugin extends ReadWriteSplittingPlugin implem
   public static final AwsWrapperProperty RESTRICT_WRITER_TO_HOME_REGION =
       new AwsWrapperProperty(
           "gdbRwRestrictWriterToHomeRegion",
-          "false",
+          "true",
```
karenc-bq, line 54: Is this going to be a breaking change

sergiyvamz: It changes default plugin behaviour so it may affect some users. I think the answer is yes.

PR #1732, opened 2026-02-22
Introduce `awsSecretsManager2` plugin

afds:
### Summary

Introduce `awsSecretsManager2` plugin - a Stale-While-Revalidate (SWR) variant of the AWS Secrets Manager connection plugin that eliminates connection-time blocking on credential refresh.

#### Problem

The existing `awsSecretsManager` plugin blocks every connection attempt when cached credentials have expired, waiting for a synchronous fetch from AWS Secrets Manager before proceeding. This causes:
- **Latency spikes** during credential refresh, where all connections stall until a fresh secret is retrieved.
- **Complete connection failures** during transient Secrets Manager outages, even when valid (recently expired) credentials are still cached locally.
- **Thundering herd** under burst traffic - each concurrent connection independently fetches credentials from Secrets Manager.

#### Solution

The `awsSecretsManager2` plugin uses an SWR caching strategy:
- **Cache hit (fresh):** Use cached credentials immediately — same as v1.
- **Cache hit (stale):** Use stale credentials immediately and trigger a single background async refresh. Connections are never blocked.
- **Cache miss (first connection):** Fetch synchronously — same as v1.
- **Login failure with stale credentials:** Automatically performs a synchronous re-fetch and retries the connection.

Thundering herd prevention is built in: `ConcurrentHashMap.compute()` ensures at most one in-flight refresh per `(secretId, region)` key, regardless of how many connections are opened concurrently.

Both plugins share the same credential cache (`AwsSecretsManagerCacheHolder`).

Migration is seamless — just change the plugin from `awsSecretsManager` to `awsSecretsManager2`.

#### Changes

- **New plugin:** `AwsSecretsManagerConnectionPlugin2` with SWR logic.
- **New factory:** `AwsSecretsManagerConnectionPlugin2Factory`.
- **Plugin registration:** Added `awsSecretsManager2` to `ConnectionPluginChainBuilder` and `ConnectionPluginManager`.
- **Validation:** `secretsManagerExpirationTimeSec` is clamped to a minimum of 300s to prevent excessive background refresh triggers.
- **Tests:** Unit tests covering cache hit/miss/stale paths, thundering herd prevention, interrupt handling, force-fetch joining in-flight refreshes, and cleanup verification.
- **Documentation:** Full plugin guide with SWR flow diagram, v1-vs-v2 comparison table, security considerations for stale credentials, migration guide, and telemetry behavior change notes.


docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
```diff
@@ -35,12 +35,19 @@ The following properties are required for the AWS Secrets Manager Connection Plu
...
 | `secretsManagerEndpoint`               | String  |                            No                            | Set this value to be the endpoint override to retrieve your secret from. This parameter value should be in the form of a URL, with a valid protocol (ex. `http://`) and domain (ex. `localhost`). A port number is not required. | `http://localhost:1234` | `null`        |
-| `secretsManagerExpirationSec`          | Integer |                            No                            | This property sets the time in seconds that secrets are cached before it is re-fetched.                                                                                                                                          | `600`                   | `870`         |
+| `secretsManagerExpirationTimeSec`      | Integer |                            No                            | This property sets the time in seconds that secrets are cached before it is re-fetched.                                                                                                                                          | `600`                   | `870`         |
```
sergiyvamz, line 38: That would be a breaking change. If you can, it's better to avoid such renaming.

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
```diff
@@ -35,12 +35,19 @@ The following properties are required for the AWS Secrets Manager Connection Plu
...
 | `secretsManagerSecretPasswordProperty` | String  |                            No                            | Set this value to be the key in the JSON secret that contains the password for database connection.                                                                                                                              | `readerPassword`        | `password`    |
 
 *NOTE* A Secret ARN has the following format: `arn:aws:secretsmanager:<Region>:<AccountId>:secret:SecretName-6RandomCharacters`
 
+## AWS Secrets Manager Connection Plugin v2 (Stale-While-Revalidate)
+An alternative version of this plugin, `awsSecretsManager2`, is available. It uses a **Stale-While-Revalidate (SWR)** caching strategy that connects immediately using stale cached credentials while refreshing them asynchronously in the background. This eliminates connection latency spikes during credential refresh on cache expiry and allows the driver to remain functional during temporary AWS Secrets Manager outages.
```
sergiyvamz, line 45: The SWR strategy is ideal for content that updates occasionally but doesn't require real-time accuracy. Since `awsSecretsManager` deals with passwords that are essential for passing server authentication, will this approach actually/practically bring any improvements?

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin2.md
```diff
@@ -0,0 +1,146 @@
...
+The original `awsSecretsManager` plugin blocks connection requests when cached credentials have expired, waiting for a synchronous fetch from AWS Secrets Manager before proceeding. This can cause:
+- **Connection latency spikes** during credential refresh when the cache expires.
+- **Complete connection failures** if AWS Secrets Manager is temporarily unreachable.
+
+The `awsSecretsManager2` plugin addresses both issues:
+
+| Scenario                                        | `awsSecretsManager` (original)         | `awsSecretsManager2` (SWR)                                      |
+|-------------------------------------------------|----------------------------------------|-----------------------------------------------------------------|
+| Credentials cached and not expired              | Uses cached credentials                | Uses cached credentials                                         |
+| Credentials cached but expired                  | **Blocks** while fetching fresh secret | Connects **immediately** with stale credentials; refreshes async |
```
sergiyvamz, line 19: Connecting with stale credentials can give a failure that will be propagated back to the client application to deal with. Most of the clients prefer the driver does all necessary retries rather to implement it at the application level. Just a thought.

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin2.java
```diff
@@ -0,0 +1,539 @@
...
+ * even if they are stale (expired). When stale credentials are used, an asynchronous
+ * background refresh is triggered. This ensures connections are never blocked waiting
+ * for AWS Secrets Manager, and the driver remains functional during Secrets Manager outages
+ * as long as cached credentials exist.
+ *
+ * <p>On the first connection (cache miss), a synchronous fetch is performed.
+ * If the connection fails with a login error and credentials haven't been freshly fetched,
+ * a synchronous re-fetch and retry is attempted.
+ */
+public class AwsSecretsManagerConnectionPlugin2 extends AbstractConnectionPlugin {
```
sergiyvamz, line 85: Is there any chance to extend this new plugin from the original `AwsSecretsManagerConnectionPlugin`? I see that the most of the code (80% -ish) is the same. Could it be possible to avoid code repetition?

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
afds: this is just an update to documentation to align with the code as it was referring to non-existing property

https://github.com/aws/aws-advanced-jdbc-wrapper/blob/a2c782ca9db7ae7911bb458615de1540e0b886f9/wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java#L96

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
afds: Yes, the main benefit is improved reliability during secret refresh upon expiry.

Secrets Manager rotation is not instantaneous. During rotation there is typically a transition window where both old and new credentials must work. Without SWR, cache expiry can cause sudden authentication failures or a thundering herd of refresh calls.

SWR allows continued use of the last known valid secret while refreshing in the background, reducing auth failures, load spikes, and sensitivity to transient delays.

docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin2.md
afds: That is valid trade-off.

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin2.java
afds: I’d prefer not to extend the original plugin.

While there is overlap right now, the behavior and intent are different, and the implementations may diverge further over time. Forcing inheritance here would tightly couple the two and could make future changes harder to reason about.

Keeping them separate keeps the logic clearer and avoids introducing unnecessary complexity just to reduce some code duplication.

karenc-bq: Thanks for explaining your reasoning. We agree that v2 and v1 has different behaviour and implementation might diverge further in the future.

However, a large portion of the implementation is still the same. From a maintenance and security perspective, we want to avoid code duplication so future bug fixes and security updates do not have to be applied to multiple places.

While the token caching and retrieval logic differ, both v1 and v2 of the Secrets Manager plugin still have same ultimate goal: enable customers to authenticate users via AWS Secrets Manager.
Would you be open to refactoring the shared logic by either:
1. introducing an AbstractSecretsManagerPlugin with the shared common functionality, and have both v1 and v2 implement their own token refresh strategy? We have used a similar approach for our Read Write Splitting plugin.
2. using a composition approach, where we extract the shared Secrets Manager functionality into a helper, like what we are doing with the IAM/ADFS/Okta authentication plugins. This approach would still help keep the token retrieval logic distinct while reducing duplication and improves maintainability.

sergiyvamz:
Hi @afds,

Thank you for this contribution and for the thorough write-up — the Stale-While-Revalidate idea is genuinely valuable, and the problem analysis (connection-time blocking on credential refresh, outage resilience, thundering herd) was spot on.

Since this PR was opened, `main` changed how the Secrets Manager plugin caches credentials (it now uses the internal `StorageService` rather than the `AwsSecretsManagerCacheHolder` this PR built on), so the branch no longer applies cleanly. We've taken your work as the foundation and re-implemented the SWR plugin against the current codebase in #1985, reusing most of the existing `awsSecretsManager` plugin to avoid duplication, and carrying over your SWR semantics, thundering-herd prevention, and test scenarios. Your original PR is credited as the inspiration there.

We're going to close this PR in favor of #1985 — please follow along and feel free to review or comment. Thanks again for moving this forward!

PR #1755, opened 2026-03-03
refactor: introduce an abstract SAML Plugin to reduce code duplication

karenc-bq:
### Summary

Introduce abstract class SamlAuthPlugin to reduce code duplication in okta and adfs plugin.

### Description

Implementation between Federated Auth Plugin and Okta Auth Plugin are mostly the same and can be simplified into a single SamlAuthPlugin.
Similarly, FederatedAuthCacheHolder and OktaAuthCacheHolder have the same implementation, this PR also consolidates these two classes into one AuthCacheHolder class.

These two auth plugins also share a lot of the common connection properties, for instance:
```java
public static final AwsWrapperProperty DB_USER =
      new AwsWrapperProperty("dbUser", null, "The database user used to access the database");
```
Users could access these properties by calling `OktaAuthPlugin.DB_USER.getString(props)` or `FederatedAuthPlugin.DB_USER.getString(props)`. Moving these common connection properties to `SamlAuthPlugin` would cause breaking changes for these users. Therefore, to preserve backwards compatibility, this PR introduces the following abstract methods to retrieve the corresponding connection property:
```
  abstract String getDbUserProperty(Properties props);
  abstract String getIamHostProperty(Properties props);
  abstract int getIamPortProperty(Properties props);
  abstract int getIamTokenExpiration(Properties props);
  abstract void checkIdpCredentialsWithFallback(Properties props);
```


wrapper/src/main/java/software/amazon/jdbc/plugin/federatedauth/SamlAuthPlugin.java
```diff
@@ -0,0 +1,230 @@
...
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.RdsUrlType;
+import software.amazon.jdbc.util.RdsUtils;
+import software.amazon.jdbc.util.RegionUtils;
+import software.amazon.jdbc.util.StringUtils;
+import software.amazon.jdbc.util.telemetry.TelemetryCounter;
+import software.amazon.jdbc.util.telemetry.TelemetryFactory;
+import software.amazon.jdbc.util.telemetry.TelemetryGauge;
+
+public abstract class SamlAuthPlugin extends AbstractConnectionPlugin {
```
sergiyvamz: SamlAuthBasePlugin? SamlAuthAbstractPlugin?

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/federatedauth/SamlAuthPlugin.java
karenc-bq: I like BaseSamlAuthPlugin

wrapper/src/main/java/software/amazon/jdbc/plugin/federatedauth/AuthCacheHolder.java
```diff
@@ -19,11 +19,7 @@
 import java.util.concurrent.ConcurrentHashMap;
 import software.amazon.jdbc.plugin.TokenInfo;
 
-/* The main plugin code OktaAuthPlugin depends on AWS SDK. In order to avoid unnecessary dependencies,
- the plugin cache has been extracted into this OktaAuthCacheHolder class. This cache holder class doesn't depend
- on AWS SDK and can be safely cleared if needed.
- */
-public class OktaAuthCacheHolder {
+public class AuthCacheHolder {
```
karenc-bq, line 22: May consider moving this to utils and update IAM auth plugin to also use this cache holder instead of another IamAuthCacheHolder class

PR #1782, opened 2026-03-17
chore: use ScheduledExecutorService for background connection pruning thread

karenc-bq:
### Summary

PR updating OpenedConnectionTracker to use the existing scheduled executor service for a simpler code 

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
@@ -29,6 +29,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 import software.amazon.jdbc.HostSpec;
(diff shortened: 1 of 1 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/OpenedConnectionTracker.java
```diff
@@ -90,7 +79,7 @@ public void populateOpenedConnectionQueue(final HostSpec hostSpec, final Connect
 
     // Check if the connection was established using an instance endpoint
     if (rdsUtils.isRdsInstance(hostSpec.getHost())) {
-      trackConnection(hostSpec.getHostAndPort(), conn);
+      trackConnection(hostSpec.asAlias(), conn);
```
sergiyvamz, line 82: Do we need this change?

karenc-bq: Not really, asAlias calls getHostAndPort internally. This is just to align with how the rest of this file uses asAlias instead of getHostAndPort

sergiyvamz: What's the side effects of this change?

karenc-bq:
```
 public String asAlias() {
    return this.getHostAndPort();
  }
```
asAlias calls getHostAndPort. They are functionally the same so there shouldn't be any side effect.

sergiyvamz (approved):

PR #1783, opened 2026-03-19
chore: test hibernate against java 17+

karenc-bq:
### Summary

Update Hibernate version to 7.3.0 and run Hibernate ORM tests against Java versions 17, 21, 22 and 24.
Update sample code to use the latest Spring and Hibernate versions.

### Description

This project is built with Java 8, but the latest stable Hibernate version requires at least Java 17.
This PR creates a separate Java 17 source set for the Hibernate integration tests.
This PR also updates `settings.build.kts` to only compile the Spring and Hibernate sample code if the environment is using at least Java 17.

To run the tests with Java versions 17, 21, 22, 24, we had to override two additional test files:
- settings.gradle = the minimum JDK version required to compile the project was set to 25. We had to override it to 17.
- gradle.properties = the original derby version was 10.17.1.0, which requires at least Java 19, causing tests to fail against Java 17.


wrapper/src/test/java/integration/host/TestEnvironmentProvider.java
```diff
@@ -136,10 +136,6 @@ public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContex
               if (jvm == TargetJvm.OPENJDK24 && config.noOpenJdk24) {
                 continue;
               }
-              if (jvm != TargetJvm.OPENJDK24 && config.testHibernateOnly) {
```
sergiyvamz, line 139: Do you want to change it to Java17 instead or removing this entire condition?

sergiyvamz (approved):

PR #1901, opened 2026-05-07
chore: run integration tests by schedule every workday

sergiyvamz:
### Summary

Run integration tests daily by schedule. To minimize quota effect, workflows run sequentially with a few hours shift.

- `run-integration-tests-default` 8:00 PM PST, Mon–Fri
- `run-integration-tests-latest` 12:30 AM PST Tue–Sat
- `run-integration-tests-mysql-aurora-java-lts` 5:00 AM PST Tue–Sat

### DIFF
--- a/.github/workflows/run-integration-tests-default.yml
+++ b/.github/workflows/run-integration-tests-default.yml
@@ -2,10 +2,10 @@ name: Run Aurora Integration Tests Default
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - main
-      - main-2.x
+  schedule:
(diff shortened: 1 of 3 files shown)


.github/workflows/run-integration-tests-default.yml
```diff
@@ -2,10 +2,10 @@ name: Run Aurora Integration Tests Default
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - main
-      - main-2.x
+  schedule:
```
sophia-bq, line 5: Maybe we should have some team discussion on switching to a daily schedule running these instead of per PR?

PR #1902, opened 2026-05-07
chore: separate tests with Aurora 5-instances configuration into new workflows

sergiyvamz:

### DIFF
--- a/.github/workflows/run-integration-tests-5-instance-default.yml
+++ b/.github/workflows/run-integration-tests-5-instance-default.yml
@@ -1,24 +1,20 @@
-name: Run Aurora Integration Tests Default
+name: Run Aurora 5-Instance Integration Tests Default
 
 on:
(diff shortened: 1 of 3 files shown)


.github/workflows/run-integration-tests-5-instance-default.yml
```diff
@@ -0,0 +1,73 @@
+name: Run Aurora 5-Instance Integration Tests Default
+
+on:
+  workflow_dispatch:
```
sophia-bq, line 4: What is the plan for scheduling when these will run?

sergiyvamz: There's no plan to schedule it. 5-instance clusters shows the same behaviour as 3-instance clusters so running 5-instance cluster tests doesn't bring much value. Dev can run this test on demand.

PR #1905, opened 2026-05-09
fix: added allowlist classes for CacheResultSet

Frank-Gu-81:
### Summary

Fix unsafe Java deserialization issue in RemoteQueryCachePlugin

### Description

**Fix:** Added `SafeObjectInputStream` that overrides `resolveClass()` with a class allowlist. Non-allowlisted classes are rejected before instantiation, blocking gadget chains. Compatible with all Java 8+ (no `ObjectInputFilter` dependency). Unknown types gracefully fall back to cache miss.

**Files changed:**
- `CachedResultSet.java` — added `SafeObjectInputStream`, replaced `ObjectInputStream` at both deserialization points
- `aws_advanced_jdbc_wrapper_messages.properties` — added blocked deserialization error message

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
@@ -63,6 +63,7 @@ import java.util.Set;
 import java.util.TimeZone;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.WrapperUtils;
 
 public class CachedResultSet implements ResultSet {
 
@@ -193,6 +194,12 @@ public class CachedResultSet implements ResultSet {
             || ZoneId.class.isAssignableFrom(cls)) {
           return cls;
         }
+      }
+      // Allow user-registered third-party classes and packages (via Driver.skipWrappingForType
+      // or Driver.skipWrappingForPackage). See security note in UsingTheJdbcDriver.md.
+      if (WrapperUtils.skipWrappingForClasses.stream().anyMatch(c -> c.getName().equals(className))
+          || WrapperUtils.skipWrappingForPackages.stream().anyMatch(p -> className.startsWith(p + "."))) {
+        return super.resolveClass(desc);
       }
       throw new ClassNotFoundException(
           Messages.get("CachedResultSet.blockedDeserialization", new Object[]{className}));
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
@@ -50,8 +50,11 @@ import java.time.OffsetDateTime;
 import java.time.OffsetTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalAmount;
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


Frank-Gu-81:
cc: @QuChen88

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
```diff
@@ -103,6 +107,126 @@ public Object get(final int columnIndex) throws SQLException {
...
+      allowed.add("java.util.Collections$EmptyMap");
+      allowed.add("java.util.Collections$EmptySet");
+      allowed.add("java.util.Collections$SingletonList");
+      allowed.add("java.util.Collections$SingletonMap");
+      allowed.add("java.util.Collections$SingletonSet");
+      allowed.add("java.util.Collections$UnmodifiableList");
+      allowed.add("java.util.Collections$UnmodifiableMap");
+      allowed.add("java.util.Collections$UnmodifiableSet");
+      allowed.add("java.util.Collections$UnmodifiableRandomAccessList");
+      allowed.add("java.util.Arrays$ArrayList");
```
QuChen88: How do you know if this is the exhaustive list of classes we can have for collection object types and we didn't miss something?

Also going forward for each incremental Java version, if a new subtype of say List or Set comes out, do we need to update this list?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
```diff
@@ -103,6 +107,126 @@ public Object get(final int columnIndex) throws SQLException {
...
+  /**
+   * A restricted ObjectInputStream that only allows deserialization of known-safe classes.
+   * This prevents Remote Code Execution via cache poisoning attacks where an attacker
+   * injects malicious serialized objects (gadget chains) into the remote cache.
+   *
+   */
+  private static class SafeObjectInputStream extends ObjectInputStream {
+    private static final Set<String> ALLOWED_CLASSES;
+
+    static {
```
QuChen88, line 123: I am not sure how we can ensure the object types are exhaustive and we didn't miss anything.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
Frank-Gu-81: With the current static allowlist approach, we do have to manually maintain the list. If new standard library types appear in future JDK versions, we'd need to update it. Another limitation is that it will always reject all unlisted classes (i.e. custom classes) from a security standpoint.

To address the collection enumeration concern, I can add a dynamic fallback in resolveClass: after the static allowlist check, additionally permit any class whose package starts with `java.util` and implements `java.util.Collection` or `java.util.Map`. This should cover all current and future JDK collection subtypes without manual enumeration, while still blocking arbitrary third-party classes.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
Frank-Gu-81: Maybe I can add a dynamic check in addition like this?

```
@Override
protected Class<?> resolveClass(ObjectStreamClass desc)
    throws IOException, ClassNotFoundException {
  String className = desc.getName();
  if (ALLOWED_CLASSES.contains(className)) {
    return super.resolveClass(desc);
  }
  if (className.startsWith("java.util.")) {
    Class<?> cls = super.resolveClass(desc);
    if (Collection.class.isAssignableFrom(cls) || Map.class.isAssignableFrom(cls)) {
      return cls;
    }
  }
  throw new ClassNotFoundException(
      Messages.get("CachedResultSet.blockedDeserialization", new Object[]{className}));
}
```

QuChen88: In principle, all Java object types are hierarchical. This gives us room for simplification from a hard-coded static list of all possible class types that we can de-serialize from raw data. One thing which comes to my mind is that you can  track only the set of base classes of the various data types we support, and then do the `superClass.isAssignableFrom(subClass)` check as you mentioned. 

Also this approach makes sense as the de-serialization logic will fail fast if we try to de-serialize objects with types we don't recognize. In the future if we add more data types support such as CLOB / BLOB / Array, we need to expand this list explicitly in the code. 

I just did a quick pass through our code and I came up with the following trimmed down list - 

- All primitive types: byte, double, float, int, long, and short. 
- java.lang.Number: superclass of platform classes representing numeric values that are convertible to primitive types like Double / Float / Integer / Long / Short / BigDecimal etc
- java.lang.Boolean
- java.lang.String
- byte[] : Please check how this works and ensure unit tests still pass
- java.util.Collection : Handles List and Set
- java.util.Map
- java.util.Date: Handles basic Date, DateTime and Timestamp
- java.time.temporal.Temporal: this is the base interface type for date, time and offset objects that are complete enough to be manipulated using plus and minus, such as LocalDate/LocalDateTime/LocalTime/OffsetDateTime/OffsetTime/ZonedDateTime
- java.net.URL
- java.sql.RowId
- java.sql.SQLXML

Please check to see if I missed something here.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
QuChen88: See my comment below.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
```diff
@@ -201,7 +325,7 @@ public byte[] serializeIntoByteArray() throws SQLException {
    */
   public static ResultSet deserializeFromByteArray(byte[] data) throws SQLException {
     try (ByteArrayInputStream bis = new ByteArrayInputStream(data);
-         ObjectInputStream ois = new ObjectInputStream(bis)) {
+         ObjectInputStream ois = new SafeObjectInputStream(bis)) {
       CachedResultSetMetaData metadata = (CachedResultSetMetaData) ois.readObject();
```
QuChen88, line 308: Can we add a unit test to cover the case when `ois.readObject()` actually produced an object of type other than `CachedResultSetMetadata` and verify this method throws exception?

QuChen88 (commented):
Left a few comments

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
Frank-Gu-81: Yes, will add a test case for that.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
Frank-Gu-81: I think we still need a few classes from the allowlist on top of this trimmed one:
1. Our own classes: `CachedResultSetMetaData`, `CachedResultSetMetaData$Field`, `CachedSQLXML`
2. `ZoneId`, `ZoneOffset`, `Duration`, and `Period` are not `Temporal`, so I think I need to check against `java.time.` for broader coverage.
3. Some other common classes that might be serialized/deserialized: `UUID`, `URI`, `URL`, `Character`

QuChen88: 1. `CachedSQLXML` implements `SQLXML`, so `isAssignableFrom()` check should work. No need to explicitly declare it in the list. Agree about adding `CachedResultSetMetaData`. 
2. Do we really store and de-serialize them from the database column data? Or are they only used in our internal logic to process this data in `getXXX()`? Probably run a few quick tests and see what gets stored in the cache for a sample table that contains timestamp columns. 
3. Sounds fair to me. 

I am mainly trying to limit the size of this list and make it comprehensive so that it can be easily maintained and reasoned about going forward.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
QuChen88: You can take the input string from the security RCE and call `deserializeFromByteArray()` on that as input in your unit test

Frank-Gu-81: I believe using the exact PoC bytes would require adding a gadget library as a test dependency, which we don't want to do. I can addd a simpler version of it that would prove the same mechanism.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
Frank-Gu-81: > 2. Do we really store and de-serialize them from the database column data? Or are they only used in our internal logic to process this data in getXXX()? Probably run a few quick tests and see what gets stored in the cache for a sample table that contains timestamp columns.

Confirmed that we do need them during de-serialization. I added a test case that round-trips all timestamp-related types (`LocalDateTime`, `OffsetDateTime`, `ZonedDateTime`, `OffsetTime`, `LocalDate`) through serialize → deserialize.If I comment out the `java.time.*` prefix check, the test fails with:
```
ClassNotFoundException: Deserialization of class [java.time.Ser] is not allowed.
```

This proves that allowlisting individual ones doesn't work. We also need to allowlist the internal proxy.

QuChen88: Please don't hard code `java.time.*` in the check. It would be good idea to check for the specific base class instead.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedResultSet.java
```diff
@@ -103,6 +110,93 @@ public Object get(final int columnIndex) throws SQLException {
...
+      if (ALLOWED_CLASSES.contains(className)) {
+        return super.resolveClass(desc);
+      }
+      // Only load classes from standard Java/javax packages for dynamic hierarchy checks.
+      // Never call super.resolveClass for third-party or application classes — loading them
+      // before throwing gives the JVM a window to invoke readObject() on a partially
+      // constructed object in some implementations.
+      if (className.startsWith("java.") || className.startsWith("javax.")) {
+        Class<?> cls = super.resolveClass(desc);
+        if (Number.class.isAssignableFrom(cls)
```
QuChen88, line 186: Add a small comment here indicating this is the list of data types we support for de-serialization today. 

i.e. if in the future we want to support more types, we need to add to this list.

Frank-Gu-81: added

QuChen88 (approved):

sergiyvamz (approved):

Issue #1896, opened 2026-05-07
NPE in `efm2` after upgrading to 4.0.0


bahaa:
### Describe the bug

After upgrading from 3.3.0 to 4.0.0, `efm2` plugin throws NullPointerException.

### Expected Behavior

No Exception is thrown as the upgrade is backwards compatible. 

### What plugins are used? What other connection properties were set?

failover2,efm2

### Current Behavior

NPE is thrown with this root cause:
```
Caused by: java.lang.NullPointerException: Cannot invoke "software.amazon.jdbc.plugin.efm.v2.HostMonitorServiceV2Impl$HostMonitorKey.toString()" because "this.monitorKey" is null
	at software.amazon.jdbc.plugin.efm.v2.HostMonitorServiceV2Impl.getSnapshotState(HostMonitorServiceV2Impl.java:188) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.util.WrapperUtils.addSnapshotState(WrapperUtils.java:839) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.util.WrapperUtils.addSnapshotState(WrapperUtils.java:826) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.plugin.efm.base.HostMonitoringConnectionBasePlugin.getSnapshotState(HostMonitoringConnectionBasePlugin.java:289) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.ConnectionPluginManager.getSnapshotState(ConnectionPluginManager.java:638) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.util.WrapperUtils.addSnapshotState(WrapperUtils.java:839) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.util.WrapperUtils.addSnapshotState(WrapperUtils.java:826) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
	at software.amazon.jdbc.util.WrapperUtils.collectState(WrapperUtils.java:754) ~[aws-advanced-jdbc-wrapper-4.0.0.jar:na]
[...]

PR #1907, opened 2026-05-11
fix: issue with localhost in efm/efm2 plugin

sergiyvamz:
### Summary

Fixes an issue where efm/efm2 plugins throw a RuntimeException ("Unable to identify the given connection") when used with non-RDS URLs such as localhost, custom hostnames, or IP addresses.
https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1896

### Description

When identifyConnection() returns null, keep the original monitoringHostSpec (from getCurrentHostSpec() or getRoutedHostSpec()) instead of throwing a RuntimeException. This allows EFM to monitor the connection using the host from the connection string.

- Preserves the design intent: all non-RDS_INSTANCE connections still attempt identification. RDS cluster endpoints continue to be resolved correctly.
- Graceful degradation: when identification fails for non-RDS URLs, the plugin uses the connection string host — which is valid for monitoring purposes.
- No behavioural change for RDS users.
- Fixes both efm and efm2 since they share the same base class.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/base/HostMonitoringConnectionBasePlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/efm/base/HostMonitoringConnectionBasePlugin.java
@@ -262,14 +262,11 @@ public abstract class HostMonitoringConnectionBasePlugin extends AbstractConnect
           this.monitoringHostSpec = this.pluginService.identifyConnection(
               this.pluginService.getCurrentConnection(), this.pluginService.getCurrentHostSpec());
           if (this.monitoringHostSpec == null) {
-            throw new RuntimeException(Messages.get(
-                "HostMonitoringConnectionPlugin.unableToIdentifyConnection",
-                new Object[] {
-                    this.pluginService.getCurrentHostSpec().getHost(),
-                    this.pluginService.getHostListProvider()}));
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/plugin/efm/base/HostMonitoringConnectionBasePlugin.java
```diff
@@ -262,14 +262,13 @@ public HostSpec getMonitoringHostSpec() {
           this.monitoringHostSpec = this.pluginService.identifyConnection(
               this.pluginService.getCurrentConnection(), this.pluginService.getCurrentHostSpec());
           if (this.monitoringHostSpec == null) {
-            throw new RuntimeException(Messages.get(
-                "HostMonitoringConnectionPlugin.unableToIdentifyConnection",
-                new Object[] {
-                    this.pluginService.getCurrentHostSpec().getHost(),
-                    this.pluginService.getHostListProvider()}));
+            this.monitoringHostSpec = this.pluginService.getRoutedHostSpec() != null
```
sophia-bq: This is exactly what we do on line 253. Would it be better to have the result of identify connection be stored in a variable so we do not assign monitoringHostSpec to a value, then null, then the same value again in these cases? Will probably simplify the fix

karenc-bq (approved):

PR #1914, opened 2026-05-12
fix: make static executors recreatable after releaseResources()

sergiyvamz:
### Summary

PR #1906 introduced `releaseResources()` methods that shut down static `ExecutorService` instances in `OpenedConnectionTracker` and `BatchingEventPublisher`. These executors were declared as static final and initialized once at class load time. Once shut down, they could not be restarted within the same JVM.

In integration tests, `TestDriverProvider.beforeEach()` calls `Driver.releaseResources()` before each test, which permanently killed the executors. When subsequent tests triggered failover and the driver attempted to invalidate idle connections via `OpenedConnectionTracker.invalidateAllConnections()`, the task was submitted to a shut-down executor and silently discarded. This caused `testServerFailoverWithIdleConnections` to fail consistently across all Aurora environments (MySQL, MariaDB, PostgreSQL) in both 2-instance and 3-instance configurations.

### Description

Changed static executors from eagerly-initialized static final fields to lazily-initialized static volatile fields that are recreated on demand after shutdown:

**OpenedConnectionTracker**:
- Removed the static initializer block and eager executor creation
- Executor fields start as null and are created lazily
- `getOrCreatePruneExecutor()` — creates the prune executor and schedules the prune task on first use (triggered when tracking a connection)
- `getOrCreateInvalidateExecutor()` — creates the invalidate executor on first use (triggered when invalidating connections)
`releaseResources()` — shuts down both executors and clears tracked connections; next usage recreates them
- Uses double-checked locking with `ResourceLock` for thread safety

**BatchingEventPublisher**:
- `getOrCreatePublishingExecutor()` — recreates the executor if it was shut down
- `ensureExecutorRunning()` — called from publish() to detect a shut-down executor and reinitialize the publishing thread
- `releaseResources()` — shuts down the executor; next usage recreates it

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/events/BatchingEventPublisher.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/events/BatchingEventPublisher.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import software.amazon.jdbc.util.ExecutorFactory;
+import software.amazon.jdbc.util.ResourceLock;
 
 /**
  * An event publisher that periodically publishes a batch of all unique events encountered during the latest time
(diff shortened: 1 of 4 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/events/BatchingEventPublisher.java
```diff
@@ -50,14 +53,29 @@ public BatchingEventPublisher() {
...
 
   protected void initPublishingThread(long messageIntervalNanos) {
-    publishingExecutor.scheduleAtFixedRate(
+    getOrCreatePublishingExecutor().scheduleAtFixedRate(
         this::sendMessages, messageIntervalNanos, messageIntervalNanos, TimeUnit.NANOSECONDS);
   }
 
+  private static ScheduledExecutorService getOrCreatePublishingExecutor() {
+    ScheduledExecutorService executor = publishingExecutor;
+    if (executor.isShutdown()) {
```
karenc-bq, line 67: Do we need null checks in this method too? i.e. `executor == null || executor.isShutdown()`

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/util/events/BatchingEventPublisher.java
sergiyvamz: We never set executors to null and they are initialized in static constructors. So there's no chance for null pointer here.

PR #1916, opened 2026-05-13
chore: use secure connection for MariaDb driver in integration tests

sergiyvamz:
### Summary

Fix MariaDB driver integration test failures caused by require_secure_transport=ON

### Description

The mysql-aurora-mariadb-driver integration test job has been timing out at the 6-hour limit. Root cause analysis of the raw job logs revealed that Aurora MySQL test instances now have `require_secure_transport=ON` enabled (a recent RDS default change). The MariaDB driver's health check connections in `AuroraTestUtility.makeSureInstancesUp` don't use SSL, causing every connection attempt to be rejected with:

Connections using insecure transport are prohibited while `--require_secure_transport=ON`.
This results in an infinite retry loop in the test setup (`beforeEach`), with each test spending ~28 minutes retrying before timing out. The job produced 6,353 occurrences of this error before being killed.

The MySQL Connector/J driver was unaffected because it defaults to `sslMode=PREFERRED`.

### Changes

**Client-side: Enable SSL for MariaDB driver connections**
- `ConnectionStringHelper.getDefaultProperties()` — Added `sslMode=trust` to the MariaDB driver's default connection properties. This covers all connections made via Properties objects (including `makeSureInstancesUp`).
- `DriverHelper.getDriverRequiredParameters()` — Added `sslMode=trust` to the MariaDB driver's URL parameters (`?permitMysqlScheme&sslMode=trust`). This covers connections made with just URL + username/password.

**Server-side: Create test clusters with require_secure_transport=OFF**
- `AuroraTestUtility.createMysqlClusterParameterGroup()` — New method that creates a cluster parameter group with `require_secure_transport=OFF` for MySQL Aurora test clusters.
- `TestEnvironment` — Modified the AURORA cluster creation path (both initial and retry) to create a MySQL-specific parameter group for all MySQL Aurora clusters when no custom parameter group is already set (e.g., by Blue/Green Deployment tests).
- `TestEnvironment` cleanup — Updated teardown logic to delete the MySQL parameter group when the test environment is destroyed.

### DIFF
--- a/wrapper/src/test/java/integration/DriverHelper.java
+++ b/wrapper/src/test/java/integration/DriverHelper.java
@@ -181,7 +181,7 @@ public class DriverHelper {
   public static String getDriverRequiredParameters(
       DatabaseEngine databaseEngine, TestDriver testDriver) {
     if (testDriver == TestDriver.MARIADB && databaseEngine == DatabaseEngine.MYSQL) {
-      return "?permitMysqlScheme";
+      return "?permitMysqlScheme&sslMode=trust";
     }
     return "";
   }
--- a/wrapper/src/test/java/integration/DriverHelper.java
+++ b/wrapper/src/test/java/integration/DriverHelper.java
@@ -181,7 +181,6 @@ public class DriverHelper {
   public static String getDriverRequiredParameters(
       DatabaseEngine databaseEngine, TestDriver testDriver) {
     if (testDriver == TestDriver.MARIADB && databaseEngine == DatabaseEngine.MYSQL) {
-      //return "?permitMysqlScheme&sslMode=trust";
       return "?permitMysqlScheme";
     }
     return "";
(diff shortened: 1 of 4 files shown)


wrapper/src/test/java/integration/container/ConnectionStringHelper.java
```diff
@@ -226,6 +226,10 @@ public static Properties getDefaultProperties() {
       // will get the error "RSA public key is not available client side" when connecting. The mariadb driver may not
       // fully support mysql 8.4's SSL mechanisms, which is why this property is only required for newer mysql versions.
       props.setProperty("allowPublicKeyRetrieval", "true");
+      // Enable SSL for the MariaDB driver to satisfy Aurora MySQL instances that have
+      // require_secure_transport=ON. Using "trust" mode enables SSL without certificate verification,
+      // which is appropriate for test environments.
+      //props.setProperty("sslMode", "trust");
```
sophia-bq: If we are not intending to set this property we can remove the whole change from this file?

wrapper/src/test/java/integration/DriverHelper.java
```diff
@@ -181,6 +181,7 @@ public static String getDriverRequiredParameters(TestDriver testDriver) {
   public static String getDriverRequiredParameters(
       DatabaseEngine databaseEngine, TestDriver testDriver) {
     if (testDriver == TestDriver.MARIADB && databaseEngine == DatabaseEngine.MYSQL) {
+      //return "?permitMysqlScheme&sslMode=trust";
```
sophia-bq: Same comment

sophia-bq (approved):

Issue #1800, opened 2026-03-26
ConfigurationProfile plugin list ignored; createMinimalServiceContainer falls back to DEFAULT_PLUGINS


ivankhoosty:
### Describe the bug

NOTE: this bug report is entirely generated by Copilot (opus 4.6) , which actually found the bug by analysing the code , given specific symptoms: 3 mins old pods manage in-region failover fine, but 20 mins old pods fail with authentication errors; workaround validated and works as predicted by the gizmo

The AWS Advanced JDBC Wrapper offers two ways to configure the plugin list:

1. **`ConfigurationProfile`** (recommended for HikariCP) — defines plugins programmatically via `ConfigurationProfileBuilder.withPluginFactories()`
2. **`wrapperPlugins` property** — defines plugins as a comma-separated string in connection properties

When using `ConfigurationProfile`, the main connection path correctly reads the plugin list from the profile. However, **internal components** (`ClusterTopologyMonitorImpl` → `NodeMonitoringWorker`) create their own service containers via `ServiceUtility.createMinimalServiceContainer()`, which passes `null` for the `configurationProfile` parameter. This causes `ConnectionPluginChainBuilder.getPlugins()` to ignore the profile entirely and fall back to `DEFAULT_PLUGINS` (`"initialConnection,auroraConnectionTracker,failover2,efm2"`).

**The fundamental flaw:** `ConfigurationProfile` and `wrapperPlugins` are two parallel configuration paths, but only the `wrapperPlugins` property path is respected by internal components. If you configure plugins via `ConfigurationProfile` (the recommended approach for HikariCP), internal monitoring threads silently get a completely different plugin stack — `DEFAULT_PLUGINS` — with no warning. Any plugin added to the profile but not in `DEFAULT_PLUGINS` (such as `iam`, `gdbFailover`, `gdbReadWriteSplitting`) is silently dropped for monitoring workers.

The most visible consequence is with **IAM authentication**: monitoring workers initially authenticate using the cached IAM token from properties, but after the token TTL expires (~14.5 minutes), they have no `iam` plugin to refresh it and die permanently with a `RuntimeException` on login failure.

### Expected Behavior

`NodeMonitoringWorker` threads should inherit the IAM authentication plugin from the parent connection's `ConfigurationProfile`, so they can refresh IAM tokens and survive beyond the token TTL.

### What plugins are used? What other connection properties were set?

[...]

PR #1919, opened 2026-05-14
fix: propagate configuration profile to PartialPluginService

sergiyvamz:
### Summary

https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1800

`createMinimalServiceContainer` in `ServiceUtility.java` was not properly propagating the `ConfigurationProfile` to the `PartialPluginService` or the service container itself. This caused internal monitoring workers (e.g., `NodeMonitoringWorker`) to lose access to profile-configured plugins like IAM, leading to permanent authentication failures after the IAM token TTL (~15 minutes).

### Description

When using `ConfigurationProfile`, the main connection path correctly reads plugins from the profile. However, `ClusterTopologyMonitorImpl.getNodeMonitoringWorker()` creates service containers via `createMinimalServiceContainer`, which:
- Constructed `PartialPluginService` with the 6-arg constructor that delegates to the 8-arg constructor with configurationProfile = null
- Never called `setConfigurationProfile()` on the container

This meant monitoring workers had no access to profile-dependent plugin behaviour (e.g., IAM token refresh), causing them to die permanently once the cached token expired.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/util/ServiceUtility.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/util/ServiceUtility.java
@@ -25,6 +25,7 @@ import software.amazon.jdbc.PartialPluginService;
 import software.amazon.jdbc.PluginServiceImpl;
 import software.amazon.jdbc.dialect.Dialect;
 import software.amazon.jdbc.dialect.HostListProviderSupplier;
+import software.amazon.jdbc.exceptions.ExceptionManager;
 import software.amazon.jdbc.hostlistprovider.HostListProvider;
 import software.amazon.jdbc.profile.ConfigurationProfile;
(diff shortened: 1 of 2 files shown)


wrapper/src/main/java/software/amazon/jdbc/util/ServiceUtility.java
```diff
@@ -117,18 +118,22 @@ public FullServicesContainer createMinimalServiceContainer(
 
     PartialPluginService pluginService = new PartialPluginService(
```
sophia-bq, line 119: We should be able to safely remove the shorter PartialPluginService constructor (the one without ExceptionManager and ConfigurationProfile parameters) as this was the only use of it in the codebase

sophia-bq (approved):

PR #1934, opened 2026-05-22
chore: fix compile warnings

sergiyvamz:


wrapper/src/main/java/software/amazon/jdbc/plugin/cache/CachedSQLXML.java
```diff
@@ -118,7 +119,7 @@ public <T extends Source> T getSource(Class<T> sourceClass) throws SQLException
       }
 
       if (SAXSource.class.equals(sourceClass)) {
-        XMLReader reader = XMLReaderFactory.createXMLReader();
+        XMLReader reader = SAXParserFactory.newInstance().newSAXParser().getXMLReader();
```
sergiyvamz, line 122: @QuChen88 How does it look to you?

QuChen88: Looks good to me. If the unit test `CachedSQLXMLTest.java` passes then I am good.

PR #1969, opened 2026-06-12
Feat: LowestLoadHostSelector + HighestLoadHostSelector

aaronchung-bitquill:
### Summary

Introducing Lowest Load Host Selector and Highest Load Host Selector
Deprecating Highest Weight Host Selector

### Description

- Based initially on `davecramer:lowest_load_strategy`  as seen in https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1950.

- Differences here are keeping cpu, and lag on the HostSpec. 

- Calculating load at the host selector level with configurable weights on cpu and lag metrics.

- Lowest and Highest Load Host Selectors have been simplified.


wrapper/src/main/java/software/amazon/jdbc/util/events/TopologyRefreshedEvent.java
```diff
@@ -0,0 +1,65 @@
...
+package software.amazon.jdbc.util.events;
+
+import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Event published after a topology monitor successfully refreshes its cached cluster topology. Subscribers can use this
+ * to invalidate any per-host state that should be recomputed once fresh load/availability data is available.
+ */
+public class TopologyRefreshedEvent implements Event {
```
sergiyvamz: Do we still need this event?

wrapper/src/main/java/software/amazon/jdbc/HighestWeightHostSelector.java
```diff
@@ -25,6 +25,7 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.jdbc.hostavailability.HostAvailability;
 
+@Deprecated
```
sergiyvamz: This deprecation will probably cause us (developers) more comfortless rather than notify users about its deprecation. We just want to use doc to inform our users that this host selector is deprecated.

wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
```diff
@@ -36,13 +31,17 @@ public class HostSpec {
...
+  public static final float UNKNOWN_CPU_PERCENT = -1.0f;
+  public static final float UNKNOWN_LAG_MS = -1.0f;
 
   protected final String host; // full domain name
   protected final int port;
   protected final HostRole role;
   protected final Timestamp lastUpdateTime;
   protected volatile HostAvailability availability;
   protected long weight; // Greater or equal 0. Lesser the weight, the healthier node.
+  protected volatile Float cpuPercent;
```
sergiyvamz: Why volatile?

wrapper/src/main/java/software/amazon/jdbc/HighestLoadHostSelector.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public HostSpec getHost(
+      @NonNull final List<HostSpec> hosts,
+      @Nullable final HostRole role,
+      @Nullable final Properties props) throws SQLException {
+
+    final List<HostSpec> eligible = hosts.stream()
+        .filter(h -> (role == null || role.equals(h.getRole()))
+            && !(h.getCpuPercent() == null && h.getLagMs() == null)
```
sergiyvamz: Why we want to skip records with null cpu and lag? The formular below should handle that properly.

wrapper/src/main/java/software/amazon/jdbc/HostSpec.java
```diff
@@ -154,6 +168,28 @@ public void setWeight(long weight) {
     }
   }
 
+  public Float getCpuPercent() {
+    return this.cpuPercent;
+  }
+
+  public void setCpuPercent(Float cpuPercent) {
```
sergiyvamz: We probably need to remove setters. HostSpec is supposed to be immutable.

sergiyvamz (approved):

PR #1977, opened 2026-06-15
fix: distribute concurrent initial reader connections across readers (#1968)

### LINKED ISSUE #1968: AuroraInitialConnectionStrategyPlugin roundRobin distributes sequential connections correctly but sends concurrent/prefill connections to a single reader
### Environment

- AWS Advanced JDBC Wrapper: 4.0.1
- Database: Aurora PostgreSQL
- Aurora Topology:
  - 1 Writer
  - 2 Readers
- Application Server: WildFly / IronJacamar
- OS: Linux

Configuration:

```java

props.setProperty("user", "<user>");
props.setProperty("password", "<password>");

props.setProperty(
    "wrapperPlugins",
    "initialConnection");

props.setProperty(
    "readerInitialConnectionHostSelectorStrategy",
    "roundRobin");
```

Connection URL tested:

```text
jdbc:aws-wrapper:postgresql://<aurora-reader-endpoint>:5432/<db>
```

Also tested with:

```text
jdbc:aws-wrapper:postgresql://<aurora-cluster-endpoint>:5432/<db>
```

Observed behavior is identical for both endpoints.

---

### Sequential Connection Test

Test code:

```java
for (int i = 1; i <= 20; i++) {

    try (Connection conn =
             DriverManager.getConnection(url, props);
         Statement stmt =
             conn.createStatement();
         ResultSet rs =
             stmt.executeQuery(
                 "SELECT aurora_db_instance_identifier(), inet_server_addr()")) {

        if (rs.next()) {

            System.out.println(
                "Connection " + i +
                " -> " +
                rs.getString(1) +
                " (" +
                rs.getString(2) +
                ")");
        }
    }

    Thread.sleep(1000);
}
```

Observed output:

```text
Connection 1  -> reader2
Connection 2  -> reader1
Connection 3  -> reader2
Connection 4  -> reader1
Connection 5  -> reader2
Connection 6  -> reader1
...
```

Result:

```text
Reader1 = 10
Reader2 = 10
```

RoundRobin appears to work correctly for sequential connection creation.

---

### Concurrent Connection Test

Test configuration:

```java
THREADS = 20;
TOTAL_CONNECTIONS = 20;
```

All threads are released simultaneously using a CountDownLatch.

Each thread executes:

```java
Connection conn =
    DriverManager.getConnection(url, props);

SELECT aurora_db_instance_identifier();
```

Observed result:

```text
Reader1 = 20
Reader2 = 0
```

All connections consistently land on the same reader.

---

### WildFly / IronJacamar Pool Prefill Scenario

Datasource configuration:

```xml
<pool>
    <min-pool-size>5</min-pool-size>
    <max-pool-size>20</max-pool-size>
    <prefill>true</prefill>
</pool>
```

Observed after server startup:

```text
Reader1 = 5
Reader2 = 0
```

All prefilled connections are established against the same Aurora reader instance.

---

### Additional Findings

Wrapper debug logs show topology discovery is working correctly.

Discovered topology:

```text
Topology:
  reader1 -> READER
  reader2 -> READER
  writer  -> WRITER
```

The driver is aware of both reader instances.

The issue appears only when multiple connections are created concurrently.

---

### Unsupported Strategy Test

Changing strategy to:

```java
readerInitialConnectionHostSelectorStrategy=leastConnections
```

throws:

```text
java.lang.UnsupportedOperationException:
Unsupported host selection strategy 'leastConnections'
```

which confirms the InitialConnection plugin is active and processing the strategy property.

---

### Reproduction Summary

| Scenario | Result |
|-----------|---------|
| Sequential (20 connections) | Perfect Reader1/Reader2 alternation |
| Concurrent (20 threads, 20 connections) | All connections routed to Reader1 |
| WildFly Prefill (5 connections) | All connections routed to Reader1 |

---

### Questions

1. Is `roundRobin` expected to distribute connections during concurrent connection creation scenarios?
2. Is there a known limitation where multiple simultaneous connection requests can select the same reader?
3. Is this behavior expected for connection pool initialization/prefill scenarios?
4. Is there a recommended configuration to ensure balanced reader utilization during startup and connection pool growth?
5. Could this be a thread-safety or synchronization issue within the `AuroraInitialConnectionStrategyPlugin` host selection logic?

---

### Expected Behavior

We would expect concurrent connection creation and connection pool prefill scenarios to distribute connections across available readers similarly to the sequential connection test, rather than routing all connections to a single reader.

wrapperPlugins=initialConnection, readerInitialConnectionHostSelectorStrategy=roundRobin

### Current Behavior

With `wrapperPlugins=initialConnection` and `readerInitialConnectionHostSelectorStrategy=roundRobin`:

- Sequential connection creation distributes connections evenly across both Aurora reader instances (expected behavior).
- Concurrent connection creation (e.g., 20 simultaneous connection requests) routes all connections to a single reader instance.
- WildFly/IronJacamar pool prefill (`min-pool-size=5`, `prefill=true`) also creates all initial connections on a single reader instance.

Observed example:

```text
Sequential:
Reader1 = 10
Reader2 = 10

Concurrent:
Reader1 = 20
Reader2 = 0

Prefill:
Reader1 = 5
Reader2 = 0

### Reproduction Steps

1. Create an Aurora PostgreSQL cluster with:
   - 1 Writer instance
   - 2 Reader instances

2. Use AWS Advanced JDBC Wrapper version 4.0.1.

3. Configure the connection properties:

```java
props.setProperty(
    "wrapperPlugins",
    "initialConnection");

props.setProperty(
    "readerInitialConnectionHostSelectorStrategy",
    "roundRobin");
```

4. Connect using either:
   - Aurora Reader Endpoint (`cluster-ro`)
   - Aurora Cluster Endpoint

5. Execute the following sequential test:

```java
for (int i = 1; i <= 20; i++) {
    try (Connection conn =
             DriverManager.getConnection(url, props);
         Statement stmt =
             conn.createStatement();
         ResultSet rs =
             stmt.executeQuery(
                 "SELECT aurora_db_instance_identifier()")) {

        if (rs.next()) {
            System.out.println(
                i + " -> " + rs.getString(1));
        }
    }

    Thread.sleep(1000);
}
```

6. Observe that connections alternate between the two reader instances.

7. Execute a concurrent test with:
   - THREADS = 20
   - TOTAL_CONNECTIONS = 20
   - All threads released simultaneously using a CountDownLatch.

8. For each connection, execute:

```sql
SELECT aurora_db_instance_identifier();
```

9. Observe that all connections are established to the same reader instance.

10. (Optional) Configure WildFly/IronJacamar datasource with:

```xml
<min-pool-size>5</min-pool-size>
<max-pool-size>20</max-pool-size>
<prefill>true</prefill>
```

11. Start the application server and observe that all prefilled connections are established to a single reader instance.

### Possible Solution

_No response_

_No response_

aws-advanced-jdbc-wrapper-4.0.1

### JDK version used

openjdk version "11.0.19"

sergiyvamz:
## Summary

Fixes [#1968](https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1968).

When using the Aurora Initial Connection Strategy Plugin with `readerInitialConnectionHostSelectorStrategy=roundRobin`, concurrent connection creation (e.g. a connection-pool prefill at application startup) routed all connections to a single reader instead of distributing them.

### Root cause

The round-robin host selector is thread-safe, but it can only distribute across readers once cluster topology has been discovered. On a cold cache, `getCandidateHost()` returns no instance host, so the plugin falls back to connecting via the reader cluster endpoint, which is resolved by DNS. During a concurrent burst, every thread hits this fallback before topology is populated, and Aurora's reader endpoint DNS resolves them all to the same instance. Sequential connections work because the first connection warms the topology cache before the next attempt.

### Fix

Added a new opt-in parameter `waitForInitialTopologyMs` (default `0` = previous behavior). When set and topology is cold, the plugin blocks via `forceRefreshHostList(true, waitForInitialTopologyMs)` before falling back to the DNS route. This delegates to the existing per-cluster topology monitor, which:

- serializes concurrent callers into a single shared topology fetch (no thundering herd),
- is scoped per `clusterId`, so connections to other clusters are unaffected,
- falls back to the previous DNS behavior on timeout.

Once topology is available, the configured strategy distributes connections across readers.

## Changes

- Added `waitForInitialTopologyMs` property and the wait-then-select logic in `AuroraInitialConnectionStrategyPlugin`.
- Added log message keys for the wait and timeout paths.
- Added integration tests covering both the default behavior (all concurrent connections land on a single reader) and the fixed behavior (connections distributed across readers).
- Documented the new parameter and added a usage example in `UsingTheAuroraInitialConnectionStrategyPlugin.md`.

## Testing

- New integration tests in `AuroraInitialConnectionStrategyTest` require a live Aurora cluster (1 writer + 2 readers) and run in the integration CI workflows.

## Backward compatibility

Fully backward compatible: the feature is opt-in and defaults to `0`, preserving the existing behavior.


wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -193,95 +210,149 @@ public Connection connect(
...
+                    }
+                  }
+                } else {
+                  LOGGER.finest(() -> Messages.get(
+                      "AuroraInitialConnectionStrategyPlugin.waitForTopologyTimeout",
+                      new Object[] {this.waitForInitialTopologyMs, originalConnectHostFinal.getHost()}));
+                }
+              } else {
+                // Feature disabled. Preserve the previous behavior.
+                this.pluginService.forceRefreshHostList();
```
karenc-bq: Do we want to return after a successful return here? If endTimeNano is a large value due to WAIT_FOR_INITIAL_TOPOLOGY_MS setting, would we potentially be stuck here calling refresh over and over again?

karenc-bq: Ignore, misread the brackets

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
```diff
@@ -193,95 +210,149 @@ public Connection connect(
...
+              // topology, and it serves as a fallback connection if instance selection or connection fails.
+              candidateHost = originalConnectHost;
+              candidateConnHolder.set(connectFunc.call());
+
+              if (this.waitForInitialTopologyMs > 0) {
+                // Block until the topology for this cluster has been fetched, then re-attempt instance selection.
+                // This serializes concurrent/prefill connections (all waiting on the same per-cluster topology
+                // monitor) so that the configured host selection strategy can distribute them across instances
+                // instead of all relying on the initial endpoint resolved via DNS.
+                final HostSpec originalConnectHostFinal = originalConnectHost;
```
karenc-bq: Can we store the host directly since we only use it for logging, so `final String originalConnectHostFinal = originalConnectHost.getHost();`

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
sophia-bq: Do we want to note anywhere that setting waitForInitialTopologyMs to anything <= 5000 is faster than the default or 0. Maybe just a recommendation in the docs that this should be > 5000 as the intention is to give the monitor more time

sophia-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AuroraInitialConnectionStrategyPlugin.java
sergiyvamz: Hmm that's debatable statement about "<= 5000 is faster". Users may have completely different scenarios.

sophia-bq: To clarify what I mean: 
If the user does not set this new parameter or it's set to <= 0 we hit this line 

> // Feature disabled. Preserve the previous behavior.
> this.pluginService.forceRefreshHostList();

This means we go to the forceRefreshHostList implementation of PluginServiceImpl which calls `this.forceRefreshHostList(false, DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS);` where DEFAULT_TOPOLOGY_QUERY_TIMEOUT_MS = 5000. If a user sets your new parameter to anything less than 5000 then it's actually requiring the topology to update faster than if they set it to 0. It's just a little counterintuitive

sergiyvamz: I see. It seems this code needs another round of improvement. Added to backlog.

PR #1985, opened 2026-06-19
feat: introduce awsSecretsManager2 plugin (Stale-While-Revalidate)

sergiyvamz:
### Summary

Introduce `awsSecretsManager2`, a Stale-While-Revalidate (SWR) variant of the AWS Secrets Manager connection plugin that eliminates connection-time blocking on credential refresh.

Inspired by and re-implemented from [PR #1732](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1732), rebased onto current `main` (which now caches credentials via `StorageService`) and reworked to **reuse the existing plugin** instead of duplicating it.

### Problem

The existing `awsSecretsManager` plugin blocks every connection attempt when cached credentials have expired, waiting for a synchronous fetch before proceeding. This causes latency spikes on cache expiry, connection failures during transient Secrets Manager outages (even when valid credentials are cached), and a thundering herd under burst traffic.

### Solution

`awsSecretsManager2` extends `AwsSecretsManagerConnectionPlugin` and overrides only the cache-refresh strategy:
- **Fresh cache:** use cached credentials (same as v1).
- **Stale cache:** use stale credentials immediately and trigger a single async background refresh. Connections are never blocked.
- **Cache miss / login-failure retry:** synchronous fetch (same as v1), joining any in-flight refresh.

`ConcurrentHashMap.compute()` guarantees at most one in-flight refresh per `(secretId, region)`. Both plugins share the same credential cache.

### Code reuse

Rather than a parallel ~540-line copy, the original plugin was refactored with behavior-preserving extension points (`fetchAndStoreSecret`, `resolveSecretExpirationTime`, overridable `updateSecret`). The new plugin is a thin subclass and references no Jackson types, so it needs no `java17` variant.

### Changes

- New `AwsSecretsManagerConnectionPlugin2` + factory; registered `awsSecretsManager2` in `ConnectionPluginChainBuilder` and `ConnectionPluginManager`.
- Behavior-preserving refactor of `AwsSecretsManagerConnectionPlugin` (base and `java17` copies).
- `secretsManagerExpirationTimeSec` clamped to a 300s minimum.
- 18 unit tests (cache hit/miss/stale, SM-outage resilience, login-failure retry vs. skip, thundering-herd, interrupt handling, expiration clamping, ARN parsing, refresh cleanup). Existing v1 tests still pass.
- Documentation: new plugin guide, v1 cross-reference, plugin listing, and compatibility matrices.


docs/using-the-jdbc-driver/CompatibilityCrossPlugins.md
```diff
@@ -31,22 +32,23 @@ Similarly, the `limitless` plugin and `customEndpoint` plugin are incompatible b
...
+| [awsSecretsManager2](./using-plugins/UsingTheAwsSecretsManagerPlugin2.md)             | <span style="color:red;font-size:20px">&cross;</span>     | <span style="color:red;font-size:20px">&cross;</span>                   |                                                                           | <span style="color:red;font-size:20px">&cross;</span>           | <span style="color:red;font-size:20px">&cross;</span>    |
+| [federatedAuth](./using-plugins/UsingTheFederatedAuthPlugin.md)                       | <span style="color:red;font-size:20px">&cross;</span>     | <span style="color:red;font-size:20px">&cross;</span>                   | <span style="color:red;font-size:20px">&cross;</span>                     |                                                                 |                                                          |
```
karenc-bq, line 39: I think this row is missing a cross for okta

karenc-bq (approved):

PR #1843, opened 2026-04-09
Rwsplitting parse

davecramer:
### Summary

feat: Automatic ReadWrite Splitting plugin 

### Description

   feat: add SQL parser plugin with shared call context for inter-plugin communication
    
    Introduce SqlParserConnectionPlugin that parses SQL statements and shares
    results with downstream plugins via PluginCallContext, eliminating duplicate
    parsing across plugins.
    
    Changes:
    - Add PluginCallContext (thread-local per-call context) to plugin framework
    - Add getCallContext() to PluginService interface
    - Reset context before/after each JDBC call in ConnectionPluginManager
    - Create SqlParserConnectionPlugin (plugin code: sqlParser) that populates
      context with query type, table names, parameter mappings, and annotations
    - Create SqlContextKeys for standard context attribute names
    - Move JSQLParserAnalyzer, EncryptionAnnotationParser, SqlAnalysisService
      from plugin.encryption.parser/sql to shared jdbc.parser package
    - Update KmsEncryptionConnectionPlugin to read parse results from context
      instead of parsing SQL directly
    - Register sqlParser plugin in ConnectionPluginChainBuilder

### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1995, opened 2026-06-23
feat: automatic read/write splitting via SQL parsing

sergiyvamz:
## Summary

Reintroduces the automatic ReadWrite Splitting feature and adds integration test coverage for it.

Influenced by the original PR #1843 (`Rwsplitting parse`), with reviewer feedback folded in.

## What's included

- **`sqlParser` plugin** — parses each SQL statement once and shares the results (query type, tables, parameter→column mapping, `FOR UPDATE` flag, routing hint) with downstream plugins via a per-call `PluginCallContext`, avoiding duplicate parsing.
- **`autoReadWriteSplitting` plugin** — extends `ReadWriteSplitting` to route automatically based on the parsed query: `SELECT` → reader, DML/DDL → writer. `SELECT ... FOR UPDATE` goes to the writer, and `/*@reader*/` / `/*@writer*/` comment hints override the decision. Connection routing is left unchanged while a transaction is in progress.
- **Shared parser package** — `JSQLParserAnalyzer`, `EncryptionAnnotationParser`, and `SqlAnalysisService` moved to `software.amazon.jdbc.parser`; `kmsEncryption` now consumes parse results from the call context instead of parsing SQL itself.
- **Enums over strings** — `QueryType` and `RoutingHint` replace magic strings in the shared context.

## Changes from the original PR (review feedback)

- `PluginCallContext` stored on `PluginServiceImpl` (connection lifecycle), reset via `resetCallContext()`; `PartialPluginService.getCallContext()` throws as unsupported.
- `pluginWeightByPluginFactory` updated for both new plugins.
- `WrapperUtils.wrapExceptionIfNeeded()` used for exception wrapping.
- Read-only transactions stay on their current connection; `CallableStatement` execute methods handled.
- All user-facing messages externalized to the resource bundle.

## Testing

- Unit tests for `SqlParserConnectionPlugin`, `AutoReadWriteSplittingPlugin`, `QueryType`/parser, and updated `KmsEncryption` tests.
- Integration tests (`AutoReadWriteSplittingTests`) covering select→reader, DML→writer, `FOR UPDATE`→writer, hint overrides, prepared-statement SELECT, reader stickiness, and transaction routing. Integration tests require a multi-instance Aurora cluster and run in the pipeline.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Limitations
+
+### Statements are bound to a connection
+
+As with the Read/Write Splitting Plugin, a `Statement` or `ResultSet` is internally bound to the database connection that was active when it was created. If automatic routing switches the connection, statements created before the switch continue to use the previous connection. Create new `Statement`/`ResultSet` objects after a routing change. See [General plugin limitations](./UsingTheReadWriteSplittingPlugin.md#general-plugin-limitations) for more detail.
+
+### Batch and unparseable statements
+
+If a statement cannot be parsed (or carries no SQL text, such as `Statement#executeBatch`), no parse result is available and the plugin falls back to keeping the query on the writer to be safe.
```
karenc-bq: Can we use callable statements as examples here? Batch statements are for update/modifying anyway so there are no incentives to use readers in the first place

docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
```diff
@@ -0,0 +1,119 @@
...
+final Properties properties = new Properties();
+properties.setProperty(PropertyDefinition.PLUGINS.name, "sqlParser,autoReadWriteSplitting,failover2,efm2");
+```
+
+The driver performs plugin sorting by default (see the [`autoSortWrapperPluginOrder` configuration parameter](../UsingTheJdbcDriver.md#connection-plugin-manager-parameters)), which keeps `sqlParser` ahead of `autoReadWriteSplitting`.
+
+> [!WARNING]
+> Do not use the `autoReadWriteSplitting`, `readWriteSplitting`, `srw`, and/or `gdbReadWriteSplitting` plugins (or any combination of them) at the same time for the same connection. They are all read/write splitting plugins and will conflict.
+
+## Required dependency: JSQLParser
```
karenc-bq, line 35: Can we move this to the top above the "Loading..." section?

karenc-bq (approved):

PR #2009, opened 2026-07-02
feat: query-level load balancing for autoReadWriteSplitting plugin

sergiyvamz:
### Summary

Adds opt-in **query-level load balancing** to the Automatic Read/Write Splitting Plugin (`autoReadWriteSplitting`).

Two new parameters:

- `queryLevelLoadBalancing` (default `false`): when a read query is routed to a reader, select a reader **per query** using the configured `readerHostSelectorStrategy` instead of reusing a single cached reader.
- `loadBalancingIncludeWriter` (default `false`): include the writer instance in the pool of balancing candidates (only consulted when `queryLevelLoadBalancing` is enabled).

### Behavior

- Only reader-routed queries are balanced. Writes, `SELECT ... FOR UPDATE`, and `/*@writer*/` / `/*@keep*/`-hinted statements are unaffected.
- Balancing is suppressed inside a transaction and while autocommit is disabled, preserving transactional guarantees (reuses the plugin's existing pinning logic).
- The change is confined to `AutoReadWriteSplittingPlugin`. The `readWriteSplitting`, `srw`, and `gdbReadWriteSplitting` plugins are not affected — no shared base-class behavior was modified.
- Enabling the internal connection pool is recommended to avoid per-query connection churn.

### Testing

- Unit tests: property parsing, candidate-role selection, and connection switching (writer→reader, reader→reader with previous-reader cleanup, no-op when already on the selected host, include-writer null-role selection).
- Integration test scenarios (`AutoReadWriteSplittingTests`): reader rotation across multiple readers, `loadBalancingIncludeWriter`, and suppression inside a transaction.
- `checkstyleMain`, `checkstyleTest`, and the `AutoReadWriteSplittingPluginTest` suite pass locally.


wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/AutoReadWriteSplittingPlugin.java
```diff
@@ -119,6 +166,127 @@ public <T, E extends Exception> T execute(
...
+    }
+
+    this.isReaderConnFromInternalPool = Boolean.TRUE.equals(this.pluginService.isPooledConnection());
+    this.inReadWriteSplit = true;
+    // Store the new reader in the base reader cache (honoring cachedReaderKeepAliveTimeoutMs) so
+    // notifyConnectionChanged treats it as a cached connection and does not close it.
+    setReaderConnection(newConnection, target);
+    switchCurrentConnectionTo(newConnection, target);
+    LOGGER.finest(() -> Messages.get(
+        "AutoReadWriteSplittingPlugin.switchedToBalancedReader",
```
karenc-bq, line 246: Can we also log the strategy used

docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
```diff
@@ -84,6 +84,36 @@ The `/*@keep*/` hint is useful when you want to observe or operate on the curren
...
+
+By default, once the connection has switched to a reader it stays on that single reader for all subsequent read queries. Enabling query-level load balancing makes the plugin select a reader **per read query**, spreading reads across the available reader instances using the configured [reader selection strategy](../HostSelectionStrategies.md).
+
+| Parameter                     | Default | Description                                                                                                                                                 |
+|-------------------------------|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `queryLevelLoadBalancing`     | `false` | When `true`, a read query routed to a reader triggers a per-query reader selection instead of reusing the single cached reader.                              |
+| `loadBalancingIncludeWriter`  | `false` | When `true` (and `queryLevelLoadBalancing` is enabled), the writer instance is also included in the pool of balancing candidates. Only relevant with `random`/`roundRobin` strategies. |
```
karenc-bq: Curious why are we only supporting two strategies?

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/AutoReadWriteSplittingPlugin.java
```diff
@@ -119,6 +166,127 @@ public <T, E extends Exception> T execute(
...
+
+    if (target.getHostAndPort().equals(currentHost.getHostAndPort())) {
+      // Already connected to the selected host; nothing to switch.
+      this.inReadWriteSplit = true;
+      return;
+    }
+
+    // Preserve the writer connection so it can be reused for later write routing instead of
+    // being orphaned when we balance away from it.
+    if (isWriter(currentHost) && !isConnectionUsable(this.writerConnection)) {
```
karenc-bq, line 219: does it make sense to also check if currentConnection == this.writerConnection here?

karenc-bq (approved):

docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
sergiyvamz: Hmm what other strategies you expect to support. These new parameters are about when to switch connection (on every query or not) rather than what strategy to use to pick up a next reader.

sergiyvamz: I'll update that misleading comment.

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/AutoReadWriteSplittingPlugin.java
sergiyvamz: Condition `!isConnectionUsable(this.writerConnection)` means that the writer connection is either null or is already closed. That means it's useless and we need to store the `currentConnection`. If it's the same connection (`currentConnection == this.writerConnection`) it won't harm.

PR #2016, opened 2026-07-08
feat: unified read/write splitting plugin family

sergiyvamz:
## Summary
Refactors the read/write splitting plugins onto a unified, composition-based plugin so routing signal, switch gate, reader/writer resolution, caching, and initial-connection handling can be mixed, then builds new capabilities on top. Existing plugin codes and behavior are preserved.

## What's included
- Unified base (`UnifiedReadWriteSplittingPlugin`) with helper families; the existing codes (`readWriteSplitting`, `autoReadWriteSplitting`, `srw`, `gdbReadWriteSplitting`) are reconstructed as helper assemblies.
- New plugin codes: `autoSimpleReadWriteSplitting`, `gdbAutoReadWriteSplitting`, `gdbSimpleReadWriteSplitting`, `gdbAutoSimpleReadWriteSplitting`.
- Query-level load balancing (`queryLevelLoadBalancing`, `loadBalancingIncludeWriter`): select a fresh reader on each read-routing decision instead of reusing a sticky reader; family-wide, not auto-only.
- Statement recreation on connection switch (`allowStatementRecreationOnConnectionSwitch`, default on): reused `Statement`/`PreparedStatement`/`CallableStatement` are re-created on the routed connection (replaying recorded settings, bound parameters, and registered OUT parameters). Stream/LOB parameters or a pending batch fall back to the current connection with a one-time warning.

## Tests
- ~135 unit tests across signals, gate, cache, classifiers, resolvers, verifier, connect handlers, update policies, candidate sources, and plugin orchestration.
- Integration suites for the non-GDB plugins, plus reader-transaction and PreparedStatement-rebind scenarios. Integration tests require a live cluster.

## Docs
- Dedicated pages for the four new plugin codes, wired into the plugin index and cross-reference tables; updated configuration assistant, read/write splitting plugin pages, and README parameter table.


wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/refresher/TopologyRefresherImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+    final List<HostSpec> hosts = ctx.pluginService().getHosts();
+    if (Utils.isNullOrEmpty(hosts)) {
+      ctx.logAndThrow(Messages.get("ReadWriteSplittingPlugin.emptyHostList"));
+      return;
+    }
+
+    final HostSpec writerHost = Utils.getWriter(hosts);
+    if (writerHost == null) {
+      ctx.logAndThrow(Messages.get("ReadWriteSplittingPlugin.noWriterFound"));
```
karenc-bq, line 50: I wonder if we should change this method to `logAndCreateException`, this way the usage pattern becomes clearer, going from
```java
      ctx.logAndThrow(Messages.get("ReadWriteSplittingPlugin.noWriterFound"));
      return;
```
to 
```java
      throw ctx.logAndCreateException(Messages.get("ReadWriteSplittingPlugin.noWriterFound"));
```

sergiyvamz: I believe it's the case of personal preferences. I personally prefer the original code with `logAndThrow`.

karenc-bq (approved):

PR #2021, opened 2026-07-10
fix: propagate configuration profile to eagerly-created topology monitor (#2020)

### LINKED ISSUE #2020: ConfigurationProfile plugin list ignored - still
### Describe the bug

reopening https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1800

Hi @sergiyvamz

the good news is that I managed to get 4.1.0 working with our "no cross-regional networking" setup. Many properties need to be tweaked, but with the help of your AI skill and copilot, we got it done . if you are interested in the properties required to be switched on/configured for this setup, let me know. thank you for all the work you've put in.

the bad news is that this wrapper plugins issue failed in another place, and i still had to use workaround and specify both plugin classes and string of plugins.

copilot investigation below (this was done with our replacable datasource and not using global plugins)

In AWS Advanced JDBC Wrapper 4.1.0, when wrapper plugins are configured via the class-based ConfigurationProfile.withPluginFactories() API rather than the wrapperPlugins / PropertyDefinition.PLUGINS string property, the Aurora cluster topology monitoring thread (ctmi-*) fails to authenticate against IAM-only databases, throwing:

PSQLException: The server requested password-based authentication, but no password was provided by plugin null.
The root cause is that ClusterTopologyMonitorImpl.openAnyConnectionAndUpdateTopology() at line 731 calls forceConnect() on its own servicesContainer, whose ConnectionPluginManager was initialized with a null/absent ConfigurationProfile.

As a result, ConnectionPluginChainBuilder.getPlugins() took the string-code path (PATH B, line 205) instead of the class-factory path, and getPluginCodes() at line 254, finding no PropertyDefinition.PLUGINS value, fell back to:

DEFAULT_PLUGINS = "initialConnection,auroraConnectionTracker,failover2,efm2"
at line 145, which omits the iam plugin entirely, so the monitor's connection chain has no IAM authentication plugin.

PR https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1919 fixed the other monitor connection path:

getNodeMonitoringWorker -> ServiceUtility.createMinimalServiceContainer
at line 714 to propagate the profile, but it does not fix the CTMI's own servicesContainer used directly at line 731, so the topology monitor still authenticates without IAM.

The empirical workaround — explicitly setting:

wrapperPlugins=initialConnection,failover2,readWriteSplitting,iam
in the target datasource properties — resolves it by forcing PATH B's string fallback to include iam; the fact that this string-based workaround takes effect at all confirms the monitor's container is on PATH B with a null profile. If the class factories were active, the wrapperPlugins string would be ignored.

### Expected Behavior

bug fixed

as in issue 1800

### Current Behavior

as in issue 1800

### Reproduction Steps

as in issue 1800

### Possible Solution

_No response_

_No response_

4.1.0

### JDK version used

21

linux

sergiyvamz:
## Summary

Fixes [#2020](https://github.com/aws/aws-advanced-jdbc-wrapper/issues/2020) (a reopening of [#1800](https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1800)).

When the plugin list is configured through a class-based `ConfigurationProfile` (`ConfigurationProfileBuilder.withPluginFactories(...)`) rather than the `wrapperPlugins` string, the cluster topology monitor could still lose the profile-configured plugins (notably `iam`). Its background connections then authenticate without IAM and fail once the cached IAM token expires (~15 min), e.g.:

```
PSQLException: The server requested password-based authentication, but no password was provided by plugin null.
```

## Root cause

PR #1802 and PR #1919 wired the configuration profile through the monitor's minimal service container, so in the common case the monitor inherits the profile correctly. The remaining gap is an ordering problem in `ServiceUtility.createStandardServiceContainer`:

- `PluginServiceImpl` marks the dialect as **confirmed immediately** when a dialect is supplied via the profile (or `wrapperDialect`), and Global Aurora (GDB) deployments are confirmed up front automatically.
- With a confirmed dialect, `pluginService.refreshHostList()` inside `createStandardServiceContainer` **eagerly creates and caches** the `ClusterTopologyMonitorImpl`.
- That happened **before** `servicesContainer.setConfigurationProfile(configurationProfile)` was called, so the monitor's minimal container read a `null` profile and fell back to `DEFAULT_PLUGINS` (`initialConnection,auroraConnectionTracker,failover2,efm2`) - no `iam`.

This matches the reporter's observation that adding `iam` to the `wrapperPlugins` string works around it: the string path (PATH B) is only reached when the profile is missing.

## Fix

Set the configuration profile on the services container **before** creating the host list provider and refreshing the host list, so any eagerly-created topology monitor inherits it. This is a semantics-preserving move of a field setter to an earlier point in the same method; the container's final state is unchanged.

## Tests

- `ServiceUtilityTest` (unit): drives the real `createStandardServiceContainer` with a profile-supplied dialect and a host list provider that records the container's configuration profile at `refresh()` time (the exact point the monitor would be created). Verified it fails without the fix (`expected: <profile> but was: <null>`) and passes with it.
- `AwsIamIntegrationTest#test_AwsIam_TopologyMonitorInheritsProfilePlugins` (integration): configures plugins via `withPluginFactories` (iam, auroraConnectionTracker, failover2), confirms the dialect up front, connects IAM-only (no password) to an Aurora cluster, then forces a verified topology refresh. This drives the monitor to open its own connections through its plugin chain, which only succeeds if it inherited the `iam` plugin. Gated with `@EnableOnTestFeature(IAM)` + `@EnableOnDatabaseEngineDeployment(AURORA)` + `@DisableOnTestDriver(MARIADB)`.

### DIFF
--- a/wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
+++ b/wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
@@ -19,6 +19,8 @@ package integration.container.tests;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import integration.DatabaseEngine;
+import integration.DatabaseEngineDeployment;
 import integration.DriverHelper;
 import integration.TestEnvironmentFeatures;
 import integration.container.ConnectionStringHelper;
@@ -27,6 +29,7 @@ import integration.container.TestDriverProvider;
 import integration.container.TestEnvironment;
 import integration.container.condition.DisableOnTestDriver;
 import integration.container.condition.DisableOnTestFeature;
+import integration.container.condition.EnableOnDatabaseEngineDeployment;
 import integration.container.condition.EnableOnTestFeature;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
@@ -34,7 +37,9 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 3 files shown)


wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
```diff
@@ -258,6 +271,88 @@ void test_AwsIam_UserAndPasswordPropertiesArePreserved() throws SQLException {
     }
   }
 
+  /**
+   * Regression test for <a href="https://github.com/aws/aws-advanced-jdbc-wrapper/issues/2020">issue #2020</a>
+   * (reopening <a href="https://github.com/aws/aws-advanced-jdbc-wrapper/issues/1800">issue #1800</a>).
+   *
+   * <p>When the plugin list is configured via a class-based {@link ConfigurationProfileBuilder#withPluginFactories}
```
sophia-bq, line 278: Maybe I'm not familiar with these comments enough but what does `<p>` do here? Is it needed?

sophia-bq (approved):

wrapper/src/test/java/integration/container/tests/AwsIamIntegrationTest.java
sergiyvamz: It's a "new paragraph", or simply an extra line break.

Issue #2011, opened 2026-07-04
Add an explicit Blue/Green readiness log before manual switchover


qkrtkdwns3410:
## Describe the feature

Please add an explicit log event that indicates when the Blue/Green plugin has recognized the target green database/topology for a given deployment before a manual Blue/Green switchover is initiated.

The current Blue/Green plugin can collect source and target topology information internally and later emits a switchover summary after the switchover completes. However, there is no clear, standalone operational log that answers:

> Has this application instance already recognized the target green database/topology and built the blue-to-green mapping before I start the manual switchover?

This makes it difficult to confirm readiness per application instance before triggering a planned switchover.

## Use Case

Before manually initiating an Amazon RDS Blue/Green switchover, operators want to verify that each running application instance has already detected the green deployment and is ready for the switchover.

In our operational workflow, we planned to ship this readiness signal to OpenSearch and use it as a pre-switchover validation check per instance. The desired signal is not the final switchover summary after completion, but a pre-switchover readiness event that can be queried before the operator starts the switchover.

Without this log, teams need to implement custom application-side checks or infer readiness from very verbose driver logs, which is harder to automate and more error-prone.

## Proposed Solution

Add a concise log event from the Blue/Green plugin when the provider has enough information to confirm that the target green topology is known for the configured `bgdId`.

For example, emit an `INFO` or `FINE` level event from `BlueGreenStatusProvider` when both source and target interim statuses are available and the blue-to-green corresponding node map has been populated.

Example log shape:

```text
[bgdId: '1'] Blue/Green target topology recognized: sourceHosts=3, targetHosts=3, correspondingNodes=3
```

It would be useful if the event included structured, grep-friendly fields such as:

- `bgdId`
- source/target role
- current phase, for example `CREATED`
- source host count
- target host count
- corresponding node count
- whether the target green topology is considered ready

This would allow log pipelines such as OpenSearch to alert or dashboard on readiness before a manual switchover.

## Other Information

The current documentation recommends allowing the active `bg` plugin several minutes to collect deployment status before initiating switchover, then reviewing the switchover summary in application logs after completion.

Relevant current behavior observed from source:

- `BlueGreenStatusProvider.prepareStatus(...)` logs interim status at `FINEST`, updates internal state, updates corresponding nodes, updates summary status, and then calls `logCurrentContext()` and `logSwitchoverFinalSummary()`.
[...]

Issue #1996, opened 2026-06-25
Rename `latestStatus` in `BlueGreenStatusProvider.updateStatusCache` to reflect the previous cached status


qkrtkdwns3410:
### Describe the feature

In `BlueGreenStatusProvider.updateStatusCache`, the local variable name `latestStatus` is misleading.

```java
final BlueGreenStatus latestStatus = this.storageService.get(BlueGreenStatus.class, this.bgdId);
this.storageService.set(this.bgdId, this.summaryStatus);
```

At the point where `latestStatus` is read, it represents the previously cached status object for the given `bgdId`, not the latest status after the update.

Because the method then replaces the cached value with `summaryStatus`, the variable name can be misread as the newest status, while its actual role is closer to the previous cached status or the status object that waiting threads are currently synchronized on.

### Use Case

I was tracing the Blue/Green plugin flow and had to stop and verify whether `latestStatus` really meant the latest status or the previously cached status.

That naming makes the wait/notify flow in this method harder to understand than necessary, especially for contributors trying to reason about:

- why the object is retrieved before `set(...)`
- why `notifyAll()` is called on that object after replacing the cache entry
- how waiting threads detect that the cached status reference has changed

A clearer name would reduce confusion during debugging, maintenance, and onboarding.

### Proposed Solution

Rename `latestStatus` to something that reflects its actual role more precisely, for example:

- `previousStatus`
- `cachedStatusBeforeUpdate`
- `existingStatus`

Optionally, add a short comment explaining that the old cached object is retained so threads currently waiting on it can be notified after the cache entry is replaced.

### Other Information

Relevant code:

- `wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusProvider.java`
- `updateStatusCache()`

Current implementation:

```java
protected void updateStatusCache() {
  final BlueGreenStatus latestStatus = this.storageService.get(BlueGreenStatus.class, this.bgdId);
  this.storageService.set(this.bgdId, this.summaryStatus);
  this.storePhaseTime(this.summaryStatus.getCurrentPhase());

  // Notify all waiting threads that status is updated.
  // Those waiting threads are waiting on an existing status so we need to notify on it.
  if (latestStatus != null) {
    synchronized (latestStatus) {
      latestStatus.notifyAll();
    }
  }
}
```

The concern here is naming clarity rather than a confirmed runtime malfunction.

### Acknowledgements

[...]

PR #2032, opened 2026-07-14
feat: B/G pre-switchover readiness log

sergiyvamz:
## Summary

Adds a concise, one-time **INFO** readiness event to `BlueGreenStatusProvider` that is emitted once the plugin recognizes the target green topology for a `bgdId`. This lets operators confirm per-instance readiness (blue -> green corresponding-node mapping) before initiating a manual Blue/Green switchover.

## Related issues

- #1996
- #2011

## What it does

- Emits `logGreenTopologyRecognized()` once readiness is reached: both the source and target interim statuses are available and the blue-to-green corresponding node map is populated.
- Logged **at most once per deployment cycle** (the flag is reset on context completion), so it does not produce hot-path log spam.
- The event prints a stable, grep-friendly summary and the blue -> green host mapping, sized/dividered like the existing switchover summary.
- Guarded at method entry with `LOGGER.isLoggable(Level.INFO)`: when INFO is disabled, the readiness computation, the corresponding-node mapping build, and the once-per-cycle flag are all skipped. This matches the existing `isLoggable(FINEST)` guard pattern already used in this class.

## Changes

- `BlueGreenStatusProvider.java`: new `greenTopologyRecognizedLogged` flag, the `logGreenTopologyRecognized()` method, its call site after `logCurrentContext()`, and the once-per-cycle reset.
- `aws_advanced_jdbc_wrapper_messages.properties`: new `bgd.greenTopologyRecognized.title` / `.summary` / `.mappingHeader` keys (user-facing strings kept in the resource bundle).

## Testing

- `:aws-advanced-jdbc-wrapper:compileJava` and `:checkstyleMain` pass.
- Behavior is INFO-log-only; no functional path is changed. The readiness computation is fully short-circuited when INFO is disabled.


wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
```diff
@@ -529,6 +529,12 @@ bgd.unknownPhase=[bgdId: ''{0}''] Unknown BG phase ''{1}''.
 bgd.blueDnsCompleted=[bgdId: ''{0}''] Blue DNS update completed.
 bgd.greenDnsRemoved=[bgdId: ''{0}''] Green DNS removed.
 bgd.greenTopologyChanged=[bgdId: ''{0}''] Green topology changed.
+bgd.greenTopologyRecognized.title=[bgdId: ''{0}''] Blue/Green target topology recognized
+bgd.greenTopologyRecognized.labelPhase=phase
+bgd.greenTopologyRecognized.labelSourceHosts=sourceHosts
+bgd.greenTopologyRecognized.labelTargetHosts=targetHosts
+bgd.greenTopologyRecognized.labelCorrespondingNodes=correspondingNodes
+bgd.greenTopologyRecognized.labelReady=ready
```
sophia-bq: Wondering what additional value these add defined like this?

sophia-bq (approved):

wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
sergiyvamz: That's a legit question. I think it'd be easier to hardcode them.

marcicodes:
when will this feature be released to production?

PR #2034, opened 2026-07-17
chore: improve CachedSQLXML user experience

Frank-Gu-81:
### Summary

  Harden XML parser factories and tighten the deserialization allowlist in the Remote Query Cache Plugin

  ### Description
  
  **XML parser configuration in `CachedSQLXML.getSource()`**

  Configures the XML parser factories used for the `DOMSource`, `SAXSource`, and `StAXSource` paths:

  - Disable DTD processing (`disallow-doctype-decl`) and external DTD loading
  - Disable external general and parameter entity resolution
  - Enable secure processing (`FEATURE_SECURE_PROCESSING`) on the DOM and SAX factories
  - Set `ACCESS_EXTERNAL_DTD` and `ACCESS_EXTERNAL_SCHEMA` to empty on the DOM factory
  - Set `ACCESS_EXTERNAL_DTD` and `ACCESS_EXTERNAL_SCHEMA` on the SAX parser
  - Set `SUPPORT_DTD=false` and `IS_SUPPORTING_EXTERNAL_ENTITIES=false` on the StAX input factory

  `getSource(StreamSource.class)` is now disabled by default because the driver cannot control how the returned stream is subsequently parsed. Callers that require the previous passthrough behavior can opt in via the new plugin property `cacheAllowStreamSource`.

  **Deserialization allowlist in `CachedResultSet.SafeObjectInputStream`**

  - Overrides `resolveProxyClass` to reject dynamic proxies (`java.lang.reflect.Proxy`), which are not a data type produced by any driver's `ResultSet.getObject()`.
  - Removes `java.net.URL` from the default deserialization allowlist because its equality/hash semantics involve network resolution, which is not appropriate for values reconstructed from the cache. Prefer `java.net.URI`. Users who require URL deserialization can opt in via the new plugin property `cacheAllowUrl`.

  **Per-connection configuration (`CacheDeserializationConfig`)**

  Adds `CacheDeserializationConfig`, an immutable per-connection configuration object built by `RemoteQueryCachePlugin` from the `cacheAllowUrl` and `cacheAllowStreamSource` properties and threaded through `CachedResultSet`, `CachedRow`, `SafeObjectInputStream`, and `CachedSQLXML` as instance state. Two connections with different `cacheAllow*` settings do not interfere.

  `CachedSQLXML` holds its config in a `transient` field injected by `CachedRow.get()` immediately after deserialization, so a poisoned cache entry cannot supply its own config. If the config is not injected, `getSource(...)` falls back to `CacheDeserializationConfig.STRICT` (fail-closed). Adds explicit `serialVersionUID = 1L` on `CachedSQLXML`.

  **New plugin properties**

  | Property | Default | Description |
  |---|---|---|
  | `cacheAllowStreamSource` | `false` | Allow `SQLXML.getSource(StreamSource.class)` on cached XML values. |
  | `cacheAllowUrl` | `false` | Allow `java.net.URL` to be reconstructed from the cache. |

  Both properties are described in `UsingTheRemoteQueryCachePlugin.md`, along with the exceptions users can expect if their existing code relied on the previous behavior and how to opt in.
  
  **Tests**

  Added coverage in the existing test files:

  - `CachedSQLXMLTest` — DOCTYPE-bearing XML is rejected for `DOMSource`, `SAXSource`, and `StAXSource`; `StreamSource` is rejected by default and works when the property is enabled.
  - `CachedResultSetTest` — `java.net.URL` is rejected by default and reconstructed when the property is enabled; dynamic proxies are rejected.

  **Behavior changes**

  - `getSource(StreamSource.class)` on cached XML now throws `SQLException` by default; set `cacheAllowStreamSource=true` to restore the previous behavior.
  - Deserialization of `java.net.URL` values from the cache now throws `SQLException` by default; set `cacheAllowUrl=true` to restore the previous behavior.
  - Well-formed XML without a DOCTYPE and all other cached types are unaffected.

  ### Additional Reviewers
  
  


docs/using-the-jdbc-driver/using-plugins/UsingTheRemoteQueryCachePlugin.md
```diff
@@ -55,6 +55,8 @@ ResultSet rs = stmt.executeQuery("/* CACHE_PARAM(ttl=300s) */ select * from myta
 | `failWhenCacheDown`                | Boolean |    No    | Whether to throw SQLException on cache failures under Degraded mode or make queries fall back to the database.                                          | `false`       |
 | `cacheInFlightWriteSizeLimitBytes` | Integer |    No    | Maximum in-flight write size in Bytes to the cache server before triggering degraded mode.                                                              | `50MB`        |
 | `cacheHealthCheckInHealthyState`   | Boolean |    No    | Whether to run health checks (pings) in healthy state.                                                                                                  | `false`       |
+| `remoteQueryCachePlugin.allowStreamSourceFromCache` | Boolean | No | Whether `SQLXML.getSource(StreamSource.class)` is allowed for XML values retrieved from the cache. See [XML Columns](#xml-columns) below. | `false` |
```
sergiyvamz: You need to add these parameters to README.md as well.

sergiyvamz: It's better to avoid dot in parameter names. Camel notation is our style.

Frank-Gu-81: Goof catch. Fixing this now. I should also update the config param to be prefixed with `cache*` for consistency.

Frank-Gu-81: Fixed

sergiyvamz (approved):

PR #2045, opened 2026-07-22
feat: add XA distributed transaction support (AwsWrapperXADataSource)

sergiyvamz:
## Summary

Adds XA / distributed-transaction support to the AWS Advanced JDBC Wrapper via a new
`software.amazon.jdbc.ds.AwsWrapperXADataSource` (`javax.sql.XADataSource`). It wraps a target
driver-specific `XADataSource` (PostgreSQL, MySQL, MariaDB) so the wrapper's plugin pipeline applies
to the application-facing `Connection`, while the XA control surface (`XAConnection` / `XAResource`)
is delegated to the target driver and coordinated by a JTA transaction manager (WildFly/Narayana,
Atomikos, Bitronix).

## What's included

**XA datasource and wrappers**
- `AwsWrapperXADataSource` - same configuration surface as `AwsWrapperDataSource`; requires a target
  `XADataSource` class and fails fast otherwise.
- `ds/xa/`: `XADataSourceConnectionProvider`, `XAConnectionWrapper`, `XAResourceWrapper`,
  `XaTransactionStateCallback`. Each `getConnection()` builds a fresh `ConnectionWrapper`/pipeline over
  the same pinned physical XA connection.
- Shared `DataSourceConfigHelper` extracted from `AwsWrapperDataSource` (no behavior change on the
  non-XA path).

**Connection pinning for XA correctness**
- `PluginService.isXaTransactionActive()` flag, set around the XA branch by `XAResourceWrapper`.
- Read/write splitting is skipped while a branch is active (`TransactionAwareGate`); failover fails
  fast with `XaFailoverNotSupportedSQLException` (SQLSTATE `08007`) in failover v1/v2/gdb.
- `PluginServiceImpl.setCurrentConnection` backstop rejects a connection swap during an active branch
  (SQLSTATE `25001`).
- Internal connection pool / custom connection provider are rejected on the XA path (they can't
  expose an `XAResource`).

**Auth and configuration**
- IAM authentication works on the XA path: `XADataSourceConnectionProvider.connect()` applies the
  per-connect credentials (the IAM token) to the target `XADataSource` before opening the session.
- Wrapper parameters are stripped from the URL handed to the target driver
  (`DataSourceConfigHelper.removeWrapperPropertiesFromUrl`), so `wrapperPlugins`, `wrapperDialect`,
  etc. never leak to the target (which some drivers reject as unknown parameters). Configuration is
  via the `jdbcUrl` query string and/or `targetDataSourceProperties`.
- `XAResourceWrapper.isSameRM` returns `false` so the transaction manager performs a full 2PC across
  separate branches (avoids MySQL Connector/J rejecting `XA START ... JOIN`).
- All user-facing strings are externalized to the message bundle.

## Testing

- **Unit tests** (all passing, full suite green): `AwsWrapperXADataSourceTest`,
  `XADataSourceConnectionProviderTest`, `XAResourceWrapperTest`, plus backstop/fail-fast/analyzer
  cases in `PluginServiceImplTests`, `FailoverConnectionPluginTest`, `TransactionAwareGateTest`,
  `SqlMethodAnalyzerTest`.
- **Integration tests, run against real databases via `test-all-docker` (PostgreSQL + MySQL) - 21
  passed / 0 failed** across the PG, MySQL, and MariaDB drivers:
  - `XaTransactionTest`: direct `XAResource` lifecycle (commit, rollback, one-phase commit, recovery).
  - `XaTwoPhaseCommitTest`: TM-coordinated two-resource 2PC under **both** Narayana and Atomikos.
- **Integration tests authored for a real Aurora/RDS environment** (not runnable on the docker suite):
  - `XaFailoverTest`: failover mid-branch fails fast; the unprepared branch is rolled back; a new
    transaction commits on the promoted writer.
  - `XaIamAuthenticationTest`: IAM-authenticated XA connection and branch lifecycle.

## Documentation

- New `docs/using-the-jdbc-driver/UsingXADataSource.md` (configuration, per-driver target classes,
  WildFly + IAM examples, benefits/limitations, the recommended two-datasource pattern), linked from
  `docs/Documentation.md` and `docs/using-the-jdbc-driver/DataSource.md`.
- XA coverage added to the configuration assistant skill (`docs/JDBC-WRAPPER-CONFIGURATION-ASSISTANT.md`).

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperXADataSource.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperXADataSource.java
@@ -1,0 +1,467 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.jdbc.ds;
+
+import java.io.PrintWriter;
+import java.io.Serializable;
+import java.sql.SQLException;
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Properties;
+import java.util.Set;
+import java.util.logging.Logger;
+import javax.sql.XAConnection;
+import javax.sql.XADataSource;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.Driver;
+import software.amazon.jdbc.PropertyDefinition;
+import software.amazon.jdbc.authentication.AwsCredentialsManager;
+import software.amazon.jdbc.ds.xa.XAConnectionWrapper;
+import software.amazon.jdbc.ds.xa.XADataSourceConnectionProvider;
+import software.amazon.jdbc.profile.ConfigurationProfile;
+import software.amazon.jdbc.profile.DriverConfigurationProfiles;
+import software.amazon.jdbc.targetdriverdialect.TargetDriverDialect;
+import software.amazon.jdbc.targetdriverdialect.TargetDriverDialectManager;
+import software.amazon.jdbc.util.ConnectionUrlParser;
+import software.amazon.jdbc.util.CoreServicesContainer;
+import software.amazon.jdbc.util.FullServicesContainer;
+import software.amazon.jdbc.util.Messages;
+import software.amazon.jdbc.util.PropertyUtils;
+import software.amazon.jdbc.util.ServiceUtility;
+import software.amazon.jdbc.util.SqlState;
+import software.amazon.jdbc.util.StringUtils;
+import software.amazon.jdbc.util.WrapperUtils;
+import software.amazon.jdbc.util.events.EventPublisher;
+import software.amazon.jdbc.util.monitoring.MonitorService;
+import software.amazon.jdbc.util.storage.StorageService;
+import software.amazon.jdbc.util.telemetry.DefaultTelemetryFactory;
+import software.amazon.jdbc.util.telemetry.TelemetryFactory;
+import software.amazon.jdbc.wrapper.ConnectionWrapper;
+
+/**
+ * An {@link XADataSource} entry point for the AWS Advanced JDBC Wrapper. It wraps a target
+ * driver-specific {@code XADataSource} (for example {@code org.postgresql.xa.PGXADataSource} or
+ * {@code com.mysql.cj.jdbc.MysqlXADataSource}) so the wrapper's plugin pipeline applies to the
+ * application-facing {@link java.sql.Connection}, while the XA control surface is delegated to the
+ * target's {@code XAResource}.
+ *
+ * <p>Connection-switching plugins (read/write splitting, failover) are not supported on the XA
+ * path; see the XA documentation for the recommended two-datasource pattern. The wrapper's internal
+ * connection pool is also not applicable to the XA path (the transaction manager / application
+ * server provides XA connection pooling).
+ */
+public class AwsWrapperXADataSource implements XADataSource, Serializable {
+
+  private static final long serialVersionUID = 1L;
+  private static final Logger LOGGER = Logger.getLogger(AwsWrapperXADataSource.class.getName());
+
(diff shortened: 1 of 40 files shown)


wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperXADataSource.java
```diff
@@ -0,0 +1,467 @@
...
+  public XAConnection getXAConnection(final @Nullable String username, final @Nullable String password)
+      throws SQLException {
+    this.user = username;
+    this.password = password;
+
+    // Reject the wrapper's internal connection pool on the XA path: those providers pool plain
+    // Connections and cannot expose an XAResource. XA connection pooling is provided by the
+    // transaction manager / application server.
+    final Properties rawProps = PropertyUtils.copyProperties(
+        this.targetDataSourceProperties != null ? this.targetDataSourceProperties : new Properties());
```
karenc-bq:
```suggestion
    final Properties rawProps = PropertyUtils.copyProperties(this.targetDataSourceProperties);
```
copyProperties already has a null check and will return new Properties() if null.

wrapper/src/main/java/software/amazon/jdbc/ds/AwsWrapperXADataSource.java
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  /**
+   * Resolves the final target URL, effective properties, target driver dialect, and configuration
+   * profile from the configured server/port/database or JDBC URL. (Kept local to the XA datasource
+   * for now; a shared resolver with {@link AwsWrapperDataSource} is a later optimization.)
+   */
+  ResolvedConfig resolveConfig() throws SQLException {
+    final Properties props = PropertyUtils.copyProperties(
+        this.targetDataSourceProperties != null ? this.targetDataSourceProperties : new Properties());
```
karenc-bq:
```suggestion
    final Properties props = PropertyUtils.copyProperties(this.targetDataSourceProperties);
```

wrapper/src/main/java/software/amazon/jdbc/ds/xa/XADataSourceConnectionProvider.java
```diff
@@ -0,0 +1,273 @@
...
+          put(WeightedRandomHostSelector.STRATEGY_WEIGHTED_RANDOM, new WeightedRandomHostSelector());
+        }
+      });
+
+  private final @NonNull XADataSource xaDataSource;
+  private final @NonNull String xaDataSourceClassName;
+  private final @Nullable String resolvedUrl;
+  private final ResourceLock lock = new ResourceLock();
+  private volatile @Nullable XAConnection xaConnection;
+  private volatile @Nullable Connection logicalConnection;
```
karenc-bq:
```suggestion
  private @Nullable Connection logicalConnection;
```
logicalConnection is only ever accessed and updated within locks so the volatile isn't necessary here

wrapper/src/main/java/software/amazon/jdbc/ds/DataSourceConfigHelper.java
```diff
@@ -0,0 +1,199 @@
...
+
+      finalUrl = ConnectionUrlBuilder.buildUrl(jdbcProtocol, effectiveServerName, port, databaseName);
+
+      // Override credentials with the ones provided through the data source property.
+      setCredentialProperties(props, user, password);
+
+      // Override database with the one provided through the data source property.
+      if (!StringUtils.isNullOrEmpty(databaseName)) {
+        PropertyDefinition.DATABASE.set(props, databaseName);
+      }
```
karenc-bq: This is called in both branches. Can we move it out?
i.e. Going from
```java
if {
      ...
      setCredentialProperties(props, user, password);

      // Override database with the one provided through the data source property.
      if (!StringUtils.isNullOrEmpty(databaseName)) {
        PropertyDefinition.DATABASE.set(props, databaseName);
      }
} else {
     ...
      setCredentialProperties(props, user, password);

      // Override database with the one provided through the data source property.
      if (!StringUtils.isNullOrEmpty(databaseName)) {
        PropertyDefinition.DATABASE.set(props, databaseName);
      }
}
```
to
```java
if {
      ...
} else {
     ...
}
      setCredentialProperties(props, user, password);

      // Override database with the one provided through the data source property.
      if (!StringUtils.isNullOrEmpty(databaseName)) {
        PropertyDefinition.DATABASE.set(props, databaseName);
      }
```

wrapper/src/main/java/software/amazon/jdbc/ds/DataSourceConfigHelper.java
```diff
@@ -0,0 +1,199 @@
...
+    final String base = url.substring(0, queryStart);
+    final String query = url.substring(queryStart + 1);
+    if (StringUtils.isNullOrEmpty(query)) {
+      return base;
+    }
+
+    // Collect the parameter names, then let PropertyDefinition.removeAll decide which are AWS
+    // Wrapper properties; whatever it leaves behind are the target-driver parameters to keep.
+    final Properties names = new Properties();
+    for (final String token : query.split("&")) {
```
karenc-bq: `query.split("&")` is computed twice. Can we save it to a variable and reuse?

```suggestion
    final String[] tokens = query.split("&");
    for (final String token : tokens) {
```

karenc-bq (approved):

PR #2057, opened 2026-07-28
fix: pin connection when autocommit state is unknown

sergiyvamz:
Follow-ups from the review that produced #2056. Independent of that PR - no overlapping files, branched from `main`.

## 1. `TransactionAwareGate`: pin when the autocommit state cannot be read

The gate refuses to move to another physical connection while a transaction is in progress, and (for `autoReadWriteSplitting`, via `pinOnAutoCommitOff`) while autocommit is explicitly disabled, because the next statement would join an implicit transaction. If reading the autocommit state threw, it fell back to allowing the switch:

```java
} catch (final SQLException e) {
  // If autocommit state cannot be determined, fall back to allowing the switch.
  return true;
}
```

That is the wrong default for the one input the gate cannot verify. An unknown autocommit state cannot rule out an implicit transaction, and switching on a wrong guess moves later statements to a different connection while the transaction stays open on the original one. Pinning instead costs a missed routing opportunity on a connection whose session state cannot even be queried - a bounded, recoverable cost against an unbounded correctness one.

Unchanged on purpose:

- An empty (never explicitly set) autocommit value still allows the switch. Empty means the JDBC default is in effect, so there is no implicit transaction to protect. Pinning there would disable read/write splitting for the common case.
- The classic gate (`new TransactionAwareGate()`, used by `readWriteSplitting`) does not consult autocommit at all and is unaffected. A test now covers that.

The existing `autoCommitIndeterminate_fallsBackToAllowingSwitch` test asserted the old behavior; it is renamed and inverted, since the assertion itself is the decision being changed. A `FINE` log line and a message bundle entry were added so the pinning is visible when diagnosing why splitting stopped.

## 2. Remove `DefaultConnectionPlugin.parseMultiStatementQueries()`

A package-private copy of the helper in `SqlMethodAnalyzer` with no production callers - the only reference was its own unit test. Package-private visibility also rules out external callers. Removed along with that test and the imports it needed.

Keeping a second copy of statement splitting invites the two from drifting apart, which is exactly what happened around comment handling (see #2056).

## Tests

Verified locally: full wrapper unit suite passes (1302 completed, 0 failed, 81 skipped - skips pre-existing), plus `checkstyleMain` and `checkstyleTest`. The count drops because the five parameterized cases for the removed dead method are gone; one gate test was added.

## Considered and not done

The review also floated having `SqlMethodAnalyzer` report an explicit "unknown" for SQL it cannot classify, with the gate refusing to switch on unknown. On closer reading that does not pay for itself: statements that are not transaction control are the overwhelmingly common case and are indistinguishable from genuinely unclassifiable text at that layer, so an "unknown means pin" rule would pin nearly every connection and effectively disable read/write splitting.

One genuine gap did surface while looking at this, and it is deliberately left for separate work: only the *first* statement of a multi-statement query is analyzed, so `SELECT 1; BEGIN` opens a transaction that goes undetected, and `SELECT 1; COMMIT` closes one that is never cleared. Fixing it means folding over every statement in the query, which first requires the `;` split to become quote-aware - otherwise a literal such as `INSERT INTO t VALUES ('a; commit')` would produce a bogus second statement and report a commit that never happened. That is a behavior change to transaction tracking and deserves its own PR.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
@@ -18,8 +18,6 @@ package software.amazon.jdbc.plugin;
 
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
@@ -28,7 +26,6 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.jdbc.ConnectionInfo;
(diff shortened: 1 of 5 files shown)


sergiyvamz:
Superseded by #2056: both commits were moved there so the related transaction-detection fixes ship as one change. Closing this PR.

PR #2058, opened 2026-07-29
fix: forward wrapper settings to target XA data sources

sergiyvamz:
## Summary

Fixes the failures from [Run Aurora Integration Tests Latest #30384801581](https://github.com/aws/aws-advanced-jdbc-wrapper/actions/runs/30384801581) on `main` (all 5 jobs failed). The failures fell into five root causes; this PR addresses all of them.

## Main code

**Wrapper timeouts were silently dropped on the XA datasource path.** `AwsWrapperXADataSource` configures the target driver's `XADataSource` directly instead of going through `TargetDriverDialect#prepareConnectInfo`, and hands it only non-wrapper properties. `connectTimeout` / `socketTimeout` / `tcpKeepAlive` were therefore discarded, so an XA connection ran with no socket timeout: when the network was blackholed, the pinned XA session waited instead of surfacing a connection failure, and failover never triggered. This is what made `XaFailoverTest` fail on RDS multi-AZ (where the test simulates an outage with a proxy blackhole rather than a real server failover) with "Expected `XaFailoverNotSupportedSQLException` to be thrown, but nothing was thrown".

- New `TargetDriverDialect#prepareTargetDataSource(CommonDataSource, url, props)` hook (default implementation is a no-op returning the URL unchanged), called from `AwsWrapperXADataSource#configureTargetXaDataSource` and again at connect time in `XADataSourceConnectionProvider` against the final per-connect properties.
- PostgreSQL and MySQL Connector/J dialects apply keep-alive and timeouts to the target in the unit each driver expects (seconds for PostgreSQL, milliseconds for Connector/J). The timeout/keep-alive block in `PgDriverHelper` was extracted so both entry points share it, with no change to the existing call order.
- Both dialects now recognize their XA data source class names (`org.postgresql.xa.PGXADataSource`, `com.mysql.cj.jdbc.MysqlXADataSource`), so the XA path resolves the real driver dialect instead of falling back to the generic one.

**MariaDB Connector/J rejected the target URL.** `MariaDbDataSource#setUrl` refuses a `jdbc:mysql://` URL ("Wrong mariaDB url") unless the URL itself carries `permitMysqlScheme`, which cannot be set as a bean property. The MariaDB dialect now adds it for a MySQL-scheme URL, along with the connect/socket timeouts that driver only accepts as URL parameters (existing parameters are never duplicated or overridden).

## Tests

- `ReadWriteSplittingTests#test_pooledConnection_differentUsers` asserted on `HikariPool.PoolInitializationException`, which #2051 now wraps in a `SQLException` inside `HikariPooledConnectionProvider`. The test expects the `SQLException` and asserts on its cause. This one assertion accounted for 4 failures per environment (the test is inherited by `Auto`, `Simple`, and `AutoSimple` subclasses) across every job.
- `XaTransactionTest#test_readWriteSplittingDoesNotSwitchDuringXaBranch` tolerates a driver that refuses `setReadOnly` while a transaction is active. pgjdbc rejects it client-side ("Cannot change transaction read-only property in the middle of a transaction"), which the JDBC spec allows and which equally proves no connection switch happened. The instance-id assertion is unchanged.
- XA tests that need a prepared branch are skipped (JUnit assumption with an explanatory message) when the server has prepared transactions disabled. PostgreSQL defaults to `max_prepared_transactions = 0`, which makes `XAResource.prepare` fail with "prepared transactions are disabled"; that is a server prerequisite set through the DB/cluster parameter group, not something the driver can work around. This was the largest single contributor to the PostgreSQL job failures. Enabling `max_prepared_transactions` on the PostgreSQL test clusters is a follow-up needed to get real XA coverage there; the containerized standard tests already enable it.

## Testing

- `:aws-advanced-jdbc-wrapper:test` - 1315 unit tests pass.
- New unit tests cover the timeout forwarding for `PGXADataSource` (ms to seconds) and `MysqlXADataSource` (ms passthrough), and the MariaDB URL handling (adds `permitMysqlScheme` only for the MySQL scheme, adds timeouts, does not duplicate or override existing parameters, and the URL is accepted by a real `MariaDbDataSource`).
- `checkstyleMain` / `checkstyleTest` pass; the Checker Framework nullness build compiles with no new findings.
- The Aurora/RDS integration tests themselves were not run locally (they need a live AWS test environment).

### DIFF
--- a/wrapper/src/test/java/integration/util/XaTestUtility.java
+++ b/wrapper/src/test/java/integration/util/XaTestUtility.java
@@ -56,7 +56,9 @@ public final class XaTestUtility {
         maxPreparedTransactions > 0,
         "PostgreSQL prepared (two-phase) transactions are disabled on this server "
             + "(max_prepared_transactions=" + maxPreparedTransactions + "). Set max_prepared_transactions "
-            + "to a value greater than 0 in the DB (cluster) parameter group to run the XA prepare tests.");
+            + "to a value greater than 0 in the DB (cluster) parameter group to run the XA prepare tests. "
+            + "The test framework enables it on the PostgreSQL databases it creates, so this test is only "
+            + "skipped when running against a database that was created without it (for example a reused one).");
   }
 
   private static int queryMaxPreparedTransactions() throws SQLException {
--- a/wrapper/src/test/java/integration/util/XaTestUtility.java
+++ b/wrapper/src/test/java/integration/util/XaTestUtility.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package integration.util;
+
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import integration.DatabaseEngine;
+import integration.container.ConnectionStringHelper;
+import integration.container.TestEnvironment;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.logging.Logger;
+
+/** Shared helpers for the XA ({@code XADataSource}) integration tests. */
+public final class XaTestUtility {
+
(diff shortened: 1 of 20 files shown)


wrapper/src/test/java/integration/util/XaTestUtility.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private static final Logger LOGGER = Logger.getLogger(XaTestUtility.class.getName());
+
+  private XaTestUtility() {
+  }
+
+  /**
+   * Skips the calling test when the database cannot prepare (two-phase) transactions.
+   *
+   * <p>PostgreSQL disables prepared transactions by default ({@code max_prepared_transactions = 0}),
```
sophia-bq, line 42: Just a question: for our automated integration tests are we able to adjust the default through our testing RDS/Aurora parameter group such that we do run these tests? I agree this is the right change overall but ideally we'd test against PG if possible

sophia-bq (approved):

wrapper/src/test/java/integration/util/XaTestUtility.java
sergiyvamz: That's a good call.

PR #2074, opened 2026-08-05
fix: repair the JMH benchmarks and fill their coverage gaps

sergiyvamz:
## Summary

The JMH benchmarks compiled but did not run. A full-suite run produced **one** result out of 24, and that one was `PluginBenchmarks.initAndReleaseBaseLine` - an empty method body. Every other benchmark aborted in setup or warmup. Nothing in CI compiles this module, so the breakage was invisible.

This PR repairs the suite, wires it into CI so it cannot rot again, and fills the coverage gaps that were left. The suite goes from **24 benchmarks (23 failing)** to **149 benchmarks, all reporting, zero exceptions**.

Six commits, each one compiled and smoke-run before being committed.

## Part 1: the repair

Three independent breakages, all caused by wrapper internals moving on without the benchmark harness.

| Symptom | Cause |
|---|---|
| All 12 `ConnectionPluginManagerBenchmarks` aborted with `IllegalArgumentException: pluginManagerService` | `DefaultConnectionPlugin` now resolves `PluginManagerService` and `ImportantEventService` from the services container; only `getPluginService()` was stubbed |
| All 11 non-empty `PluginBenchmarks` failed with NPEs | Three separate points: `ConnectionWrapper.init` calling `getTargetDriverDialect()`, `WrapperUtils.doExecuteWithPlugins` calling `getServicesContainer()`, and `ConnectionWrapper.close` calling `getSessionStateService()` |
| `PgCacheBenchmarks` could only ever fail | Endpoints were hardcoded placeholder host names (`db-0.XYZ.us-east-2...`) |

`TestConnectionWrapper` now builds through the production `ConnectionWrapper` constructor with a services container, instead of the `protected` "for testing purposes only" constructor. That constructor leaves `servicesContainer` null, which every JDBC call now dereferences, and it has no remaining callers anywhere in the repo - it is the thing that drifted, and is a candidate for removal in a follow-up.

`PgCacheBenchmarks` endpoints now come from system properties (`benchmarks.pg.url`, `benchmarks.cache.rw`, `benchmarks.cache.ro`) and setup fails fast naming the missing one. It also gained the license header it was missing and a null-safe teardown, so a failed setup is no longer masked by an NPE.

Three correctness bugs found while fixing the above:

- `initConnectionPluginManagerWithNoPlugins` passed the 10-plugin configuration profile. A profile overrides the `wrapperPlugins` property, so the benchmark was a duplicate of the `WithPlugins` case rather than a no-plugin baseline.
- `initHostProvidersWithNoPlugins` and `releaseResourcesWithNoPlugins` returned the wrong plugin manager.
- `ConnectionPluginManagerBenchmarks.main()` ran `PluginBenchmarks`.

### Why CI missed it

The benchmarks live in the `jmh` source set, so `./gradlew test` reports `:benchmarks:compileTestJava NO-SOURCE`, and checkstyle/spotbugs are configured only on the `wrapper` module. `:benchmarks:jmhJar` is now part of the `lint` job, which sits behind the `ci-gate` required status check.

## Part 2: filling the coverage gaps

Two gaps mattered most. There was no target-driver baseline, so the module could not answer "what does the wrapper cost" - every benchmark compared the wrapper against itself with a different plugin list. And no benchmark instantiated a real plugin: `ConnectionPluginManagerBenchmarks` uses chains of a no-op `BenchmarkPlugin`, and `PluginBenchmarks` passes `wrapperPlugins` values to a mocked plugin manager where they have no effect.

New classes:

- **`WrapperOverheadBenchmarks`** - the missing baseline. Every `wrapped*` benchmark has a `raw*` twin performing the identical operation directly against the same target, so the difference is the wrapper's contribution. Covers per statement, per parameter (`PreparedStatement` setters) and per row (`ResultSet.next()` plus column reads over 1, 100 and 1000 rows - the highest-multiplier path in the driver and the largest previous gap). Also prices `ImportantEventService`, which is enabled by default and records an event on every JDBC call.
- **`RealPluginChainBenchmarks`** - `ConnectionPluginManager.execute` through chains holding one real plugin each: `executionTime`, `connectTime`, `logQuery`, `driverMetaData`, `dataCache`, `sqlParser`, `auroraConnectionTracker`, `auroraStaleDns`, `efm2`, `failover2`, `dev`, plus two combined chains.
- **`PluginServiceBenchmarks`** - the real `PluginServiceImpl` and `SessionStateServiceImpl`, both previously mocked out of every benchmark.
- **`SqlMethodAnalyzerBenchmarks`** - consulted two to four times per SQL-carrying JDBC call; splits by input shape so the early-return and full-parse costs are directly comparable.
- **`RdsUtilsBenchmarks`** - measured on the cached path and with a fresh host name per invocation, since the static matcher cache otherwise hides the real first-seen cost.
- **`ConnectionUrlParserBenchmarks`** - per-connection URL and property work.
- **`StorageBenchmarks`** - `CacheMap`, `ExpirationCache`, `SlidingExpirationCache`, `SlidingExpirationCacheWithCleanupThread`, `StorageServiceImpl`.
- **`HostSelectorBenchmarks`** - all seven `HostSelector` implementations against one shared topology.

`StorageBenchmarks`, `HostSelectorBenchmarks` and `RdsUtilsBenchmarks` carry explicit `contended*` variants under `@Threads`, because several driver caches take a write path on read and the round-robin selector holds a lock across its whole selection - costs that are invisible single-threaded.

### Measurement approach

The target driver is stood in for by `support/FakeJdbc`: plain `java.lang.reflect.Proxy` instances serving constant values. This is deliberate. The pre-existing benchmarks mock the target with Mockito, whose dispatch and invocation recording cost more than the wrapper code under test - which is why they reported milliseconds for work that takes nanoseconds. The stand-in cost sits underneath both sides of every pair, so it cancels out of the difference. It does not cancel out of an absolute number, and the README says so.

Where the driver has a real fallback implementation, the real one is used rather than a stand-in: `UnknownDialect`, `GenericTargetDriverDialect`, the process-wide `MonitorService`, `SimpleConnectionContextServiceImpl` and `ImportantEventService`. Real plugins read these during construction and on the hot path, so standing them in would both crash and misrepresent cost.

`jsqlparser` is added to the benchmark classpath. It is an optional wrapper dependency required at runtime by `sqlParser`, `autoReadWriteSplitting` and `kmsEncryption`; without it those plugins fail with `NoClassDefFoundError` instead of being measured. Version pinned to match `wrapper/build.gradle.kts`.

### Deliberately not covered

Documented in the classes and the README rather than faked:

- Failover, monitor probes, topology refreshes and credential fetches - driven by network events and background threads, not by a JDBC call.
- Connect-only plugins (`iam`, `awsSecretsManager`, `federatedAuth`, `okta`) - their real cost is a network call; on the execute path they would score flat.
- Dialect detection - needs a live database. `PluginServiceBenchmarks` supplies a fixed dialect so the rest of `PluginServiceImpl` can be measured.

## Findings

These came out of the new benchmarks. They are from single-iteration smoke runs, so they are **leads for follow-up, not conclusions**, and nothing here is acted on in this PR.

- **Plugin cost is a cliff, not a gradient.** `makePluginChainFunc` computes `isSubscribed` while explicitly excluding `DefaultConnectionPlugin`, so when nothing subscribes the whole pipeline is bypassed and `executeQuery` costs about 9 ns. The first plugin that subscribes brings it to about 2.5 us, and most of that is the default plugin's SQL analysis rather than the plugin itself. Three plugins together cost about 4.3 us, barely more than one.
- **`sqlParser` costs on the order of 1.7 ms per query**, three orders of magnitude above every other plugin, because jsqlparser re-parses the statement on every execute with no cache.
- **A non-RDS host name costs about 22 us to classify**, roughly 100x the cached RDS path, and the cached and uncached numbers are indistinguishable - the static matcher cache appears not to help the no-match case. Anyone connecting through a custom DNS name pays this repeatedly.
- **`CacheMap.get` degrades about 21x under four threads** on a single hot key (52 ns to 1090 ns), because the read goes through `computeIfPresent` and takes the `ConcurrentHashMap` bin lock. `ExpirationCache.get`, a plain read, stays flat.
- **Round robin costs about 1.5 us against about 0.11 us for random, and about 9.2 us on four threads** - it holds a `ResourceLock` across its whole selection including a sort. Under query-level load balancing that makes selector choice a throughput decision, not just a distribution one.
- **`PluginServiceImpl.getHosts` goes from about 13 ns to about 246 ns once a host-permission entry exists**, which both the custom endpoint and blue/green plugins install.
- **Wrapper overhead itself is modest**: about 76 to 195 ns for `createStatement`, about 129 to 228 ns for `executeQuery`. `PluginServiceImpl`'s query-path methods are 2 to 4 ns field reads - worth stating plainly, since that was previously assumed rather than measured.

## Testing

- `./gradlew :benchmarks:jmhJar` - BUILD SUCCESSFUL.
- Full-suite run excluding `PgCacheBenchmarks`: **149 benchmarks, 0 exceptions, 0 `<failure>` markers**. Before this PR the same command yielded 1 result out of 24.
- Each of the six commits was compiled and smoke-run individually before being committed.

Not verified:

- `PgCacheBenchmarks` was not executed - it needs a live PostgreSQL instance and cache server. Its endpoint handling and fail-fast path were exercised, but not a real run.
- The per-row traversal numbers are still within noise of their raw twins at one iteration and occasionally invert. They need a full-length run before anything is concluded from them; the README warns against quoting smoke-run numbers.
- No unit tests were added. This module has no test source set, and the benchmarks are themselves the verification artefact.

## Notes

No production code under `wrapper/` was changed. Every finding above is recorded in the commit messages for follow-up rather than acted on here.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,6 +95,11 @@ jobs:
             ${{ runner.os }}-gradle-
       - name: 'Run checkstyle and spotbugs'
         run: ./gradlew --build-cache checkstyleMain checkstyleTest spotbugsMain spotbugsTest
+      # The JMH benchmarks live in the `jmh` source set of the :benchmarks module, so neither
+      # `test` nor `checkstyleMain` compiles them. They reach into the wrapper's internal APIs
+      # and silently stop compiling when those change, so compile them here.
+      - name: 'Compile JMH benchmarks'
+        run: ./gradlew --build-cache :benchmarks:jmhJar
 
   # Nullness analysis in enforcing mode: any new Checker Framework finding fails the build.
   # This lives in main.yml (rather than only in run-checker-framework.yml) so it is covered by
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,11 +1,62 @@
 # Benchmarks for AWS Advanced JDBC Wrapper
 
-This directory contains a set of benchmarks for the AWS Advanced JDBC Wrapper.
-These benchmarks measure the overhead from executing JBDC method calls with multiple connection plugins enabled.
-The benchmarks do not measure the performance of target JDBC drivers nor the performance of the failover process.
+This directory contains a set of [JMH](https://github.com/openjdk/jmh) benchmarks for the AWS Advanced JDBC Wrapper.
+These benchmarks measure the overhead the wrapper adds to JDBC method calls with multiple connection plugins enabled.
+They do not measure the performance of target JDBC drivers nor the performance of the failover process.
+
+## Benchmark classes
+
+| Class                               | What it measures                                                                              | Needs a database? |
+|-------------------------------------|-----------------------------------------------------------------------------------------------|-------------------|
+| `ConnectionPluginManagerBenchmarks` | Plugin-chain construction and the `connect` / `execute` / `notifyConnectionChanged` pipelines | No (mocked)       |
+| `PluginBenchmarks`                  | `ConnectionWrapper` lifecycle and statement execution through the wrapper                     | No (mocked)       |
+| `PgCacheBenchmarks`                 | The remote query cache plugin against a real PostgreSQL instance and cache server             | Yes               |
 
 ## Usage
-1. Build the benchmarks with the following command `../gradlew jmhJar`.
-    1. the JAR file will be outputted to `build/libs`
-2. Run the benchmarks with the following command `java -jar build/libs/benchmarks-4.3.0-jmh.jar`.
-    1. you may have to update the command based on the exact version of the produced JAR file
+
+1. Build the benchmark JAR:
+
+   ```shell
+   ../gradlew :benchmarks:jmhJar
+   ```
+
+   The shaded JAR is written to `build/libs/benchmarks-<version>-jmh.jar`.
+
+2. Run the benchmarks that do not require infrastructure:
+
+   ```shell
+   java -jar build/libs/benchmarks-<version>-jmh.jar -e PgCacheBenchmarks
+   ```
+
+   `-e PgCacheBenchmarks` excludes the benchmark that needs a live database and cache server. Without
+   it, that class fails during setup and its results are reported as `<failure>`.
+
+Useful JMH flags while iterating: `-f 1 -wi 1 -i 1` for a fast smoke run, `-rf json -rff result.json`
+to capture machine-readable results, and `-prof gc` to add allocation numbers.
+
+### Running `PgCacheBenchmarks`
+
+This class needs a reachable PostgreSQL instance holding the pre-populated `test` table (see the
+class javadoc for the schema) plus a reachable cache server. Endpoints are supplied as system
+properties, so nothing environment-specific is hardcoded:
+
+```shell
+java -Dbenchmarks.pg.url=jdbc:aws-wrapper:postgresql://my-db:5432/postgres \
+     -Dbenchmarks.cache.rw=my-cache:6379 \
+     -Dbenchmarks.cache.ro=my-cache:6380 \
+     -jar build/libs/benchmarks-<version>-jmh.jar PgCacheBenchmarks
+```
+
+If a property is missing, setup fails immediately with a message naming it rather than timing out
+against a placeholder host name.
+
+## Notes
+
+- The mocked benchmarks measure the wrapper's own overhead. `PluginBenchmarks` drives a
+  `ConnectionWrapper` backed by a mocked `ConnectionPluginManager`, so the `wrapperPlugins` values in
+  its property helpers select *which* JDBC pipeline shape is exercised, not a real plugin chain.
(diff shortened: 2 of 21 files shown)


.github/workflows/main.yml
```diff
@@ -95,6 +95,11 @@ jobs:
             ${{ runner.os }}-gradle-
       - name: 'Run checkstyle and spotbugs'
         run: ./gradlew --build-cache checkstyleMain checkstyleTest spotbugsMain spotbugsTest
+      # The JMH benchmarks live in the `jmh` source set of the :benchmarks module, so neither
+      # `test` nor `checkstyleMain` compiles them. They reach into the wrapper's internal APIs
+      # and silently stop compiling when those change, so compile them here.
+      - name: 'Compile JMH benchmarks'
+        run: ./gradlew --build-cache :benchmarks:jmhJar
```
karenc-bq, line 102: Curious if we also want to run and store the benchmarking result

karenc-bq (approved):

Issue #1367, opened 2025-04-22
jdbc array resultsets cause issues with read-write-splitting plugin with spring boot


treetip:
### Describe the bug

In spring boot 3.4.x `java.sql.Array#getResultSet` causes a "underlying connection has changed" exception when the surrounding transaction/connection is read only.

### Expected Behavior

no exception thrown

### What plugins are used? What other connection properties were set?

F0 profile, postgres 17.4 aurora cluster

### Current Behavior

Reading a jdbc array via its `java.sql.ResultSet` throws exception about the read only connection changing.

### Reproduction Steps

```
@Component
@AllArgsConstructor
public FancyDao {

    private final NamedParameterJdbcTemplate jdbc;

    @Transactional(readOnly = true)
    public List<String> getStuffs() {
        var sql = """
                with stuff(text_col) as (
                  values ('a'),('b'),('c'),('d'),('a')
                )
                select array(select distinct text_col from stuff) as text_values""";
        var result = new ArrayList<String>();
        jdbc.query(sql, rs -> {
            result.addAll(collect(rs.getArray("text_values"), ResultSet::getString, Collectors.toList()));
        });
        return result;
    }

    static <T, A, R> R collect(Array array, RowMapper<T> mapper, Collector<T, A, R> collector) throws SQLException {
        var accumulator = collector.accumulator();
        var container = collector.supplier().get();
        if (array != null) {
            try (var rs = array.getResultSet()) {
                while (rs.next()) {
                    accumulator.accept(container, mapper.mapRow(rs, 2));
                }
            }
        }
        return collector.finisher().apply(container);
    }
}
```

### Possible Solution

Not familiar with the JDBC spec, but the Array data should already be client side at this point, making checking the underlying connection a false positive.

Can also use `java.sql.Array#getArray` to go around the issue, but it's a bit of a trap.

### The AWS Advanced JDBC Driver version used

2.5.6

### JDK version used

temurin 21

### Operating System and version

fedora 42

treetip:
actual exception, up to internal code which starts at the above examples `array.getResultSet()`

```
 java.sql.SQLException: The internal connection has changed since 'org.postgresql.jdbc.PgResultSet@77b31135' was created. This is likely due to failover or read-write splitting functionality. To ensure you are using the updated connection, please re-create Statement and ResultSet objects after failover and/or calling setReadOnly.
[...]

PR #2077, opened 2026-08-05
fix: stop reporting a stale connection for driver-bound JDBC objects

sergiyvamz:
Fixes the false `The internal connection has changed` error raised when operating on a `ResultSet` obtained from `java.sql.Array#getResultSet` (#1367).

### Problem

The wrapper rejected a call on a `Statement` / `ResultSet` / `DatabaseMetaData` whenever the connection the target driver reported for the underlying object was not reference-identical to the wrapper's current connection. With internal connection pools the current connection is a pooled handle (for example `HikariProxyConnection`), while pgjdbc builds the array's `ResultSet` on the physical `PgConnection` underneath it. Same session, different object, so the call was rejected. A pooled XA connection has the same shape, which is why the guard already carried an `isXaTransactionActive()` exemption.

### Change

- New `ConnectionBoundObject`: a wrapper records the internal connection it was created on. Implemented by the `Statement`, `PreparedStatement`, `CallableStatement`, `ResultSet` and `DatabaseMetaData` wrappers. `Rebindable` extends it, and `rebind` updates the recorded connection so a rerouted statement is not considered stale.
- `WrapperUtils` compares the recorded connection against the current one (`executeWithPluginsWithBoundObject` plus a private `checkNotStale`) instead of calling `Statement.getConnection()` on the target driver for every guarded invocation. All 52 call sites of the 25 `JdbcMethod` constants declared with `checkBoundedConnection` now pass the invoking wrapper.
- Removed the XA exemption from the guard. Connection swaps are rejected during an XA branch, so the guard is inert there; the only case it now catches is a statement created before a swap that predates the branch, which is genuinely stale.
- `Statement.getResultSet` and `Statement.getMoreResults` are now declared with `checkBoundedConnection`, matching the `PreparedStatement` and `CallableStatement` equivalents. This is needed because a freshly created `ResultSet` wrapper always records the current connection, so a stale parent has to be caught at the parent.

Follow-ups from review:

- `getConnectionFromSqlObject` now has no production caller. `DefaultConnectionPlugin` and `UnifiedReadWriteSplittingPlugin` compare only a `Connection`, by identity, instead of asking the target driver which connection a `Statement` or `ResultSet` is bound to. A `Connection` is its own bound connection, so that comparison cannot produce the false positive, while a `Statement` or `ResultSet` needs no comparison there: every method the transaction analysis reacts to is declared with `checkBoundedConnection`, so it is rejected before the pipeline runs. This also removes two driver calls (`getConnection`, `isClosed`) from every guarded invocation. The method is `@Deprecated` rather than deleted, since it is public API, and keeps its unit tests.
- The pairing between `checkBoundedConnection` and the bound-object call sites is now enforced, instead of documented only. An assertion in `WrapperUtils` fails fast on a violating call site (evaluated before the connection lock is taken, so it cannot leak the lock; disabled in production, so no running application changes behavior), and `BoundObjectCoverageTest` invokes all 52 guarded method/overload combinations on a wrapper whose connection was swapped out and requires rejection, which also catches a call site passing the wrong bound object.

Detection semantics are otherwise unchanged: it is still connection reference identity, so an object stays valid when read/write splitting switches away and back to the same cached connection. A swap-generation counter was considered and rejected for exactly that reason.

### Behavior notes

- A stale *and already closed* object now surfaces the connection-changed error where it previously surfaced the target driver's "already closed" error. Both are `SQLException`.
- Methods not declared with `checkBoundedConnection` (`close`, `next`, `cancel`) still work on a stale object, so an application can release it and cancellation keeps working during a switch.

- Read/write splitting with XA: a statement from a pooled XA connection reports the physical connection, so routing used to be skipped for it. Routing now applies, and where routing is wanted but the statement cannot be rebound, the existing one-time reuse warning can now appear for such a statement. A switch during an active XA branch is still refused by `TransactionAwareGate`. This is the intended consequence of not trusting the driver-reported binding, and is the change in this PR most worth a second opinion.

### Testing

- New `ConnectionBoundObjectTest` (9 tests) drives the real wrappers through the pipeline: the recorded connection on creation for all five wrapper types, rejection of every guarded method after a swap (with the message asserted against the resource bundle), unguarded methods still usable, validity restored after switching back to the same connection, rejection during an XA branch, rebinding clearing staleness, and the `Array#getResultSet` scenario end to end.
- `WrapperUtilsTest` updated for the new mechanism, including the pooled-handle vs physical-connection case.
- Verified the new tests actually cover the bug: restoring the old driver-based comparison produces 9 failures, with the array test reproducing the reported exception.
- 1271 unit tests pass. `checkstyleMain` and `checkstyleTest` are clean. NullnessChecker (`-PenableCheckerFramework`) reports no findings.
- New read/write splitting integration test `test_pooledConnection_arrayResultSetAfterSetReadOnly` (PostgreSQL only) reproduces the reported scenario end to end: an internal connection pool makes the current connection a pooled handle, then the array's `ResultSet` is read and a connection-checked method is called on it.

### Not run locally

The new integration test needs a live Aurora PostgreSQL cluster, so it has not been executed here; it runs with the PostgreSQL Aurora integration jobs.

### API note

`Rebindable` now extends `ConnectionBoundObject`, so an external implementer of that interface would need two additional methods.


wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
```diff
@@ -372,21 +419,7 @@ private static <T, E extends Exception> T doExecuteWithPlugins(
...
-      // holds, even though it is the same physical session. Skip while an XA transaction is active.
-      if (jdbcMethod.shouldLockConnection && jdbcMethod.checkBoundedConnection
-          && !connectionWrapper.getServicesContainer().getPluginService().isXaTransactionActive()) {
-        final Connection conn = WrapperUtils.getConnectionFromSqlObject(methodInvokeOn);
-        if (conn != null && conn != connectionWrapper.getCurrentConnection()) {
-          throw new SQLException(Messages.get(
-              "ConnectionPluginManager.invokedAgainstOldConnection", new Object[]{methodInvokeOn}));
-        }
-      }
+      checkNotStale(connectionWrapper, boundObject, methodInvokeOn, jdbcMethod);
```
sophia-bq, line 445: From my understanding we are replacing the calls to getConnectionFromSqlObject with ConnectionBoundObject usage. I'm curious what the plan is around the remaining usages of getConnectionFromSqlObject which exist in the DefaultPlugin and the UnifiedReadWriteSplittingPlugin

wrapper/src/main/java/software/amazon/jdbc/JdbcMethod.java
```diff
@@ -1300,11 +1300,11 @@ public enum JdbcMethod {
   STATEMENT_EXECUTE(607, "Statement.execute",
 false, false, true, true),
   STATEMENT_GETRESULTSET(608, "Statement.getResultSet",
-false, true, true, false),
+false, true, true, true),
```
sophia-bq, line 1303: Oh good catch to adjust these!

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
```diff
@@ -225,6 +226,11 @@ public static <E extends Exception> void runWithPlugins(
         jdbcMethodArgs);
   }
 
+  /**
+   * Runs a method that does not throw a checked exception through the plugin pipeline. There is no
+   * bound object, so no stale-object check is performed; a {@link JdbcMethod} declared with
+   * {@code checkBoundedConnection} must use {@link #executeWithPluginsWithBoundObject} instead.
+   */
```
sophia-bq, line 234: Do we have any check that this requirement is being honoured? It is unlikely but if a JdbcMethod is declared with checkBoundedConnection and goes through executeWithPlugins instead of executeWithPluginsWithBoundObject it could cause issues

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
sergiyvamz: getConnectionFromSqlObject should be completely retired. Let me verify it and remove left overs.

wrapper/src/main/java/software/amazon/jdbc/util/WrapperUtils.java
sergiyvamz: I'll add a guard to prevent such cases.

sophia-bq (approved):

Issue #2083, opened 2026-08-07
[Bug/Question] autoSimpleReadWriteSplitting blocks host switch to WRITER when autoCommit(false) is called before write query in non-transactional Spring Data JPA


KP-story:
### Description
When using the `autoSimpleReadWriteSplitting` plugin in combination with Spring Data JPA without an explicit `@Transactional` annotation on a write operation (`repository.save()`), statement execution fails with MySQL error 1290 (`--super-read-only`).

The issue occurs because Hibernate calls `Connection.setAutoCommit(false)` while the connection is currently routed to a Reader endpoint. Since `setAutoCommit(false)` resolves to `NO_DECISION`, the connection remains on the Reader host and gets pinned by `TransactionAwareGate`. When the subsequent `INSERT` query is prepared, `SqlRoutingSignal` correctly identifies the query type as `WRITER`, but `UnifiedReadWriteSplittingPlugin` blocks the host switch due to active transaction pinning on the Reader node.

---

### Environment
* **AWS Advanced JDBC Driver  master
* **MySQL Connector/J Version:** `9.7.0`
* **JSqlParser Version:** `4.9`
* **Spring Boot Version:** `3.4.4` (`spring-boot-starter-data-jpa`)
* **Java Version:** Java 17
* **Plugins Configured:** `sqlParser,autoSimpleReadWriteSplitting`

---
public Map<String, Object> saveNoTransaction(String name) {
    final TestEntity entity = new TestEntity();
    entity.setName(name);

    // Fails here with MySQL Error 1290
    final TestEntity saved = testRepository.save(entity);
    return Map.of("status", "success");
}
### Configuration
```yaml
spring:
  datasource:
    url: jdbc:aws-wrapper:mysql://localhost:3306/test

DEBUG ... autoSimpleReadWriteSplitting : AutoSimpleReadWriteSplittingPlugin routing attempt: method='Connection.setAutoCommit', currentHost='localhost:3307', isInTransaction=false.
DEBUG ... SqlRoutingSignal : SqlRoutingSignal abstains on plain Statement execute: method='Connection.setAutoCommit', resolvedRole=NO_DECISION.
TRACE ... SessionStateServiceImpl : autoCommit: true -> false

DEBUG ... autoSimpleReadWriteSplitting : AutoSimpleReadWriteSplittingPlugin routing attempt: method='Connection.prepareStatement', currentHost='localhost:3307', isInTransaction=false.
[...]

PR #2094, opened 2026-08-12
feat: add assumeWriteTransaction for SQL-routed read/write splitting

sergiyvamz:
## Summary

Adds an opt-in `assumeWriteTransaction` parameter so the SQL-routing read/write splitting plugins can route a framework-managed read-write transaction to the writer, and documents the `setReadOnly`/`setAutoCommit` sequence that makes routing deterministic.

Follow-up to #2086 and the second scenario reported in #2083, where a `@Transactional` method that calls `findAll()` before `save()` was observed using two different physical connections for one transaction.

## Problem

Transaction managers announce only one of the two transaction kinds:

| Transaction | What the driver observes |
|---|---|
| Spring `@Transactional(readOnly = true)` | `setReadOnly(true)`, then `setAutoCommit(false)` |
| Spring `@Transactional` (default, `readOnly = false`) | `setAutoCommit(false)` only |
| MyBatis, jOOQ, plain JDBC | `setAutoCommit(false)` only |

A read-write transaction is expressed by the absence of a read-only declaration, which is indistinguishable from a plain non-transactional connection. The `auto*` plugins therefore route the first statement of such a transaction on its SQL alone: a leading `SELECT` goes to a reader, the transaction physically begins there, and the write that follows cannot be served (MySQL error 1290, `--super-read-only`). Whether it happens depends on where the pooled connection happened to be, so it shows up intermittently.

The `setReadOnly`-driven plugins are not affected in the same way: with no signal they make no routing decision, so there is no inferred decision to correct.

## Change

`assumeWriteTransaction` (default `false`). When enabled, `SqlRoutingSignal` routes a plain read to the writer if all of the following hold:

- the connection was not declared read-only (`setReadOnly(true)` never called, or `setReadOnly(false)` called), and
- a transaction is open, or autocommit is disabled so the next statement opens one.

Autocommit-off counts as a transaction here on purpose: that is the moment the decision has to be made, because the statement being routed is the one that physically opens the transaction.

Unchanged in every other case:

- reads outside a transaction are still offloaded to readers;
- read-only transactions still run on a reader, since `setReadOnly(true)` states read intent;
- an explicit `/*@reader*/` or `/*@keep*/` hint still wins;
- writes, `SELECT ... FOR UPDATE`, and unparseable SQL already route to the writer;
- with the flag off, the routing path is bit-for-bit the previous one (the check short-circuits before reading any session state).

The trade-off is that transactional reads are no longer offloaded unless the transaction is marked read-only, which is why it is opt-in and recommended rather than default. Only the four `auto*` plugins consult it; `readWriteSplitting`, `srw`, and the `gdb` `setReadOnly`-driven variants ignore it.

## Docs

`UsingTheAutoReadWriteSplittingPlugin.md` gains a rewritten "Transactions and autocommit" section covering the missing write-intent signal, the new parameter, and the recommended call sequence for Spring/JPA and for plain JDBC, for both read-only and read-write transactions. Two ordering rules are called out: declare the role before `setAutoCommit(false)` (a `setReadOnly(true)` issued afterwards is vetoed by the transaction guard, so the read-only transaction silently runs on the writer), and never call `setReadOnly` while a transaction is open (SQL state 25001). The sibling `auto*` plugin docs, the `readWriteSplitting` Spring section, and the configuration assistant get cross-references.

## Backward compatibility

Additive only. New property defaults to `false`; no public API removed, renamed, or re-signatured. `SqlRoutingSignal()` is kept and delegates to the new `SqlRoutingSignal(boolean)`. No enum ordering, plugin code, or existing property changed.

## Testing

- `RoutingSignalsTest`: 11 new cases for the signal - disabled/enabled, autocommit off/on/unset, open transaction, explicit `setReadOnly(true)` and `(false)`, reader hint precedence, unreadable session state, bound-statement path, and that a write query never reads session state.
- `WriteTransactionIntentTest` (new): plugin-level flow through `AutoReadWriteSplittingPlugin` starting from a connection left on a reader - enabled switches the leading read to the writer, disabled does not, read-only transactions stay, reads outside a transaction stay, reader hints stay, and an already-open transaction is never moved.
- `AutoReadWriteSplittingTests`: two integration tests - the read-then-write transaction now runs entirely on the writer (including the write), and a `readOnly` transaction still stays on the reader with the flag enabled.

Gates run locally: `:aws-advanced-jdbc-wrapper:test` (1307 passed, 0 failed), `checkstyleMain`, `checkstyleTest`, and `compileJava -PenableCheckerFramework`. The integration tests require a provisioned cluster and were not executed here.

## Known gap left for a follow-up

Transaction tracking itself is still blind to prepared statements: `SqlMethodAnalyzer.doesOpenTransaction` needs the SQL in the method arguments, and `PreparedStatement.executeQuery()` carries none, so `isInTransaction()` never becomes true for Hibernate/JPA (or `JdbcTemplate`, MyBatis, jOOQ) workloads. That independently disables the `setReadOnly`-in-transaction guard, `rollbackOnSwitch`, and failover's transaction-resolution-unknown handling. It is a driver-wide behavior change that deserves its own PR and release note; `assumeWriteTransaction` fixes the reported routing problem without depending on it.


docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
```diff
@@ -129,8 +129,98 @@ This is required because switching the underlying physical connection in the mid
...
+
+Everything else is unchanged:
+
+- Reads outside a transaction (autocommit on) are still offloaded to readers.
+- Read-only transactions still run on a reader, because `setReadOnly(true)` states read intent explicitly.
+- An explicit `/*@reader*/` or `/*@keep*/` hint still wins, so an individual read inside a read-write transaction can be sent to a reader deliberately.
+- Writes, `SELECT ... FOR UPDATE`, and unparseable statements already route to the writer.
+
+The trade-off is direct: transactional reads are no longer offloaded to readers unless the transaction is marked read-only. This is why the setting is opt-in. It is recommended for Spring/JPA applications, where `@Transactional(readOnly = true)` is available to mark read-only work, and for any application that wraps mixed read-then-write logic in a transaction. It applies to the SQL-routing plugins (`autoReadWriteSplitting`, `autoSimpleReadWriteSplitting`, `gdbAutoReadWriteSplitting`, `gdbAutoSimpleReadWriteSplitting`); the `setReadOnly`-driven plugins (`readWriteSplitting`, `srw`, `gdbReadWriteSplitting`, `gdbSimpleReadWriteSplitting`) ignore it, since they only ever route where the application tells them to.
```
kenrickyap, line 177: Not really a concern, mostly just curious on what the reasoning is to have the default be `false`. With `assumeWriteTransaction` set to false users with a transaction manager would hit the read-only error unless they know to opt in. Why did we prefer the risk of hitting an error over the risk of reducing read offloading? Is it mostly to maintain the existing behavior?

kenrickyap (approved):

docs/using-the-jdbc-driver/using-plugins/UsingTheAutoReadWriteSplittingPlugin.md
sergiyvamz: Maintaining back compatibility is the only reason here. assumeWriteTransaction=true sounds reasonable for all Spring apps, however it's not that obvious for other type of applications and framework. I think it's better to leave users an option.

PR #2097, opened 2026-08-17
feat: opt-in connect retry to bridge Secrets Manager rotation window

sergiyvamz:
## Summary

The `awsSecretsManager` plugin cannot connect during a Secrets Manager rotation window. This adds an opt-in, configurable connect retry with capped exponential backoff so it can poll across that window instead of giving up.

This is the JDBC counterpart of aws/aws-advanced-go-wrapper#470, which reports the same gap in the Go wrapper. The JDBC `connectInternal` had the same shape as the Go `connectInternal` quoted there.

## The problem

A rotation runs as `createSecret` -> `setSecret` -> `testSecret` -> `finishSecret`. Between `setSecret` (the database password has already been changed) and `finishSecret` (`AWSCURRENT` is promoted to the new version) the database expects the new password while `GetSecretValue` still returns the old one. With RDS managed rotation this window has been reported at roughly a minute.

The plugin re-fetched and retried at most once, and that single retry resolved `AWSCURRENT` to the same old secret, so it failed too. Existing connections kept working, but every new physical connection failed for the duration of the window: application startup, connection pool growth, and pool max-lifetime recycling. The failure surfaced as a login error (`28P01` / `28000`).

The `!secretWasFetched` guard made the worst case sharper. `updateSecret(hostSpec, false)` returns `true` whenever it actually called the service, so on a cold cache (new JVM, Lambda cold start) the retry branch was skipped entirely and there was exactly one attempt.

Nothing else in the driver absorbed this. Every other retry loop that inspects `isLoginException` treats it as terminal and rethrows immediately - `AuroraInitialConnectionStrategyPlugin`, `LimitlessRouterServiceImpl`, `SubstituteConnectRouting`, `TopologyReaderResolver`, `EndpointConnectionVerifier`, `ClusterTopologyMonitorImpl` - so the login error went straight to the caller, which is usually a pool's `getConnection`.

## The change

Two new properties on `AwsSecretsManagerConnectionPlugin`, inherited by `awsSecretsManager2`:

| Property | Default | Description |
|---|---|---|
| `secretsManagerConnectRetryTimeoutMs` | `0` | Total time budget for retrying a connection that failed to log in, re-fetching the credentials before each retry. `0` disables retrying. |
| `secretsManagerConnectRetryIntervalMs` | `1000` | Initial delay before a retry. Doubles after every failed attempt, capped at 30000ms and at the remaining budget. |

Naming follows the existing `srwConnectRetryTimeoutMs` / `srwConnectRetryIntervalMs` precedent in `SimpleReadWriteSplittingPlugin`.

`connectInternal` now dispatches on the budget:

- `connectWithSingleRetry` holds the previous body verbatim and is selected whenever the budget is `0`.
- `connectWithRetryBudget` repeats "force a re-fetch, then reconnect" until the connection succeeds or the budget runs out. It ignores `updateSecret`'s return value, so retries also apply to the empty-cache case that the `!secretWasFetched` guard skipped.

Only login failures are retried. Any other error is reported immediately, since re-fetching would not help. A transient `GetSecretValue` failure mid-loop no longer consumes the whole budget, and the login error stays the reported cause. Interruption restores the interrupt flag and surfaces as a `SQLException`.

The change is mirrored into `wrapper/src/main/java17/.../AwsSecretsManagerConnectionPlugin.java`. That file is a multi-release override packaged into `META-INF/versions/17` and replaces the main class on Java 17+ runtimes, so changing only the main source set would have shipped a feature that silently does nothing for most users.

Also in scope, both small and adjacent to the edited method:

- The two hardcoded empty-property exception messages are moved into the resource bundle with identical rendered text, per the message externalization convention.
- The plugin doc listed the cache TTL parameter as `secretsManagerExpirationSec`; the code registers `secretsManagerExpirationTimeSec`. Corrected.

## Backward compatibility

`secretsManagerConnectRetryTimeoutMs` defaults to `0`, which selects a verbatim copy of the previous connect body: same `&& !secretWasFetched` guard, same two `connectFunc.call()` sites, same catch ordering, same rethrow. Existing configurations behave exactly as before. The change is additive - no public API removed, renamed, or re-signatured.

The only delta on the default path is the message source for the two empty-property exceptions. `Messages.get` always runs `MessageFormat`, so the `''` escaping in the bundle renders the original text character for character.

A `RuntimeException` from a misconfigured endpoint still propagates uncaught in both paths, matching the original where `updateSecret` sat outside the `try`.

## Notes

Each retry issues one `GetSecretValue` call and each attempt blocks the calling thread for the backoff interval, both documented. Sleeping on the connect path is new for this plugin but opt-in, and `EndpointConnectionVerifier` already does it for `srwConnectRetryIntervalMs`.

The docs also point out that this is a client-side mitigation: RDS Proxy or a multi-user rotation strategy avoids the window rather than waiting it out.

## Testing

Six new unit tests in `AwsSecretsManagerConnectionPluginTest` cover retry-until-success, the empty-cache case, budget exhaustion, a fetch failure mid-retry, non-login errors bypassing the retry, and retries being off by default.

Verified locally:

- `./gradlew :aws-advanced-jdbc-wrapper:compileJava :aws-advanced-jdbc-wrapper:compileJava17Java` - both build
- `./gradlew :aws-advanced-jdbc-wrapper:test` - 1329 tests, 0 failures, 2 pre-existing skips
- `./gradlew :aws-advanced-jdbc-wrapper:checkstyleMain :aws-advanced-jdbc-wrapper:checkstyleJava17 :aws-advanced-jdbc-wrapper:checkstyleTest` - pass
- `./gradlew :aws-advanced-jdbc-wrapper:compileJava -PenableCheckerFramework` - clean

Not verified: there is no integration test against a live rotating secret, so the real-world window timing rests on the reporter's measurement in the linked issue. The retry timing is covered only by unit tests using short budgets.


docs/using-the-jdbc-driver/using-plugins/UsingTheAwsSecretsManagerPlugin.md
```diff
@@ -41,12 +41,42 @@ The following properties are required for the AWS Secrets Manager Connection Plu
...
 | `secretsManagerEndpoint`               | String  |                            No                            | Set this value to be the endpoint override to retrieve your secret from. This parameter value should be in the form of a URL, with a valid protocol (ex. `http://`) and domain (ex. `localhost`). A port number is not required. | `http://localhost:1234` | `null`        |
-| `secretsManagerExpirationSec`          | Integer |                            No                            | This property sets the time in seconds that secrets are cached before it is re-fetched.                                                                                                                                          | `600`                   | `870`         |
+| `secretsManagerExpirationTimeSec`      | Integer |                            No                            | This property sets the time in seconds that secrets are cached before it is re-fetched.                                                                                                                                          | `600`                   | `870`         |
```
sergiyvamz, line 44: That is not a parameter renaming. It's just a doc fix.

docs/JDBC-WRAPPER-CONFIGURATION-ASSISTANT.md
```diff
@@ -801,6 +801,8 @@ Reads `username`/`password` from an AWS Secrets Manager secret (JSON format).
 | `secretsManagerSecretUsernameProperty` | `username` | Key in JSON secret containing username. |
 | `secretsManagerSecretPasswordProperty` | `password` | Key in JSON secret containing password. |
 | `secretsManagerExpirationTimeSec` | (driver-defined) | Secret cache TTL. |
+| `secretsManagerConnectRetryTimeoutMs` | `0` | Time budget for retrying a login failure while re-fetching the secret. Set above `0` to bridge a rotation window. |
```
sophia-bq: Should there be more of a note here? As well across the docs for this change should there be a note that these properties are only available on the next wrapper version?

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
sophia-bq, line 549: Currently I think we have one new client per attempt - with retry the impact of this increases - is there a way to reuse a SecretsManagerClient?

sergiyvamz: I can investigate on this idea. It's probably out of scope of this fix.

sophia-bq (approved):

PR #2101, opened 2026-08-18
fix: strip node-selection properties from monitoring connections

### LINKED ISSUE #2096: Blue/Green plugin never detects switchover completion on single Postgres RDS DB instances
## Summary

On RDS for PostgreSQL **DB instances** (not Aurora), the Blue/Green plugin's completion
detection does not work: the `rds_tools` status metadata it polls is frozen on the old
instance and deleted from the new one at switchover. The plugin therefore always holds
suspended traffic until the `bgSwitchoverTimeoutMs` timer expires, and the IP-substitution
routing that is supposed to beat stale DNS never activates (the timer-expiry path installs
no routing). Net result: with the plugin, client downtime is deterministically
`bgSwitchoverTimeoutMs`, even when the database itself completed the switchover in ~2 seconds.
This is the same symptom as #1920; the 4.1.0 fix made the timer escape hatch reliable, but
the underlying detection gap on RDS instances remains.

## Environment

- aws-advanced-jdbc-wrapper 4.3.0, `wrapperPlugins=bg`, `bgSwitchoverTimeoutMs=60000`
- RDS for PostgreSQL 18.3, single DB instance (not Aurora, not a Multi-AZ DB cluster)
- `rds_tools` extension v1.9 installed, with the documented `USAGE`/`EXECUTE` grants applied
  before the Blue/Green deployment was created
- Blue/Green deployment for a parameter-group change (physical replication); plugin had
  ~30 minutes of dwell in `CREATED` to collect topology and IP addresses

## What we observed (timestamps from one switchover)

| Time offset | Event | Source |
|---|---|---|
| 0s | Switchover starts; plugin reads `PREPARATION` then `IN_PROGRESS`, suspends all traffic | plugin DEBUG logs |
| ~2s | RDS event: "write downtime during the switchover lasted approximately 2 seconds" | RDS events |
| ~8s | RDS event: instances renamed (green → blue name, blue → `-old1`) | RDS events |
| 13–25s | Plugin logs "Blue DNS updated" (per JVM), but still does not switch | plugin summary log |
| 13–59s | Monitors are connected and polling, but only ever read `IN_PROGRESS` | plugin DEBUG logs |
| 59s | "Blue/Green switchover has timed out." - timer forces `COMPLETED`, traffic released | plugin WARN log |
| ~60s | RDS event: "Switchover completed on blue/green deployment" | RDS events |
| 65s | Monitors finally read `NOT_CREATED` (status table now empty) | plugin DEBUG logs |

Application-side write outage: **59.2s** - the full timer value, not the ~2s database outage.
A second run with the default 180s timer produced a 187s outage.

## Cause

The plugin's only exits from `IN_PROGRESS` are (a) a monitor reading
`SWITCHOVER_IN_POST_PROCESSING` / `SWITCHOVER_COMPLETED` from
`rds_tools.show_topology()`, or (b) timer expiry. On RDS DB instances, (a) is impossible:

1. **The old instance's status freezes at `SWITCHOVER_IN_PROGRESS`.** It is severed from
   the deployment at switchover start and never receives another status update. (The
   monitor code itself comments that this is expected.)
2. **The new primary's status is deleted.** Queried live *while the Blue/Green deployment
   object still existed*, `rds_tools.show_topology()` on the new primary returned **0 rows**.
   the status is instance-local state that RDS's
   orchestration clears at finalization rather than advancing through
   `POST`/`COMPLETED`.

On RDS DB instances, the plugin currently *increases* downtime relative to a plain driver
with a low DNS TTL: it converts a ~2s database outage plus ~15–25s of DNS propagation into
a fixed hold of `bgSwitchoverTimeoutMs`.

sergiyvamz:
## Summary

Fixes #2096, where the Blue/Green plugin held all client traffic for the full `bgSwitchoverTimeoutMs` even though the database completed its switchover in about two seconds. Reported downtime was 59.2s with a 60s timeout, and 187s with the default 180s timeout.

Credit to @eddking for diagnosing the root cause.

## Root cause

`targetServerType=primary` was reaching the plugin's monitoring connections. With that property the PostgreSQL JDBC driver runs `SHOW transaction_read_only` after authenticating and refuses read-only servers. A green node is a physical replica until it is promoted, so:

- Both monitors start against the current host (blue). Blue is a primary, so that connects, and the green monitor learns the green endpoint from the deployment status.
- It reconnects to the green endpoint, and *that* connection is rejected.
- The green monitor therefore never reads green's status, while the blue monitor is pinned by IP to the old instance, whose metadata stops being updated once it is detached from the deployment.
- The plugin only ever observes `SWITCHOVER_IN_PROGRESS`, leaving the switchover timer as the only exit.

Monitoring connections build their properties by copying the user's connection properties wholesale, so any such restriction is inherited. This was never Blue/Green specific: host monitoring (EFM v1 and v2), topology monitoring, fastest-response, and Limitless all copy properties the same way and also connect to readers, so all of them were silently disabled by the same misconfiguration.

## 1. Monitoring connections no longer inherit node-selection restrictions

Host selection is the wrapper's job. A monitoring connection is opened against one specific node the wrapper has already chosen, often a reader or a replica on purpose, so a target driver setting that rejects a node by role can only break it.

Adds `TargetDriverDialect.removeHostSelectionProperties`, which returns the names of the properties it removed and **defaults to removing nothing**, so other target drivers and any implementation outside this library are unaffected. Implemented for PostgreSQL, which is the only place `targetServerType` is named - consistent with `updateInternalState`, which already handles `currentSchema` and `readOnly` there.

Applied at all six places where monitoring properties are derived from user properties. The removal runs **before** the `<prefix>` monitoring overrides are applied, so only the inherited application-level value is dropped and an explicit `blue-green-monitoring-targetServerType` (or `monitoring-`, `topology-monitoring-`, `frt-`, `limitless-router-monitor-`) still takes effect.

Removal logs a `WARNING` naming the properties, once per property per JVM, so the misconfiguration is visible rather than silently corrected.

## 2. Release held traffic once the blue endpoints resolve to new IPs

A safety net for any situation where a status after `SWITCHOVER_IN_PROGRESS` never reaches the driver, whatever the reason.

`updateDnsFlags()` already sets `blueDnsUpdateCompleted` when every host in the blue start topology resolves to a *different* IP address than it did before the switchover. That can only happen after the deployment has repointed those endpoints at the new topology, at which point holding traffic protects nothing - a new connection to a blue endpoint would reach the new instance anyway. So while the phase is `IN_PROGRESS` and that flag is set, proceed as if `POST` had been reported, releasing the suspended calls and installing the substitution routing.

Gated by a new `bgReleaseOnBlueDnsUpdate` property, default `true`.

**Dormant on a healthy switchover.** Measured on both Blue/Green integration environments, the metadata reports `POST` well before DNS propagates, so the metadata path always wins and this check never fires:

| Environment | metadata `POST` | blue DNS flip |
|---|---|---|
| RDS instance, Multi-AZ, PG latest | 9498 ms | 20485 ms |
| RDS instance, single-AZ, PG 18.4 | 6997 ms | 20163 ms |

## 3. Stop treating NOT_CREATED as a rollback

`updatePhase()` set `rollback = true` whenever the target monitor reported a phase lower than the previous one. But a monitor reports `NOT_CREATED` both when the deployment is genuinely absent and when its status simply cannot be read - the metadata view is missing, the query fails to parse, or the server reports an error retrieving it. All of those are expected once a switchover tears the deployment down.

Treating them as a rollback returns `getStatusOfCreated()`, which reinstates the pre-switchover routing and actively steers clients back to the **old** topology. That was reachable in the reported scenario, where the phase stayed pinned at `IN_PROGRESS` and the deployment metadata was then removed. A genuine rollback reports the deployment as `AVAILABLE` again, which maps to `CREATED`, so excluding `NOT_CREATED` loses no real signal. Not gated - a straight bug fix.

## Testing

Two new unit test classes, including the first unit tests for `BlueGreenStatusProvider`:

- `HostSelectionPropertiesTest` - PostgreSQL removes `targetServerType` and nothing else; other dialects remove nothing; the `PropertyUtils` helper delegates correctly and leaves timeouts intact.
- `BlueGreenStatusProviderTest` - monitoring properties drop the restriction while an explicit monitoring override survives; traffic stays suspended while blue DNS still points at the old topology, when `bgReleaseOnBlueDnsUpdate` is disabled, and during a rollback; `NOT_CREATED` does not trigger a rollback while a real `CREATED` regression still does; phase monotonicity is preserved.

Gates run locally, all passing:

- `./gradlew :aws-advanced-jdbc-wrapper:test` - 1353 tests, 0 failures
- `./gradlew :aws-advanced-jdbc-wrapper:checkstyleMain :aws-advanced-jdbc-wrapper:checkstyleTest`
- `./gradlew :aws-advanced-jdbc-wrapper:compileJava -PenableCheckerFramework`

Not yet run: the Blue/Green integration tests, which need live AWS resources. Worth running before merge to confirm both environments still reach `POST` from metadata at 7-9s with an unchanged traffic-suspension window, i.e. that change 2 stays dormant on the healthy path.

Suggested follow-up: adding `targetServerType=primary` to the PostgreSQL Blue/Green integration test properties would turn this into a true regression test for #2096, failing before change 1 and passing after. That requires the traffic-suspension assertion from #2098 to be meaningful.

## Backward compatibility

- `TargetDriverDialect.removeHostSelectionProperties` is a **new `default` method** on a public interface. Existing implementations, including any outside this library, compile and behave exactly as before because the default removes nothing.
- `bgReleaseOnBlueDnsUpdate` is additive with a default.
- No signatures changed, no enum values added or reordered, no properties or plugin codes renamed or removed.
- Change 1 is a deliberate behavior change for monitoring connections only, scoped to properties a dialect explicitly nominates. Change 2 is property-gated. Change 3 alters behavior only for a case that was previously mishandled.

## Notes

Two new resource bundle messages: `PropertyUtils.hostSelectionPropertiesRemoved` and `bgd.releaseTrafficOnBlueDnsUpdate`. The latter is logged once per switchover and recorded in the switchover summary table as "Traffic released on blue DNS update", rather than on every status update, to keep the monitoring path quiet.

Documentation updated in `UsingTheBlueGreenPlugin.md` (property table plus a note on monitoring connections) and `JDBC-WRAPPER-CONFIGURATION-ASSISTANT.md` (property table, Blue/Green guidance, and a new anti-pattern row covering all affected monitor prefixes).


docs/JDBC-WRAPPER-CONFIGURATION-ASSISTANT.md
```diff
@@ -1065,6 +1065,7 @@ Tracks Blue/Green deployment status and routes connections to the right side dur
...
 - **Pre-switchover readiness signal (since 4.3.0):** the plugin logs a one-time `INFO` event, `[bgdId: '<id>'] Blue/Green target topology recognized`, as soon as it has both the blue and green topologies and the blue → green node mapping. It also prints that mapping and the resolved node IPs. Wait for this event before initiating a manual switchover — no non-default log level is required. On versions before 4.3.0 the same check requires `FINE`/`FINEST` logs — see §13.5.
+- **Do not restrict node selection in the target driver.** Parameters such as the PostgreSQL JDBC driver's `targetServerType=primary` make the driver reject a node by role. The wrapper strips these from its monitoring connections (a `WARNING` is logged when it does), because the green node is a replica until promotion and the monitor would otherwise never connect — leaving traffic held until `bgSwitchoverTimeoutMs` elapsed. Let the wrapper handle host selection instead. On versions without this behavior, override it per monitor with `blue-green-monitoring-targetServerType=any`.
```
sophia-bq: Let's add the version this is introduced in so it can properly apply the last statement

sophia-bq (approved):

PR #2111, opened 2026-08-31
fix: allow topology monitor context recovery

### LINKED ISSUE #2110: Shared topology monitor retains first connection context across logical databases
## Summary

A topology monitor shared by `clusterId` permanently retains the first connection's database and credential context. When that context later becomes invalid, the monitor can remain unable to recover while otherwise valid connections to other logical databases on the same Aurora cluster block waiting for topology refresh.

This is not the different-physical-cluster collision described in #1827: these connections correctly share one physical topology and failover domain. The defect is that cluster-scoped topology is coupled to a shorter-lived, first-caller connection context.

The immediate workarounds are either `wrapperDialect=pg`, which avoids topology monitoring but gives up topology-aware failover and read/write splitting, or a distinct `clusterId` per database/credential context, which preserves those capabilities but duplicates monitors and topology queries. The proposed fix keeps one topology monitor per physical cluster while allowing a disconnected monitor to recover using a current caller's valid context.

### Describe the bug

`RdsHostListProvider` shares `ClusterTopologyMonitorImpl` instances using only `clusterId`. The first connection that creates a monitor supplies the monitor's `initialHostSpec`, original URL, database, user, and connection properties. Later connections with the same `clusterId` reuse that monitor, but cannot replace or rebind its connection context.

This is unsafe when connections correctly share one physical Aurora cluster and topology, but target logical databases or credential contexts with independent lifecycles. If the first database is removed or its credentials become invalid, the shared monitor continues attempting recovery with that stale context. Otherwise valid connections to another database on the same cluster can then block while waiting for topology refresh.

This differs from #1827: the connections target the same physical cluster and therefore should share topology. The collision is between the cluster-scoped topology lifecycle and the shorter-lived database/credential lifecycle captured by the monitor.

The problem was observed in a five-minute Java Flight Recorder:

- 86,561 waits included `ClusterTopologyMonitorImpl.waitForTopologyUpdate`.
- Aggregate wait time was 86,596.6 seconds, approximately 289 concurrent waiters.
- Individual waits were almost exactly one second.
- Five thread dumps showed 242-292 callers in the same path.
- One `ctmi-1` monitor remained in panic mode without cached topology.
- No node-monitor threads were present, consistent with failure to establish the initial monitoring connection.
- The JVM was not CPU- or garbage-collection bound.

Source inspection of 4.4.0 and current `main` shows that `RdsHostListProvider.getOrCreateMonitor()` is still keyed only by `clusterId`, while the constructed monitor permanently retains the first caller's `initialHostSpec` and copied connection properties.

Related reports cover individual parts of the failure:

- #1827: `clusterId` cache collision across different physical clusters.
- #2020 / #2021: a topology monitor failed IAM authentication because it did not inherit the intended plugin configuration.
- #2096 / #2101: monitoring connections inherited caller properties that prevented them connecting to required nodes.
- #2035 / #2036: unrecoverable topology-monitor connections caused long synchronous waits and pool exhaustion.

None of those changes lets a shared monitor replace stale database or credential context with a currently healthy caller.

### Expected Behavior

Connections that share one physical cluster should be able to share cluster topology without permanently sharing the first connection's database and credential lifecycle.

If the context that created a topology monitor can no longer connect, a healthy caller for the same `clusterId` should be able to replace or rebind the monitor's connection context. A failed monitor context must not indefinitely delay otherwise valid connections to the same physical cluster.

`wrapperPlugins=iam`; no explicit `wrapperDialect` or `clusterId`, so the Aurora PostgreSQL dialect was detected and `clusterId` defaulted to `1`. A custom credentials handler selected an IAM provider from a connection-specific property. Connections used the same standard Aurora cluster endpoint but different logical databases and credential contexts.

### Current Behavior

The first connection creates the shared topology monitor. The monitor copies that connection's properties and retains its `initialHostSpec`. When the retained database or credential context becomes invalid, `openAnyConnectionAndUpdateTopology()` repeatedly attempts to reconnect with that stale context.

Connection attempts using another valid database and credential context call `RdsHostListProvider.forceRefreshMonitor()`, then wait for the shared monitor. Because the monitor cannot recover with its retained context and has no cached topology, callers repeatedly enter `waitForTopologyUpdate`.

The immediate recovery `SQLException` is not exposed to the callers in the affected version, making the initiating database, authentication, DNS, or network failure difficult to diagnose.

### Reproduction Steps

1. Create one Aurora PostgreSQL cluster with logical databases A and B.
2. Configure separate valid IAM credential contexts for A and B.
3. Configure a credentials handler that chooses the provider using a connection property such as `credentialContextId`.
4. Connect to A first using the cluster endpoint, `wrapperPlugins=iam`, Aurora dialect detection, and the default `clusterId`. Confirm that this creates the topology monitor.
5. Connect to B using the same physical cluster and `clusterId`, but B's database and credential context.
6. Invalidate A's credentials or remove database A.
7. Expire or clear the topology cache, or otherwise force the monitoring connection to be recreated.
8. Open concurrent connections to B.
9. Observe that the shared monitor reconnects using A's retained properties while B's callers wait for topology refresh.

### Possible Solution

Separate cluster-scoped topology identity from replaceable monitoring-connection context.

One possible implementation is to allow `RdsHostListProvider.getOrCreateMonitor()` to offer its current connection context to an existing monitor. The monitor should:

- keep topology and monitor identity keyed by `clusterId`;
- accept a new connection candidate without exposing credentials outside the monitor;
- prefer the existing healthy monitoring connection;
- when in panic mode, allow a healthy current caller's context to replace the stale candidate used for recovery;
- update context atomically and wake the monitoring loop;
- avoid changing context during a successful recovery attempt;
- preserve per-monitor property overrides and sanitization;
- include concurrency tests with two contexts sharing one cluster.

Using a unique `clusterId` per database is a workaround, but duplicates monitor threads, caches, and topology queries for the same physical cluster. Setting `wrapperDialect=pg` avoids topology monitoring entirely but removes topology-aware failover and read/write splitting.

This issue is being opened as a draft for maintainer feedback before finalizing the accompanying implementation. The proposed pull request will include focused unit tests and will avoid application-specific identifiers or configuration.

#### Why this appears not to have been reported previously

The failure requires a less common combination of otherwise valid conditions:

1. Multiple connections target different logical databases or use independently managed credential contexts on the same physical Aurora cluster.
2. They share a `clusterId`, as recommended when connections share one physical topology and failover domain.
3. The context that happens to create the monitor first later becomes invalid or disappears.
4. The topology cache is empty or expires, forcing the monitor to recreate its own connection.
5. Other contexts continue creating connections concurrently.

In the common case of one database and one stable credential context per cluster, the monitor's first-caller ownership is not visible. It also remains hidden while its original monitoring connection stays healthy.

The topology monitor can be installed implicitly through Aurora dialect detection even when `iam` is the only explicitly enabled plugin. An application can therefore encounter topology-monitor behavior without intentionally configuring failover or read/write splitting.

When the defect does trigger, the external symptoms resemble generic connection-pool exhaustion or database unavailability rather than a shared-monitor identity problem. In the observed version, the monitoring reconnect `SQLException` was caught inside the recovery path while callers saw repeated topology waits. Diagnosing the ownership relationship requires correlating thread dumps or JFR stacks with the monitor cache key and constructor-captured properties.

Existing reports have consequently surfaced adjacent failure modes separately:

- different physical clusters accidentally sharing the default `clusterId`;
- monitoring connections losing IAM plugin configuration;
- monitoring connections inheriting incompatible target-driver properties;
- topology discovery retrying an unresolvable host until connection pools drain.

Those reports led to configuration guidance and narrow fixes, but none required the topology to remain shared while its database and credential connection context was replaceable. This same-cluster, independent-lifecycle distinction is why the existing reports do not fully describe this case.

Observed on 4.2.0. The monitor ownership and context-capture behavior remains present in 4.4.0 and current `main`.

### JDK version used

OpenJDK 21

Linux, containerized deployment

agesenm-ELS:
### Summary

Allows a shared Aurora topology monitor to recover with a current provider's connection context when the context that originally created the monitor can no longer connect.

Fixes #2110.

### Description

`RdsHostListProvider` shares topology monitors by `clusterId`. Before this change, the resulting `ClusterTopologyMonitorImpl` was permanently bound to the first provider's `initialHostSpec`, connection properties, and minimal services container. If that database or credential context became invalid, the shared monitor could remain in panic mode even while another provider for the same physical cluster had a valid connection context.

This change:

- lets `RdsHostListProvider` and `GlobalAuroraHostListProvider` offer their current context to an existing monitor;
- accepts replacement contexts only while the monitor has no active monitoring connection;
- applies the replacement on the monitor thread rather than a connection caller thread;
- builds an isolated minimal services container before changing monitor state;
- rebuilds monitoring properties, including host-selection removal, monitoring-prefix overrides, and timeout defaults;
- stops stale node workers and recreates the monitoring connection handler after a successful context replacement;
- discards offers that arrive during the final part of a successful recovery so they cannot be applied during a later panic episode;
- leaves healthy monitors and their current connection context unchanged.

Topology remains shared by physical cluster through the existing `clusterId`; this does not create a monitor per logical database or credential context.

The handoff is intentionally last-writer-wins when several providers offer contexts concurrently. A failed candidate leaves the monitor in panic mode, allowing a later provider interaction to offer another context. This draft requests maintainer feedback on whether a bounded multi-candidate strategy is preferable.

For Global Aurora, region topology and instance templates remain cluster-scoped and are not replaced with connection context.

### Tests

Added `ClusterTopologyMonitorRecoveryContextTest` with coverage for:

- accepting a context in panic mode and rejecting it while healthy;
- rebuilding and sanitizing monitoring properties;
- isolated minimal-container creation and failure handling;
- connection-handler recreation and node-worker state reset;
- stale offers during panic exit;
- concurrent context offers;
- provider integration while preserving shared monitor identity.

Local validation:

```text
./gradlew :aws-advanced-jdbc-wrapper:test \
  --tests software.amazon.jdbc.hostlistprovider.ClusterTopologyMonitorRecoveryContextTest \
  --tests software.amazon.jdbc.hostlistprovider.RdsHostListProviderTest \
  --tests software.amazon.jdbc.util.monitoring.MonitorServiceImplTest

34 tests passed, 0 failed

./gradlew :aws-advanced-jdbc-wrapper:checkstyleMain \
  :aws-advanced-jdbc-wrapper:checkstyleTest
```

### Additional Reviewers

None.

### DIFF
--- a/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
+++ b/wrapper/src/main/java/software/amazon/jdbc/hostlistprovider/ClusterTopologyMonitorImpl.java
@@ -125,11 +125,14 @@ public class ClusterTopologyMonitorImpl extends AbstractMonitor
   protected final long refreshRateNano;
   protected final long highRefreshRateNano;
   protected final TopologyUtils topologyUtils;
-  protected final FullServicesContainer servicesContainer;
-  protected final Properties properties;
-  protected final Properties monitoringProperties;
-  protected final HostSpec initialHostSpec;
+  protected volatile FullServicesContainer servicesContainer;
+  protected volatile Properties properties;
+  protected volatile Properties monitoringProperties;
+  protected volatile HostSpec initialHostSpec;
   protected final HostSpec instanceTemplate;
+
+  // Recovery context offered by a later provider when the monitor is in panic mode.
+  protected final AtomicReference<@Nullable RecoveryContext> pendingRecoveryContext = new AtomicReference<>(null);
 
   protected @Nullable ExecutorService nodeExecutorService = null;
   protected long highRefreshRateEndTimeNano = 0;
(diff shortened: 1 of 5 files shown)


agesenm-ELS:
Not needed with suggested configuration in #2110

PR #2117, opened 2026-09-04
fix: reset Blue/Green context when a deployment is deleted

sergiyvamz:
## Problem

When a Blue/Green Deployment is deleted and recreated while the application keeps running, the TARGET monitor stays pinned to the deleted green endpoint and never finds the new one. `correspondingNodes` also keeps mapping blue nodes to the deleted green nodes, so a later switchover would route to hosts that no longer exist. The only workarounds were renaming the new green instance to the old identifier, or restarting the application.

`connectionHostSpec` is effectively write-once per monitor: the only fallback to the blue endpoint requires it to be null, and nothing ever resets it. `resetContextWhenCompleted()` is the only path that recreates monitors, and it fires only after a switchover has actually completed.

## Changes

**Deployment removal is detected and the context is reset** (`BlueGreenStatusProvider`). New `checkDeploymentRemoved()`, keyed off the SOURCE monitor, because once green is gone the TARGET monitor cannot connect at all and keeps reporting the last phase it read. It accepts both a null phase and `NOT_CREATED`, since PostgreSQL reports a deleted deployment as an error that the monitor maps to `NOT_CREATED`. Debounced by elapsed time rather than by counting observations: `prepareStatus()` short-circuits on an unchanged interim status hash and a deleted deployment reports an identical status on every tick, so a counter would never advance; the check therefore runs on the short-circuit path as well. Gated to skip `COMPLETED`, which `resetContextWhenCompleted()` already owns, and self-terminating via `NOT_CREATED` so it cannot reset in a loop. Deletion during a switchover is covered. The body of `resetContextWhenCompleted()` was extracted into a shared `resetContext()` as a pure move.

**The failure is now visible.** A deleted deployment was previously completely silent: the only log on that path is `bgd.noEntriesInStatusTable` at FINEST, which is explicitly skipped for SOURCE. An unreachable green produced only a repeating FINEST line with no matching "Opened". Added `bgd.deploymentRemoved` (INFO), `bgd.endpointUnreachable` and `bgd.endpointReachableAgain` (WARNING / INFO, once per outage, with endpoint and duration), and `bgd.greenNotReadyForSwitchover` (WARNING). Readiness now accounts for reachability, so an unreachable green is no longer reported as ready for switchover; previously the readiness block returned early when not ready, so its `ready` field was always literally `true`. Unreachable tracking is suppressed while `useIpAddress` is set, that is during the active switchover phases, where nodes are legitimately unavailable.

**Cached topology is invalidated over the event bus.** `stopMonitor()` removes the topology monitor but leaves its cached `Topology` behind, and the monitoring cluster id is derived only from the deployment id and the role, so a replacement monitor could start from the topology of an environment that no longer exists. Added `DataInvalidationEvent`, the counterpart of `DataAccessEvent`, with immediate delivery because batching would leave the stale entry readable for up to the publisher interval. `StorageServiceImpl` now implements `EventSubscriber` and discards the item. Its `releaseResources()` intentionally does not unsubscribe, matching `MonitorServiceImpl`: the instance is a long-lived singleton that `Driver.releaseResources()` releases without replacing, so unsubscribing would permanently disable invalidation for a driver that is used again afterwards.

## New properties

- `bgDeploymentRemovedGraceMs` (default 30000): how long a deployment must be continuously absent from the status information reported by the blue database before it is treated as deleted.
- `bgEndpointUnreachableMs` (default 60000): how long a monitoring endpoint must be continuously unreachable before it is reported as unreachable and not ready for switchover.

## Testing

21 new unit tests: 13 in `BlueGreenStatusProviderTest`, 7 in a new `StorageServiceImplTest`, 1 in `BatchingEventPublisherTest`. Full unit suite passes at 1383 tests, checkstyle is clean, and the Checker Framework NullnessChecker in enforcing mode reports zero findings.

These scenarios are not reachable from the integration tests: reproducing them requires deleting a deployment and creating a second one against a still-running JVM, and the failure lives in state that survives across deployments.

## Compatibility

The public constructor of `BlueGreenInterimStatus` gains an `endpointUnreachable` parameter. Both call sites are in this repository, but it is a source-incompatible change for anyone constructing it directly.


docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
```diff
@@ -84,6 +84,8 @@ The Blue/Green Deployment Plugin supports the following configuration parameters
...
+| `bgDeploymentRemovedGraceMs`      | Integer |                              No                              | Available since version 4.5.0.<br/><br/>How long (in milliseconds) a Blue/Green Deployment must be continuously absent from the status information reported by the blue database before the driver treats it as deleted and resets its monitoring context. See [Deleting and recreating a deployment](#deleting-and-recreating-a-deployment).                                                                                                             | `30000`                  | `30000`       |
```
karenc-bq, line 87: Can we add a version column for the connection parameters

docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
```diff
@@ -201,6 +203,49 @@ timestamp                         time offset (ms)
 ---------------------------------------------------------------------------------------
 ```
 
+## Deleting and recreating a deployment
+
+> **Since version:** 4.5.0
```
karenc-bq, line 208: Can we change it to the `> [!IMPORTANT]` annotation

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
```diff
@@ -154,6 +169,17 @@ public BlueGreenStatusMonitor(
...
+    this.endpointUnreachableThresholdNano = TimeUnit.MILLISECONDS.toNanos(
+        BlueGreenStatusProvider.BG_ENDPOINT_UNREACHABLE_MS.getLong(props));
+  }
+
+  /**
+   * For testing purposes.
+   *
+   * @return current time in nanoseconds
+   */
+  protected long getNanoTime() {
```
karenc-bq, line 181: I think we have this method in several places can we centralize this method to reduce code duplication

karenc-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/bluegreen/BlueGreenStatusMonitor.java
sergiyvamz: That's a method to override to simplify unit testing. If you need to reduce code duplication (that's just a single line) then we need to introduce a time provider and pass it to every class that needs it. It changes constructor signature that I'd try to avoid.

docs/using-the-jdbc-driver/using-plugins/UsingTheBlueGreenPlugin.md
sergiyvamz: I'll address all entries like this in a separate PR.

Issue #1324, opened 2025-03-15
AWS JDBC Wrapper: Read-Write Splitting Plugin Fails to Fallback to Writer When Reader is Unavailable


seiichi1101:
### Describe the bug

I am building a web application using **ALB + ECS + Aurora**.  
I am using **Aurora MySQL - 8.0.mysql_aurora.3.04.3**, with two instances (one writer and one reader) spread across different availability zones.  

With **Read/Write Splitting Plugin**, I successfully distributed the load between the writer and reader instances. 
However, when I use **AWS Fault Injection Simulator (FIS)** with the action `aws:network:disrupt-connectivityaws` to block network access to the reader's subnet, API requests from the client get stuck.

The logs continuously output `WARN` messages indicating connection failures:

```txt
{
    "timestamp": "2025-03-14T19:18:28.634671716Z",
    "message": "HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@74fbcc42 (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.",
    "logger": "com.zaxxer.hikari.pool.PoolBase",
    "level": "WARN",
    "level_value": 30000
},
{
    "timestamp": "2025-03-14T19:18:33.591088926Z",
    "message": "HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@39f81981 (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.",
    "logger": "com.zaxxer.hikari.pool.PoolBase",
    "level": "WARN",
    "level_value": 30000
},
{
    "timestamp": "2025-03-14T19:18:33.640687431Z",
    "message": "HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@7730174e (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.",
    "logger": "com.zaxxer.hikari.pool.PoolBase",
    "level": "WARN",
    "level_value": 30000
},
{
    "timestamp": "2025-03-14T19:18:38.548419767Z",
    "message": "Failed to connect to reader host: '<reader-instance-identifier>.abcdefghijklm.ap-northeast-1.rds.amazonaws.com:3306/'",
    "logger": "software.amazon.jdbc.plugin.readwritesplitting.ReadWriteSplittingPlugin",
    "level": "WARN",
    "level_value": 30000,
}
```

[...]

PR #2126, opened 2026-09-09
fix: initialConnection plugin no longer retries internal connections it does not substitute or verify

seiichi1101:
### Summary

Follow-up to #1324 / #1966: `AuroraInitialConnectionStrategyPlugin` no longer wraps internal connections it does not act on in its `openConnectionRetryTimeoutMs` retry loop.

### Description

**Problem**

`AuroraInitialConnectionStrategyPlugin.connect()` wraps *every* connect call in a retry loop bounded by `openConnectionRetryTimeoutMs` (default 30,000 ms). This includes internal connections other plugins open through `pluginService.connect(host, props, this)` with `isInitialConnection=false`.

For such a connection to an instance endpoint the plugin has nothing to do:

- `getRoleToVerify()` returns `null` as soon as `!isInitialConnection`, so no role is verified.
- `getInstanceSubstitutionStrategy()` returns `DO_NOT_SUBSTITUTE` for instance endpoints (and non-RDS hosts), so no endpoint is substituted.

Each loop iteration therefore calls `connectFunc.call()` against the same host with the same properties. A network failure hits the `isNetworkException(...) → continue` branch with no delay (`openConnectionRetryIntervalMs` is only applied on the role-verification path), and the loop keeps going until `openConnectionRetryTimeoutMs` is exhausted, after which `AuroraInitialConnectionStrategyPlugin.timeout` is thrown.

**Impact**

The loop amplifies the retry policy of the calling plugin.

Concrete case (#1324 follow-up), plugins `initialConnection,auroraConnectionTracker,readWriteSplitting,failover2,efm2` and a 2-node cluster whose only reader becomes unreachable (e.g. AWS FIS `aws:network:disrupt-connectivity` on the reader subnet):

1. `setReadOnly(true)` makes the read/write splitting plugin open a reader connection (`TopologyReaderResolver.openNewReaderConnection()` on `main`, `ReadWriteSplittingPlugin.openNewReaderConnection()` in 4.1.0).
2. Since #1966 the plugin tries each reader at most once and falls back to the writer.
3. That single attempt goes through `AuroraInitialConnectionStrategyPlugin`, which re-attempts the same unreachable reader until `openConnectionRetryTimeoutMs` runs out.

As a result each read-only transaction spends the full `openConnectionRetryTimeoutMs` (up to 30 s by default) before falling back, instead of one connect timeout (~1 s in my configuration). Since `initialConnection` is part of the default plugin set, adding `readWriteSplitting` to the defaults is enough to hit this.

The same `pluginService.connect(host, props, this)` pattern is used by failover2 (reader candidate and original-writer attempts), limitless, blue/green routing and `RetryUtil`, so the amplification is not specific to read/write splitting.

The error reported on this path is also misleading: it says the initial connection strategy timed out and points at `verifyOpenedConnectionType`, although neither an initial connection nor role verification was involved.

The existing regression test for #1324 (`ReadWriteSplittingTests.test_setReadOnlyTrue_oneReaderDown_fallsBackToWriter`) runs with `readWriteSplitting` alone and has no timing assertion, so it does not detect this.

Observed on <version>: <log excerpt: AuroraInitialConnectionStrategyPlugin.timeout exception and the time spent in setReadOnly(true), default 30 s vs. openConnectionRetryTimeoutMs=3000>

**Change**

Before entering the retry loop, pass the call straight through when the plugin has nothing to do:

```java
if (!isInitialConnection
    && roleToVerify == null
    && substitutionStrategy == InstanceSubstitutionStrategy.DO_NOT_SUBSTITUTE) {
  final Connection conn = connectFunc.call();
  this.pluginService.setRoutedHostSpec(originalConnectHost);
  return conn;
}
```

Scope notes:

- Restricted to `!isInitialConnection`, so the application's initial connection (plain URLs and custom domains included) keeps the existing retry behavior.
- Internal connections that *do* get substituted (cluster endpoints resolving to `SUBSTITUTE_WITH_WRITER` / `SUBSTITUTE_WITH_READER`) keep the loop, since host re-selection between attempts is meaningful there.
- `roleToVerify == null` is checked explicitly even though it follows from `!isInitialConnection` today, so the pass-through does not depend on that invariant.
- `setRoutedHostSpec(originalConnectHost)` is kept, matching what the success path did for this case, so downstream plugins reading `getRoutedHostSpec()` (`auroraConnectionTracker`, `efm`) see no change.
- Loop body, exception handling and messages are unchanged.

**Tests**

- New `AuroraInitialConnectionStrategyPluginTest`:
  - non-initial connection to an instance endpoint calls `connectFunc` exactly once and returns its connection;
  - a network `SQLException` on that path propagates immediately, unwrapped, with a single attempt (fails without the fix: the timeout exception is thrown instead);
  - an initial connection to an instance endpoint still retries until `openConnectionRetryTimeoutMs` (regression guard);
  - a non-initial connection to a cluster endpoint still goes through substitution and connects to the selected writer instance.
- `ReadWriteSplittingTests.test_setReadOnlyTrue_oneReaderDown_fallsBackToWriter_withInitialConnectionPlugin`: same scenario as the #1966 test with `initialConnection` in the chain, asserting that `setReadOnly(true)` falls back well within `openConnectionRetryTimeoutMs`. Not run locally (no AWS credentials available on my side); relying on CI for this one.
- Existing unit tests (1364), checkstyle and spotbugs pass locally.

**Docs**

Added a note to `UsingTheAuroraInitialConnectionStrategyPlugin.md` clarifying which connections `openConnectionRetryTimeoutMs` applies to.

**Related**

- #1324 (read/write splitting does not fall back when the reader is unreachable)
- #1966 (read/write splitting side of the fix; this PR addresses the remaining amplification on the `initialConnection` side)

### Additional Reviewers

@sergiyvamz (handled #1324 / #1966)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sergiyvamz:
Hi @seiichi1101, thanks for the detailed write-up - the analysis of the retry amplification through `AuroraInitialConnectionStrategyPlugin.connect()` is clear and easy to follow.

This PR is currently marked as a draft, so we're treating it as work in progress and haven't scheduled it for review, integration testing, or merge yet. Draft PRs don't get our test pipelines run against them.

Whenever you'd like us to pick it up, two things:

1. Please rebase onto the latest `main`. The branch is 15 commits behind and currently has merge conflicts, so it can't be built or merged as-is.
2. Mark the PR as ready for review (remove the draft flag).

Once both are done we'll take a proper look and run it through our integration tests - including the `ReadWriteSplittingTests` case you added, since you mentioned you weren't able to run that one locally.

PR #2130, opened 2026-09-14
fix: reuse Secrets Manager client across connect retries

sergiyvamz:
## Summary

The `awsSecretsManager` plugin built and closed a brand-new `SecretsManagerClient` for every
credential fetch. Building an SDK client is not free: it stands up an HTTP client, a connection
pool, and background threads. With the connect retry budget added in #2097 re-fetching the
credentials before each retry, a single `connect()` during a rotation window could build and tear
down that client several times over the backoff sequence, once per attempt.

This addresses the review feedback on #2097
(https://github.com/aws/aws-advanced-jdbc-wrapper/pull/2097#discussion_r3798219717): reuse one
client across all the fetches of a single `connect()` instead of one per attempt.

## The change

`connectInternal` now owns a single client for the duration of one connect attempt. The client is
built lazily on the first fetch, reused by every retry, and closed exactly once when the attempt
finishes. Nothing is built when the cache is warm and no fetch happens, so the common path is
unchanged.

The client is held in a `ThreadLocal` bound to the connecting thread rather than a plain instance
field. This keeps the reuse scoped to the synchronous connect flow and, importantly, leaves the
`awsSecretsManager2` (SWR) async refresh path untouched: that refresh runs on a separate executor
thread, does not see the holder, and keeps building and closing its own client. Scoping it this
way avoids a use-after-close race where the connect thread's cleanup could close a client an
in-flight background refresh is still using.

Closing the reused client is wrapped so a failure to close can never turn a successful connection
into a failure; it is logged at `FINEST`.

## Credentials

Reusing the client is safe when AWS credentials change. The client holds an
`AwsCredentialsProvider`, not resolved credentials, and the SDK re-resolves it on every request, so
rotating STS/role/instance credentials are still picked up. The reuse is scoped to a single
`connect()` (a few seconds), and the async path that lives for the whole connection continues to
build its own client, so no provider is pinned across connections.

## Testing

- Added `testConnectRetryBudgetReusesSecretsManagerClient`: a retry-budget connect that performs
  two forced re-fetches asserts the client function is invoked once and the client is closed once.
- `./gradlew :aws-advanced-jdbc-wrapper:compileJava :aws-advanced-jdbc-wrapper:compileJava17Java` - both build.
- `./gradlew :aws-advanced-jdbc-wrapper:test` for the Secrets Manager plugin, plugin2, and rotation-window suites - 53 tests, 0 failures.
- `./gradlew :aws-advanced-jdbc-wrapper:compileJava -PenableCheckerFramework` - clean.
- `./gradlew :aws-advanced-jdbc-wrapper:checkstyleMain :aws-advanced-jdbc-wrapper:checkstyleJava17 :aws-advanced-jdbc-wrapper:checkstyleTest` - pass.

The change is mirrored into the `src/main/java17` multi-release variant so it takes effect on Java
17+ runtimes as well as the base variant.


wrapper/src/main/resources/aws_advanced_jdbc_wrapper_messages.properties
```diff
@@ -65,6 +65,7 @@ AwsSecretsManagerConnectionPlugin.invalidSecretFormat=Unable to fetch database c
 AwsSecretsManagerConnectionPlugin.jacksonDatabindNotInClasspath=Required dependency 'Jackson Databind' is not on the classpath.
 AwsSecretsManagerConnectionPlugin.failedToFetchDbCredentials=Was not able to either fetch or read the database credentials from AWS Secrets Manager. Ensure the correct secretId and region properties have been provided.
 AwsSecretsManagerConnectionPlugin.missingRequiredConfigParameter=Configuration parameter ''{0}'' is required.
+AwsSecretsManagerConnectionPlugin.reusableClientCloseFailed=Failed to close the AWS Secrets Manager client after opening a connection.
```
sophia-bq, line 68:
```suggestion
AwsSecretsManagerConnectionPlugin.reusableClientCloseFailed=Failed to close the AWS Secrets Manager Client after opening a connection.
```

sophia-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
```diff
@@ -548,16 +575,31 @@ protected Secret fetchAndStoreSecret(final HostSpec hostSpec) throws SQLExceptio
    */
   Secret fetchLatestCredentials(final HostSpec hostSpec)
       throws SecretsManagerException, JsonProcessingException, SQLException {
-    final SecretsManagerClient client = secretsManagerClientFunc.apply(
-        hostSpec,
-        Region.of(this.secretKey.getValue2()));
+    // When invoked on a connecting thread, reuse (and lazily build) the client owned by
+    // connectInternal so retries share one client. Otherwise (e.g. the awsSecretsManager2 async
+    // refresh) build a client scoped to this single fetch and close it when done.
+    final ConnectClientHolder clientHolder = this.connectClientHolder.get();
```
aaron-congo, line 581: instead of trying to distinguish between the aws secrets manager 1 vs 2 in this method, would we want to consider overriding this method in secrets manager 2 so each plugin has its own implementation of it?

aaron-congo (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/AwsSecretsManagerConnectionPlugin.java
sergiyvamz: AWS secrets manager 2 plugin uses a different approach with background threads and directly applying the same approach from AWS secrets manager 1 (as above) doesn't work. I can make a backlog ticket to improve AWS secret manager 2 as a separate PR.

PR #2131, opened 2026-09-14
docs: add Available Since Version column to parameter tables

sergiyvamz:
## Summary

Adds an **Available Since Version** column to the configuration/connection parameter reference tables across the driver documentation, so every documented parameter shows the release it first became available in. This follows up on review feedback asking for a version column on the connection parameters (originally on the Blue/Green plugin table) and extends the same column to all parameter reference tables for consistency.

## What changed

Added an `Available Since Version` column (placed as the second column, right after the parameter name) to 33 parameter reference tables across 28 doc files:

- Core: `UsingTheJdbcDriver.md` (AWS wrapper parameters, system properties, and Connection Plugin Manager parameters), `Telemetry.md`, `HostAvailabilityStrategy.md`, `ExceptionContext.md`
- Plugin docs: Blue/Green, Failover, Failover2, GDB Failover, Host Monitoring (parameters + EFM system properties), IAM, Federated Auth, Okta, Secrets Manager (v1 and v2), Read/Write Splitting (plus Auto, Simple, and the GDB/Auto/Simple variants), Custom Endpoint, Driver Metadata, KMS Encryption, Limitless, Remote Query Cache, Aurora Initial Connection Strategy, Global Aurora Accessible Regions, and Monitoring Connection Priority

The header wording `Available Since Version` matches the column already present in `HostSelectionStrategies.md`.

## How versions were determined

Each parameter's version was derived from git history: the first commit that introduced the parameter's config-key literal under `wrapper/src/main/java`, mapped to the first release tag that contains that commit. Parameters not yet in any released tag are marked `4.5.0` (the next release). Deprecated parameters show their original introduction version.

## Scope notes

Tables that are not parameter-definition lists were intentionally left unchanged: `HostSelectionStrategies.md` (already has the column), the configuration-assistant guide, the failover configuration example profiles, DataSource setup-property tables, usage-example tables, compatibility matrices, and the development/integration-test property tables.


docs/using-the-jdbc-driver/Telemetry.md
```diff
@@ -107,7 +107,7 @@ When a trace is hierarchically linked to a parent trace, we say that this trace
 
 Applications that interact with the AWS Advanced JDBC Wrapper may or may not have already opened telemetry traces on their own. In this case, it is up to the user to determine how they want to mix both application and driver traces.
 
-This can be done using the AWS Advanced JDBC Wrapper's `telemetrySubmitTopLevel` property. This property allows users to choose to submit the driver traces always as top level traces or to submit them nested to the application traces.
+This can be done using the AWS Advanced JDBC Wrapper's `telemetrySubmitToplevel` property. This property allows users to choose to submit the driver traces always as top level traces or to submit them nested to the application traces.
```
aaron-congo, line 110: looks like 'Level' became 'level' here, is that intentional?

aaron-congo (approved):

sophia-bq (approved):

docs/using-the-jdbc-driver/Telemetry.md
sergiyvamz: Yes, it's a fix. Parameter is defined as "telemetrySubmitToplevel".

PR #2136, opened 2026-09-15
fix: correct defects found by a full architecture review

sergiyvamz:
## Summary

A full architecture read of the wrapper produced fifteen candidate defects. Each was verified against the source before anything was changed. Ten held and are fixed here. Two were refuted and are deliberately left alone, with the reasoning recorded below so the same ground is not covered again. Three were superseded by the static findings they were meant to confirm.

The changes are independent of one another, so this can be split if a reviewer prefers.

## Behaviour fixes

**IAM token cache was never purged.** `IamAuthConnectionPlugin.clearCache()` cleared `Topology.class`. Tokens are registered and stored under `TokenInfo.class`, so the method purged the topology cache and left every cached token in place.

**Okta properties leaked to the target driver.** `OktaAuthPlugin` never called `PropertyDefinition.registerPluginProperties`, and neither does `BaseSamlAuthPlugin`. Registration is what drives stripping (`PropertyDefinition.removeAll` / `removeAllExceptCredentials`, called from the target driver dialects), so Okta's wrapper-only properties were handed to the target driver as unrecognised connection properties. Most of the names happen to be registered by `FederatedAuthPlugin` or `IamAuthConnectionPlugin`, but only if one of those classes is loaded, which does not happen when `okta` is the only auth plugin configured. `appId` is declared by `OktaAuthPlugin` alone and was registered by nothing anywhere. pgjdbc tolerates unknown properties; a stricter driver turns this from a leak into a connection failure.

**Okta had no plugin-weight entry.** Every other auth plugin is pinned (`iam` 1800, `awsSecretsManager` 1900, `awsSecretsManager2` 1910, `federatedAuth` 2000). `okta` fell into the `WEIGHT_RELATIVE_TO_PRIOR_PLUGIN` branch by omission, so its chain position depended on what the user listed before it. Now 2010.

**KMS annotations disabled the encryption they were asking for.** `EncryptionAnnotationParser` captures one group of the form `table.column`, and that qualified string is bound to the *column* slot in `handleSetParameter`. `MetadataManager` composes its key as `tableName + "." + columnName`, producing `users.users.ssn`, so the lookup always missed and `isColumnEncrypted` returned false, storing the value in plaintext.

The fix splits a qualified identifier on the last dot before the lookup, which also honours the table the annotation names (relevant for multi-table statements) and keeps `MetadataManager`'s contract intact. Splitting on the last dot leaves a schema-qualified `schema.table.column` correct.

Worth flagging for reviewers: an annotation was not merely inert, it was regressive. `SqlParserConnectionPlugin` builds the parameter mapping as `putAll(positional)` then `putAll(annotations)`, and the positional mapping holds *bare* column names. Annotating a parameter that already encrypted via the metadata table replaced a working mapping with an unresolvable one.

**The JDK 24 dialect variant silently degraded five methods.** A multi-release JAR replaces a class wholesale, so on JDK 24 the copy under `src/main/java24` is loaded *instead of* the base one. Both extend `GenericTargetDriverDialect`, so a method omitted from the JDK 24 variant does not fail to compile and does not disappear: it falls back to the generic implementation. Missing were `registerDataType`, `setEncryptedParameter`, `getEncryptedBytes`, `updateInternalState` and `removeHostSelectionProperties`. The generic fallbacks are wrong for PostgreSQL:

| Method | Generic fallback | Consequence on JDK 24 |
|---|---|---|
| `updateInternalState` | no-op | `currentSchema` and `readOnly` are not transferred when a connection is swapped, so the application silently gets the new connection's defaults after a failover |
| `registerDataType` | empty method | PG data type registration does nothing |
| `setEncryptedParameter` / `getEncryptedBytes` | `ps.setBytes` / `rs.getBytes` | bypasses `PgEncryptedDataHelper` |
| `removeHostSelectionProperties` | interface default, returns empty set | `targetServerType` is not stripped from monitoring connection properties, so EFM, topology, Blue/Green and Limitless monitors inherit a setting that makes pgjdbc refuse a reader |

`getSQLQueryString` had also lost its `@Nullable` return annotation.

This is the second time the two variants have drifted (#2129 patched `prepareTargetDataSource`), so the fix is not only to port the methods. `PgTargetDriverDialectVariantTest` parses both source files and fails if the JDK 24 variant does not declare every public or protected method the base variant declares. Overriding with different behaviour is still fine and expected; omitting a method is not. The test was confirmed to fail when a method is removed, not merely to pass as written.

The root cause of the drift is worth noting: the `java24` source set had no pgjdbc on its compile classpath, so the PG-specific methods could not have been written there. `wrapper/build.gradle.kts` now adds pgjdbc as `compileOnly` for that source set, matching how the main source set declares it.

**dataCache was unbounded in both size and staleness.** Results were held in a process-wide `ConcurrentHashMap` keyed on the raw SQL text with no TTL, no eviction and no size limit. Two independent failure modes: heap growth unbounded in the number of distinct SQL strings the process ever sees, which is unbounded for any application that inlines literals instead of binding parameters; and unbounded staleness, since nothing invalidates a cached result on write.

Adds `dataCacheTtlMs` (default 300000) and `dataCacheMaxSize` (default 1000). Each entry carries its own expiry deadline computed from the TTL of the connection that stored it, so one static cache can still be shared by connections configured differently. When the cache is full, expired entries are purged first; if no room can be made the result is returned to the caller without being cached, which costs a cache miss and nothing else.

The plugin had no user-facing documentation at all, unlike `remoteQueryCache`. `UsingTheDataCachePlugin.md` is new and is explicit that this is a read-through cache with no invalidation.

**secretsManagerRegion declared a default that could never apply.** The property declared `us-east-1`, but the region is resolved with `regionUtils.getRegion(props, REGION_PROPERTY.name)`, which does a raw `props.getProperty` lookup; `AwsWrapperProperty` defaults are only applied by its own accessors. The declared default is removed rather than made to apply, because a null result is precisely what lets resolution fall through to the region embedded in the secret ARN. Making the default apply would have turned the ARN fallback into dead code and sent every unconfigured user to one fixed region regardless of where their secret lives.

## Documentation fixes

**The KMS annotation is not stripped before dispatch.** `UsingTheKmsEncryptionPlugin.md` stated that it is, and `EncryptionAnnotationParser.stripAnnotations`'s javadoc says the same. The stripped text is computed and published to the call context under `sql.cleanSql`, but no production code reads it, and no stage of the pipeline substitutes it for the original statement. The annotated SQL is what the target driver receives. It is a valid SQL comment so it does not affect execution, but it names the table and column being encrypted and therefore appears in server statement logs, `pg_stat_statements`, slow-query logs and query-capture tooling.

This is documented rather than fixed in code because a real fix is a design change, not a small one. The terminal callable in `ConnectionWrapper` closes over the wrapper method's `sql` local, not over the `jdbcMethodArgs` array, and there is no reflection in the dispatch path. Writing to the args array would be visible to later plugins and invisible to the driver. Rewriting the statement text needs a per-call override channel that the terminal lambda consults, threaded into the repreparer too, since `withRebind`'s `c -> c.prepareStatement(sql)` also captures the original string and would resurrect the annotated SQL on a read/write-splitting reroute. Worth its own change.

**The writer is not at position 0.** `RdsHostListProvider.getTopology()`'s javadoc claimed it is. `TopologyUtils.verifyWriter` builds the list readers-first and appends the writer last. Nothing in `hostlistprovider` reads index 0 of the topology, so the javadoc is corrected rather than the list reordered; reordering would risk callers outside that package for no gain.

**weightedRandom lacked the polarity warning highestWeight already carries.** `HostSpec.weight` has no single meaning: the field's own convention is that lower is healthier, `AuroraTopologyUtils` follows it by storing `lag*100 + cpu`, and `LimitlessQueryHelper` deliberately does the opposite by storing a fitness score in 1..10 where higher is better. `HostSelectionStrategies.md` already documents this for `highestWeight` and deprecates it in favour of `lowestLoad`; the `weightedRandom` row did not, even though it uses `HostSpec.getWeight()` as probability mass and therefore sends proportionally more reads to the *most* loaded reader on an Aurora topology.

No formula or selector behaviour is changed. Normalising the polarity was considered and rejected: `HostSpec.weight`'s documented contract and Limitless's use of it point in opposite directions, so changing either is a behaviour change for anyone already using these strategies. The polarity is now stated at every point where a reader would need it: `HostSpec.getWeight()`, both selectors, and the Aurora weight calculation.

**efm and efm2 are not interchangeable to the millisecond.** The docs called them functionally equal and a drop-in replacement. Both measure how long a node has been failing since the first unacknowledged probe, but v1 aborts at `failureDetectionInterval x failureDetectionCount` and v2 at `failureDetectionInterval x (failureDetectionCount - 1)`. On defaults that is 15,000 ms versus 10,000 ms after the grace period.

Since the timer starts on the *first* failed probe, `efm2` is the one that aborts after exactly `failureDetectionCount` probes as the same document describes; `efm` waits one extra interval. The note is written that way round rather than presenting v2 as over-aggressive.

**loadBalancingIncludeWriter has no effect on the srw axis.** It is consumed only by `PerQueryBalancedReaderPolicy`, reachable only through `TopologyReaderResolver`. The srw axis uses `EndpointReaderResolver`, which never sees it and has no host list to widen. `UsingTheAutoSimpleReadWriteSplittingPlugin.md` listed it among the accepted parameters.

## Refuted while verifying, deliberately unchanged

**A data-modifying CTE does not route to a reader.** The concern was that `WITH inserted AS (INSERT ... RETURNING n) SELECT n FROM inserted` has a `SELECT` at top level and that nothing inspects `withItems` — which is true, there is no such reference anywhere in main. But JSQLParser 4.9 cannot parse a data-modifying CTE at all. `JSQLParserAnalyzer.analyze` catches the `ParseException` and falls back to `detectQueryType`, which prefix-matches the trimmed text; `WITH` matches nothing and yields `UNKNOWN`, and `SqlRoutingSignal.roleFromSqlContext` routes anything that is not a `SELECT` to the writer. Confirmed with a throwaway test through the real analyzer for the `INSERT`, `UPDATE` and `DELETE` variants. The conservative fallback exists, just upstream of `withItems`.

**queryLevelLoadBalancing is effective on the srw axis.** `EndpointReaderResolver.isPerQuery()` returns it and `UnifiedReadWriteSplittingPlugin` does rotate reader-to-reader per query by reconnecting to `srwReadEndpoint`. In the documented configuration that is an Aurora reader endpoint, which resolves to a different reader per lookup, so reads really do distribute. Only `loadBalancingIncludeWriter` is inert. The documentation now explains what rotation means on this axis rather than claiming the parameter does not work.

## Behaviour change to be aware of

`dataCache` entries previously never expired. The 5-minute default TTL is an intentional change and the point of the fix, but it will turn some current cache hits into misses. `dataCacheTtlMs=0` restores the old semantics, and `dataCacheMaxSize=0` removes the size limit; the new documentation notes that setting both is not recommended.

## Testing

- `checkstyleMain checkstyleTest spotbugsMain spotbugsTest` — clean.
- `compileJava24Java` — compiles, including the ported methods and the new pgjdbc `compileOnly` dependency.
- Unit test suite, roughly 1090 tests across every package, no failures. Run package-by-package because a leaked monitor thread in the shared test JVM keeps the whole-suite invocation from terminating on this machine; that is pre-existing and unrelated to these changes.
- New tests: three in `KmsEncryptionConnectionPluginTest` covering the annotation-qualified, schema-qualified and bare-column paths, including an assertion that the double-prefixed key is never composed; seven in `DataLocalCacheConnectionPluginTest` covering TTL expiry, TTL opt-out, the size limit, refreshing an existing key, and purging expired entries to make room; `PgTargetDriverDialectVariantTest` for the multi-release drift guard.

Not verified: the JDK 24 behaviour is established structurally, not by running on a JDK 24 runtime. The KMS annotation path is verified by unit test rather than against a live KMS key and database, so the end-to-end claim that a value now lands encrypted rests on the mocked `MetadataManager` receiving the correct key.


docs/using-the-jdbc-driver/using-plugins/UsingTheDataCachePlugin.md
```diff
@@ -0,0 +1,68 @@
+# Data Local Cache Plugin
```
sophia-bq, line 1: I think this needs to be linked in Documentation alongside the lists of all the other plugin docs

docs/using-the-jdbc-driver/using-plugins/UsingTheHostMonitoringPlugin.md
```diff
@@ -95,7 +98,10 @@ See [Monitoring](../Telemetry.md#list-of-metrics) for the metrics submitted by o
 The plugin is available since version 2.3.2.
 
 ## Overview
-Host Monitoring Plugin v2, also known as `efm2`, is an alternative implementation of enhanced failure monitoring and it is functionally equal to the Host Monitoring Plugin described above. Both plugins share the same set of [configuration parameters](#enhanced-failure-monitoring-parameters). The `efm2` plugin is designed to be a drop-in replacement for the `efm` plugin.
+Host Monitoring Plugin v2, also known as `efm2`, is an alternative implementation of enhanced failure monitoring and it is functionally equivalent to the Host Monitoring Plugin described above. Both plugins share the same set of [configuration parameters](#enhanced-failure-monitoring-parameters). The `efm2` plugin is designed to be a drop-in replacement for the `efm` plugin.
+
+> [!NOTE]\
+> "Functionally equivalent" does not mean identical timing. `efm2` aborts a connection to an unresponsive node one `failureDetectionInterval` earlier than `efm` does — see the [note on the detection deadline](#enhanced-failure-monitoring) above. `efm2` is the plugin whose deadline matches the documented `failureDetectionCount` semantics.
```
sophia-bq: Nit: this link will not directly go to the note but rather the title far far above - if you want to have a separate subtitle for the note you could and then the link would be helpful

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/DataLocalCacheConnectionPlugin.java
```diff
@@ -76,6 +96,13 @@ public DataLocalCacheConnectionPlugin(final PluginService pluginService, final P
     this.telemetryFactory = pluginService.getTelemetryFactory();
     this.dataCacheTriggerCondition = DATA_CACHE_TRIGGER_CONDITION.getString(props);
 
+    final long ttlMs = DATA_CACHE_TTL_MS.getLong(props);
+    this.ttlNanos = ttlMs <= 0 ? CacheEntry.NO_EXPIRY : TimeUnit.MILLISECONDS.toNanos(ttlMs);
```
sophia-bq: An unlikely case but if a user sets a negative value it will never expire - the wrapper tends to throw an error when other props are set to a negative value. Okay with either behaviour - just wanted to raise it as an interesting case

sophia-bq (approved):

docs/using-the-jdbc-driver/using-plugins/UsingTheAutoSimpleReadWriteSplittingPlugin.md
```diff
@@ -48,7 +48,12 @@ This plugin accepts the same endpoint parameters as the [Simple Read/Write Split
...
 
-It also accepts the family-wide `queryLevelLoadBalancing`, `loadBalancingIncludeWriter`, and `allowStatementRecreationOnConnectionSwitch` parameters — see [Query-level load balancing](./UsingTheReadWriteSplittingPlugin.md#query-level-load-balancing). Because each read statement is a routing point, query-level load balancing rotates reads per query.
+It also accepts the family-wide `queryLevelLoadBalancing` and `allowStatementRecreationOnConnectionSwitch` parameters — see [Query-level load balancing](./UsingTheReadWriteSplittingPlugin.md#query-level-load-balancing). Because each read statement is a routing point, query-level load balancing rotates reads per query.
+
+On this axis, rotation means opening a fresh connection to `srwReadEndpoint` for each read, since that endpoint is the only reader target the plugin knows about. Whether consecutive reads actually land on different instances therefore depends on what is behind that endpoint: an Aurora reader endpoint (`<cluster>.cluster-ro-<XYZ>.<region>.rds.amazonaws.com`) resolves to a different reader per lookup and does distribute, while a single-instance endpoint sends every read to the same host.
```
aaron-congo: note: I think we've observed that the aurora reader endpoint may send multiple connection in a row to the same reader instance, so it doesn't necessarily resolve to a different reader per lookup. It usually eventually starts targeting other reader instances too. So it distributes, but at a coarser granularity than our reader host selectors

docs/JDBC-WRAPPER-CONFIGURATION-ASSISTANT.md
```diff
@@ -1205,7 +1205,11 @@ Caches SQL `ResultSet`s in-process for queries matching a regex.
 
 | Name | Default | Description |
 |---|---|---|
-| `dataCacheTriggerCondition` | (none) | Regex; matching queries are cached. |
+| `dataCacheTriggerCondition` | (none) | Regex matched against the full SQL text; matching queries are cached. |
+| `dataCacheTtlMs` | `300000` | How long a cached result stays valid. Nothing invalidates a cached result on write, so this bounds staleness. `0` disables expiration. |
+| `dataCacheMaxSize` | `1000` | Maximum number of distinct SQL statements held in the cache. `0` means no limit. |
+
+> This is a read-through cache with **no invalidation**. Do not recommend it for data an application reads back after writing. For a shared, per-query opt-in cache use `remoteQueryCache` instead. See [Data Local Cache Plugin](./using-the-jdbc-driver/using-plugins/UsingTheDataCachePlugin.md).
```
aaron-congo, line 1212: Should we repeat the warning from the data cache docs here or do you think it is not necessary?
```
> [!WARNING]\
> Setting `dataCacheTtlMs=0` and `dataCacheMaxSize=0` together leaves the cache unbounded in both staleness and size. That was the behaviour before version 4.4.0 and is not recommended.
```

wrapper/src/main/java24/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
```diff
@@ -54,6 +60,15 @@ public class PgTargetDriverDialect extends GenericTargetDriverDialect {
       CP_DS_CLASS_NAME,
       XA_DS_CLASS_NAME));
 
+  /**
+   * Properties that make the PostgreSQL driver reject a node whose role does not match. {@code
+   * targetServerType} makes the driver run {@code SHOW transaction_read_only} after authenticating and
+   * fail the connection when the server is read-only, so a monitoring connection aimed at a reader or
```
aaron-congo: does this property always fail when the server is read-only, or does it depend on what you said the property value to? eg can you set it `targetServerType=reader`?

wrapper/src/test/java/software/amazon/jdbc/plugin/encryption/KmsEncryptionConnectionPluginTest.java
```diff
@@ -161,6 +161,93 @@ void test_setString_encryptsForEncryptedColumn() throws Exception {
...
 
+  /**
+   * An /*@encrypt:table.column*&#47; annotation puts a qualified identifier into the parameter
+   * mapping, overwriting the bare column name the positional mapping would have supplied. Since
+   * MetadataManager composes its lookup key as {@code tableName + "." + columnName}, the qualified
+   * value has to be split before the lookup, or the key becomes {@code users.users.ssn} and
+   * encryption is silently skipped for the annotated parameter.
+   */
+  @Test
+  void test_setString_encryptsForAnnotationQualifiedColumn() throws Exception {
```
aaron-congo, line 172: should we add an integration test so ensure this is working properly now?

aaron-congo (approved):

PR #2137, opened 2026-09-15
fix: verify read/write roles for connection-string host lists

sergiyvamz:
## Summary

Read/write splitting can serve a deployment that has no topology to query - an RDS instance with its read replicas, for example - by naming the hosts in the connection string and setting `singleWriterConnectionString=true`. The mechanics already work: readers are selected, `setReadOnly` switches, and IAM authentication mints a token per host. What was missing was the safety net around it. Three problems made that configuration untrustworthy, and the plugin documentation declared it unsupported.

Related to #1589, which asked for MySQL Connector/J's `loadbalance://` protocol to work with the wrapper's IAM authentication. Delegating multi-host handling to the target driver is a separate and much larger change; this PR makes the wrapper's own equivalent trustworthy for the same topology.

## Roles were never verified on a connection-string host list

`VerifyRoleOnConnect` returned early whenever the host list provider was static:

```java
if (!isInitialConnection
    || hostListProviderService == null
    || hostListProviderService.isStaticHostListProvider()) {
  return currentConnection;
}
```

So `verifyInitialConnectionRole`, which is enabled by default, did nothing for exactly the configuration that needs it most. Roles came from the ordering of the hosts alone (`i > 0 ? READER : WRITER`) and were trusted verbatim. If the writer was not listed first, or a replica had since been promoted, writes were routed to a read-only host and failed at the database.

The check now runs for a static list too. The measured role is pushed into the host list itself through a new `StaticHostListProvider.updateHostRole`, not just into the initial host spec: reader and writer selection both read `pluginService.getHosts()`, so correcting only the initial host spec would leave the stale role in effect. `HostSpec` holds its role immutably, so the entry is replaced in place; `ArrayList.set` does not change the list structure, so the unmodifiable views already handed out stay safe to iterate and report the new role.

A mismatch is logged at `WARNING`. The corrected role applies to the connection that discovered it, but the connection string is wrong or out of date and a static list is never refreshed, so the user has to fix it.

A role that cannot be measured at all no longer fails the connection on a static list - the declared role is kept and a warning is logged. Without this, a database that cannot answer a role query (`UnknownDialect.getHostRole` throws) would have had its connections refused where they previously succeeded. A topology-backed provider still fails, unchanged.

## A failed read switch was invisible

When no host in the list carries the reader role - the outcome of omitting `singleWriterConnectionString` - reader selection fails, `performSwitch` falls back to the writer, and the only trace was a `FINE` log. Reads silently never left the writer, and the application saw a perfectly healthy connection.

The first fallback on a connection is now logged at `WARNING`. Later ones stay at `FINE`, so a connection that keeps falling back does not flood the log. Same message, same call site, same code paths.

## That failure was also misreported

`ConnectionPluginManager.getHostSpecByStrategy` could not tell "no plugin supports this strategy" from "every plugin supports it but none found an eligible host", because a plugin signals the second by returning `null`. Both produced:

```
The driver does not support the requested host selection strategy: roundRobin
```

which sends the user off to rewrite a setting that was never the problem. The two outcomes are now separate messages. The no-eligible-host one names the strategy, the role that could not be satisfied, and the current host list, and stays neutral about the cause - it is reachable from any host list provider, so provider-specific advice does not belong in it. The `singleWriterConnectionString` hint lives in the read/write splitting fallback log instead, which can tell a static host list from a topology-backed one and only offers that advice when the roles actually came from the connection string. Both strings moved into the resource bundle; the old one was an inline literal. Callers still observe an `SQLException` in either case, so control flow is unchanged - only the cause type differs for the no-host case.

## Documentation

The Read/Write Splitting Plugin page previously said:

> The Read/Write Splitting Plugin is not currently supported for non-Aurora clusters.

That section is replaced with the connection-string host list setup: how to write the URL, the `singleWriterConnectionString` requirement, and what the configuration does and does not provide. Two limitations are called out explicitly, since neither is fixable without a topology: the host list is never refreshed, so an added or removed replica is not noticed, and there is no health monitoring of the listed hosts. Also noted is that per-query distribution needs the auto read/write splitting plugin, because with `setReadOnly`-driven splitting a routing decision only happens on `setReadOnly(true)`.

`singleWriterConnectionString` is now documented on that page and linked from the README property index, where its documentation cell was empty.

## Not included

Left out deliberately, and worth separate consideration:

- **Re-electing the writer on an existing connection after a promotion.** Role verification fixes new connections; making a live connection recover would mean reacting to the database's read-only error.
- **Availability bookkeeping for a static list.** With no `efm`/`failover` plugin nothing ever marks a host unavailable, so a dead replica is retried on every switch. Adding it needs a re-admit path first, since a host excluded from selection is never probed again.
- **`RoundRobinHostSelector` weight-pair key mismatch.** `roundRobinHostWeightPairs` is stored keyed by host name but read keyed by host id, so weights silently fall back to the default. Pre-existing and unrelated.

## Testing

`checkstyleMain`, `checkstyleTest`, `spotbugsMain`, `javadoc`, and the full `:aws-advanced-jdbc-wrapper:test` suite pass: 1405 tests, 0 failures. `compileJava` also passes under `-PenableCheckerFramework`.

New and updated unit tests:

- `ConnectionStringHostListProviderTest` (new): positional role assignment, `updateHostRole` replacing a role, visibility through a previously returned list, no-op for a same role or unknown host, case-insensitive host matching, and the all-writers outcome when `singleWriterConnectionString` is omitted.
- `VerifyRoleOnConnectTest`: the static-provider pass-through test is replaced by cases covering role correction into the host list, no correction when the role already matches, no host list refresh when the provider reports no change, and the unverifiable-role paths for static versus topology providers.
- `ConnectionPluginManagerTests`: the unsupported-strategy and no-eligible-host outcomes are now distinguished.

Not verified: no integration test covers this end to end. `integration.DatabaseEngineDeployment` has `RDS`, `RDS_MULTI_AZ_CLUSTER`, `RDS_MULTI_AZ_INSTANCE`, `AURORA` and `DOCKER`, none of which provisions an RDS primary with read replicas, and `AuroraTestUtility` can promote a read replica but never creates one. Exercising this against a real primary/replica pair, and the promotion scenario in particular, needs that provisioning support added first.


sophia-bq (approved):

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/handler/VerifyRoleOnConnect.java
```diff
@@ -69,18 +71,40 @@ public Connection onConnect(
...
+    } catch (final SQLException | RuntimeException e) {
+      // A topology-backed provider supplies roles of its own, so a role that cannot be measured
+      // means the connection is unusable for routing and the connection attempt fails, as before.
+      // A static host list already carries a declared role, so keep using it rather than refusing
+      // the connection: some databases reachable through a connection-string host list cannot
+      // report their role at all.
+      if (!staticHostList) {
+        throw e;
+      }
+      LOGGER.warning(() -> Messages.get("ReadWriteSplittingPlugin.staticHostListRoleNotVerified",
```
aaron-congo: Just checking, for connection-string host lists that cannot report their role at all, will this warning be logged for every connection? Would that log too many warnings or be too noisy?

wrapper/src/main/java/software/amazon/jdbc/plugin/readwritesplitting/handler/VerifyRoleOnConnect.java
```diff
@@ -99,6 +123,37 @@ public Connection onConnect(
...
+   * so it is reported at {@code WARNING}.
+   */
+  private void correctStaticHostListRole(
+      final RwSplitContext ctx,
+      final StaticHostListProvider hostListProvider,
+      final HostSpec host,
+      final HostRole measuredRole)
+      throws SQLException {
+
+    LOGGER.warning(() -> Messages.get("ReadWriteSplittingPlugin.staticHostListRoleCorrected",
```
aaron-congo: Similar question, will this be too noisy or do we want it to be noisy when a connection string is not accurate?

aaron-congo (approved):

PR #2140, opened 2026-09-16
Database Multi-Tenant Support with RQC Plugin

Frank-Gu-81:
### Summary

Improve Remote Query Cache isolation when query visibility depends on supported PostgreSQL or MySQL/MariaDB session state.

### Description

This PR adds optional authorization-aware cache isolation for PostgreSQL, MySQL, and MariaDB.

Database multi-tenancy protection is enabled through application configuration:

```java
props.setProperty("cacheEnableDatabaseMultiTenancy", "true");
```

The setting defaults to `false` to preserve existing Remote Query Cache behavior, including cache eligibility, method subscriptions, transaction-write behavior, and cache-key format. Applications whose query visibility depends on supported PostgreSQL role/search-path state or MySQL/MariaDB account/role/database state should explicitly enable it.

When enabled, the plugin:

- Includes supported PostgreSQL and MySQL/MariaDB authorization state in cache keys.
- Refreshes authorization state after recognized session-state changes, successful JDBC `setSchema()` or `setCatalog()` calls, and transaction completion.
- Recognizes additional PostgreSQL syntax, quoted parameters, and nested block comments.
- Qualifies PostgreSQL authorization-state functions with `pg_catalog` to prevent `search_path` shadowing.
- Bypasses caching when authorization state cannot be determined or represented safely.
- Conservatively disables caching after opaque operations, executable MySQL/MariaDB comments, JDBC batches, and PostgreSQL temporary-relation creation.
- Invalidates authorization state when a state-changing non-batch operation fails, while preserving the original JDBC exception.
- Skips cache writes if the physical connection changes during a cache miss.
- Avoids authorization database queries during normal cache hits and misses.

### Failed execution handling

A statement may change session state before a later part of the operation fails. For example:

```sql
USE tenant_b;
SELECT nonexistent_column;
```

The plugin handles this failure path conservatively:

```mermaid
flowchart TD
    A[Execute non-batch statement] --> B{Execution successful?}

    B -->|Yes| C{May change authorization state?}
    C -->|Recognized change| D[Refresh authorization state]
    C -->|Opaque or untracked change| E[Mark state untracked]
    C -->|No| F[Keep current state]

    B -->|No| G{Potential authorization change?}
    G -->|Recognized change| H[Mark state unknown]
    G -->|Callable or opaque operation| I[Mark state untracked]
    G -->|No| J[Keep current state]

    H --> K[Preserve and rethrow original JDBC exception]
    I --> K
    J --> K

    H --> L[Subsequent cache access bypasses stale state]
    I --> L
```

An `unknown` state may recover after a later successful authorization-state refresh. An `untracked` state disables caching for that wrapper connection.

### Scope and limitations

This setting does not automatically detect database multi-tenancy. It does not make arbitrary custom session settings or hidden function side effects safe to cache.

Cache hits do not query the database to revalidate authorization. External changes such as `GRANT`, `REVOKE`, role-membership changes, or row-level security policy changes do not invalidate entries when the tracked cache-key state remains unchanged. Previously cached results may remain available until their TTL expires or they are evicted or explicitly removed.

### Testing

- Added focused tests for PostgreSQL syntax, nested comments, temporary relations, authorization-state query hardening, failed executions, batches, JDBC database-context setters, session-state changes, and physical connection switches.
- Verified the PostgreSQL Java and Java 24 dialect variants.
- Affected focused unit tests and Checkstyle pass.


Frank-Gu-81:
cc: @QuChen88

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -64,16 +68,29 @@ public class RemoteQueryCachePlugin extends AbstractConnectionPlugin implements
   private static final String QUERY_HINT_START_PATTERN = "/*";
   private static final String QUERY_HINT_END_PATTERN = "*/";
   private static final String CACHE_PARAM_PATTERN = "CACHE_PARAM(";
+  private static final char PREPARED_STATEMENT_QUERY_QUOTE = '\'';
+  private static final String CACHE_KEY_FORMAT_VERSION = "remote-query-cache:v2";
```
QuChen88: I don't think we need to do this at this point in time. The usage of this plugin is very minimal.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -173,9 +179,95 @@ public <T, E extends Exception> T execute(
...
+    } catch (final SQLException e) {
+      sessionStateService.markAuthorizationStateUnknown();
+      LOGGER.log(
+          Level.WARNING,
+          Messages.get("DefaultConnectionPlugin.errorRefreshingAuthorizationSessionState"),
+          e);
+    }
+  }
+
+  private static @Nullable String getExecutedSql(
```
QuChen88, line 319: I remember I had such a method in the RemoteCachePlugin. Is it a re-factor?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -360,31 +464,65 @@ public <T, E extends Exception> T execute(
...
+        final int queryHintContentStart = queryHintStart + QUERY_HINT_START_PATTERN.length();
+        final int endOfQueryHint = trimmedSql.indexOf(
+            QUERY_HINT_END_PATTERN, queryHintContentStart);
+        final int nestedQueryHint = trimmedSql.indexOf(
+            QUERY_HINT_START_PATTERN, queryHintContentStart);
+        // Only the first leading, non-nested comment may opt the query into caching.
+        if (endOfQueryHint > queryHintStart
+            && (nestedQueryHint < 0 || nestedQueryHint > endOfQueryHint)) {
+          final String queryHint =
+              trimmedSql.substring(queryHintContentStart, endOfQueryHint).trim();
```
QuChen88: Why do we need to check for nested query hint? 

Let's keep this logic simple to avoid adding significant overhead per query which affects performance

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -200,26 +245,89 @@ public Set<String> getSubscribedMethods() {
...
+    }
+  }
+
+  private void logInvalidQueryHintPlacement() {
+    if (this.invalidQueryHintPlacementWarningLogged.compareAndSet(false, true)) {
+      LOGGER.warning(Messages.get("RemoteQueryCachePlugin.invalidQueryHintPlacement"));
+    }
+  }
+
+  private static boolean containsCacheParamComment(final String sql) {
```
QuChen88: Why are we changing the cache param comment logic? That should not be in scope of this PR.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -171,6 +191,31 @@ public Set<String> getSubscribedMethods() {
...
+            return null;
+          }
+          currentAuthorizationState = sessionStateService.getAuthorizationState();
+        }
+        if (!currentAuthorizationState.isPresent()) {
+          logAuthorizationStateUnavailable(null);
+          return null;
+        }
+        authorizationState = currentAuthorizationState.get();
+      }
```
QuChen88: Can you dig into this code and see how many network round-trips this requires? 

We need to be very careful to not introduce additional network hop to the database in this code flow every time as this is on the hot path of query caching.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Will remove this.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: It's one of the other issue discovered by AI when I asked it to scan through this repo. I don't think we'd want to accept query hints like below. Would you suggest I move this to another PR or just drop it and stay with our current behavior? 

```
SELECT secret_b FROM t2 /* CACHE_PARAM(ttl=60s) */ WHERE id=1
```

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
Frank-Gu-81: It is similar to the existing `RemoteQueryCachePlugin.getQuery()` method, but not a refactor currently. 

`DefaultConnectionPlugin` now needs the executed SQL independently to track authorization-state-changing statements, including queries unrelated to caching. I kept it local to avoid introducing a dependency from the default plugin to the optional cache plugin. Maybe we can extract common SQL retrieval logic later if this remains after the authorization-state redesign?

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/MysqlFamilyTargetDriverDialect.java
```diff
@@ -0,0 +1,127 @@
...
+import java.sql.Statement;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.jdbc.states.AuthorizationSessionState;
+import software.amazon.jdbc.util.SqlMethodAnalyzer;
+import software.amazon.jdbc.util.StringUtils;
+
+abstract class MysqlFamilyTargetDriverDialect extends GenericTargetDriverDialect {
```
sergiyvamz: MysqlCompatibleTargetDriverDialect?

Frank-Gu-81: Will rename this

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
QuChen88: Query hint is outside of the scope of this PR.  I would say drop this and do a follow-up to avoid making this PR too bloated.

QuChen88:
One thing about bypassing the cache if session state can't be determined - I would probably argue that we should let the customer decide if they want session state tracking or not (i.e. do they need multi-tenancy or not). If not then they should not be blocked from query caching because of some issues fetching the multi-tenancy session state information.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Removed. Will separate this improvement for another PR.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Dropped. Will move it to a separate PR.

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
```diff
@@ -38,9 +41,11 @@
...
 import software.amazon.jdbc.PropertyDefinition;
 import software.amazon.jdbc.plugin.encryption.wrapper.PgEncryptedDataHelper;
+import software.amazon.jdbc.states.AuthorizationSessionState;
 import software.amazon.jdbc.util.Messages;
 import software.amazon.jdbc.util.PropertyUtils;
 import software.amazon.jdbc.util.ResourceLock;
+import software.amazon.jdbc.util.SqlMethodAnalyzer;
 import software.amazon.jdbc.util.StringUtils;
 
 public class PgTargetDriverDialect extends GenericTargetDriverDialect {
```
sergiyvamz, line 52: Please don't forget to update the same class in Java24 folder!

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Yeah, I was a bit concerned about this as well, which was why I left out the custom RLS case. 

`hasUntrackedAuthorizationState()` and `getAuthorizationState()` are in-memory. Only `refreshAuthorizationState()` accesses the database.

### Current behavior

| Situation | Additional DB round trips |
|---|---:|
| Normal lookup with known state | 0 |
| First cacheable query on a connection | 1 |
| PostgreSQL refresh | 1 |
| MySQL/MariaDB refresh | Usually 1 |
| Older MySQL without `CURRENT_ROLE()` | 2 because of fallback |
| After `SET ROLE`, `SET search_path`, or `USE` | 1 after the statement |
| After transaction or connection-switch invalidation | 1 when reacquired |
| Refresh continually fails | 1–2 on every cacheable query |

Claude suggested this fix to me, which I think makes sense. 

>1. Recommended practical fix: move refresh outside the hot path
      - Initialize authorization state once after establishing a physical connection.
      - Refresh after successful authorization-changing statements.
      - Refresh after a connection switch.
      - getCacheQueryKey() only reads local state.
      - If state is unavailable, bypass caching.
This maintains correctness and makes cache hits perform zero DB round trips. It still adds one round trip at connection setup and after relatively uncommon state changes.

In the proposed optimization, the round trip would be as follow, which I think is the most optimal we can get as long as we support multi-tenant usage with caching feature. 

| Event | Extra DB query |
|---|---:|
| Physical connection creation | 1 |
| Normal cache hit | 0 |
| Normal cache miss | 0 for state acquisition |
| `SET ROLE` / `USE` / `SET search_path` | 1 after success |
| Connection switch | 1 |

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -173,9 +179,95 @@ public <T, E extends Exception> T execute(
       }
     }
 
+    this.updateAuthorizationSessionState(
```
sergiyvamz: This method is called all the time including statements like ResultSet.getString() that doesn't change context. Can we improve this part?

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -173,9 +179,95 @@ public <T, E extends Exception> T execute(
...
+  }
+
+  private static @Nullable String getExecutedSql(
+      final TargetDriverDialect targetDriverDialect,
+      final Object methodInvokeOn,
+      final @Nullable Object[] jdbcMethodArgs) {
+    if (jdbcMethodArgs != null
+        && jdbcMethodArgs.length > 0
+        && jdbcMethodArgs[0] instanceof String) {
+      return (String) jdbcMethodArgs[0];
```
sergiyvamz, line 332: So you assume that the first argument of JDBC call is a SQL? What about PreparedStatement.execute() ?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -460,6 +540,34 @@ public <T, E extends Exception> T execute(
...
+
+  private boolean shouldBypassCacheForTransaction() {
+    if (pluginService.isInTransaction() || pluginService.isXaTransactionActive()) {
+      return true;
+    }
+    try {
+      // With autoCommit disabled, even the first statement belongs to a transaction although the
+      // wrapper may not have observed an executing statement yet.
+      final @Nullable Connection currentConnection = pluginService.getCurrentConnection();
+      return currentConnection == null || !currentConnection.getAutoCommit();
```
sergiyvamz: currentConnection.getAutoCommit() can make a network call to the server. I think it's better to check session state service first, and if the value is unknown, then to call  getAutoCommit().

sergiyvamz: Another point is that most of frameworks (including Hibernate and Spring) usually uses autoCommit=false that will force to skip caching. Is that expected behaviour?

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
```diff
@@ -68,12 +71,23 @@ public class RemoteQueryCachePlugin extends AbstractConnectionPlugin implements
...
+          JdbcMethod.CONNECTION_SETAUTOCOMMIT.methodName,
+          JdbcMethod.STATEMENT_EXECUTEQUERY.methodName,
           JdbcMethod.STATEMENT_EXECUTE.methodName,
+          JdbcMethod.STATEMENT_EXECUTEUPDATE.methodName,
+          JdbcMethod.STATEMENT_EXECUTEBATCH.methodName,
           JdbcMethod.PREPAREDSTATEMENT_EXECUTE.methodName,
           JdbcMethod.PREPAREDSTATEMENT_EXECUTEQUERY.methodName,
+          JdbcMethod.PREPAREDSTATEMENT_EXECUTEUPDATE.methodName,
+          JdbcMethod.PREPAREDSTATEMENT_EXECUTELARGEUPDATE.methodName,
+          JdbcMethod.PREPAREDSTATEMENT_EXECUTEBATCH.methodName,
```
sergiyvamz: I'm not sure how to track authorization state for batches. Do we actually handle them?

Frank-Gu-81:
> One thing about bypassing the cache if session state can't be determined - I would probably argue that we should let the customer decide if they want session state tracking or not (i.e. do they need multi-tenancy or not). If not then they should not be blocked from query caching because of some issues fetching the multi-tenancy session state information.

@QuChen88 Sounds good. I introduced the user config in ea7129fa41e7c762a102989f173f7d377188994e. Auth session tracking is default to be `true`, but user can opt out for better performance if they do not use multi-tenant.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Optimized in [a03d11a](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/2140/commits/a03d11a6834fdf35ea6e64e412d85b546404d3f4)

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
Frank-Gu-81: Good catch! Will synchronize this in the Java 24 path.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
Frank-Gu-81: Good point. I added an early method filter so authorization-state tracking only runs for SQL execution and transaction-boundary operations. ResultSet access and other unrelated JDBC methods now bypass this logic entirely.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
Frank-Gu-81: Yeah, missed this case : ( Added in [c38fd6a](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/2140/commits/c38fd6a191d7386c8f6ba8b58d6c7a406b76f82c)

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: Thanks for catching this. I wasn't aware of Hibernate and Spring sets `autoCommit=false` by default. I think I'll just remove this check for now.

wrapper/src/main/java/software/amazon/jdbc/plugin/cache/RemoteQueryCachePlugin.java
Frank-Gu-81: We don’t track individual batch entries. My understanding is that JDBC batches usually contain DML and return update counts, so remote querying caching does not cache them. 
>Do we actually handle them?

We conservatively mark the connection’s authorization state as untracked after executeBatch(). The following statements will be bypassed if tracking is enabled. I'll add a comment on this.

Frank-Gu-81:
@QuChen88 @sergiyamz Below is an overview of the tradeoffs made for the safety improvement.

### Database multi-tenancy tradeoffs

This PR introduces optional authorization-aware cache partitioning through `cacheEnableDatabaseMultiTenancy=true`.

The setting defaults to `false` to preserve existing Remote Query Cache behavior. Applications using supported database authorization state for tenant isolation should explicitly enable it.

| Scenario | Database multi-tenancy enabled | Default behavior (`false`) | Reason for enabled behavior |
|---|---|---|---|
| Single query outside a transaction | Cached with supported authorization state in the key | Cached without authorization state | Prevents results from being shared across tracked database contexts |
| `SET ROLE`, `USE`, or another recognized state change | Bypasses cache and refreshes state afterward | State is not tracked | Subsequent queries require the updated context |
| Recognized state change fails | Marks state unknown and preserves the original exception | State is not tracked | Part of the operation may have changed session state before failing |
| Callable statement | Bypasses cache and marks state untracked | May be cached if it returns a `ResultSet` | Procedure bodies and side effects are opaque |
| Multi-statement query | Bypasses cache | May be cached if it returns a `ResultSet` | Earlier statements may affect later statements |
| JDBC batch | Not cached; state is marked untracked before execution | Not cached; future query caching continues | JDBC does not portably expose every queued operation |
| Transactional query | Bypasses cache reads and writes | Bypasses reads, but its result may populate the cache | Results may contain uncommitted data or depend on a private snapshot |
| Authorization state temporarily unavailable | Bypasses cache until state is refreshed | State is ignored | A complete authorization-aware key cannot be constructed |
| Opaque authorization-state change | Marks state untracked and disables caching for the wrapper connection | State is not tracked | The resulting state cannot be represented reliably in the cache key |
| Physical connection changes during execution | Skips the cache write if the execution context changed | Uses existing behavior | Prevents a result from being stored under a key created for another connection context |

When enabled, cache keys include:

- PostgreSQL session and effective users and search paths.
- MySQL/MariaDB session users, authenticated accounts, active roles, and current databases.

Authorization state is read during initialization, physical connection changes, recognized state changes, and transaction completion. Normal cache lookups do not perform an authorization query.

### State recovery

An **unknown** state may recover after a later successful authorization-state refresh.

An **untracked** state disables caching for that wrapper connection. Returning the connection to a pool does not automatically clear this state.

### Enabling database multi-tenancy protection

Applications whose query visibility depends on supported database roles, row-level security identities, search paths, active roles, or current databases should enable:

`cacheEnableDatabaseMultiTenancy=true`

This is an application configuration setting. The plugin does not automatically detect whether an application uses database multi-tenancy.

This setting does not make arbitrary custom session variables, custom row-level security settings, or hidden function side effects safe to cache.

Frank-Gu-81:
Quick recap on what I've hardened:

- Added **opt-in database multi-tenancy protection** via `cacheEnableDatabaseMultiTenancy=true`, defaulting to `false` to preserve legacy cache behavior.
- Added **authorization/session state to cache identity** so cached results are isolated across relevant PostgreSQL/MySQL tenant contexts.
- Added **PostgreSQL tracking** for role, session authorization, schema/search path, `set_config`, `RESET`, and related session-state changes.
- Added **MySQL/MariaDB tracking** for database changes, role changes, account/session state, `USE`, and JDBC `setCatalog()`.
- Hardened SQL recognition for **quoted identifiers, alternate valid syntax, comments, and MySQL executable comments**.
- Added handling for **JDBC `setSchema()` / `setCatalog()`** so API-level session changes refresh tenant state too.
- Added **fail-closed behavior for untracked state changes** such as stored procedures, dynamic SQL, user variables, opaque batches, and other operations whose effects cannot be safely inferred.
- Added **pre-execution invalidation for batches**, so a partially successful batch cannot leave the previous tenant snapshot usable after an exception.
- Added **failure-path invalidation** for non-batch execution, covering cases where part of a multi-statement operation changes session state before a later statement fails.
- Added a **physical-connection/context generation guard** so a query result is not written under a cache key captured before a routing/failover connection switch.
- Hardened PostgreSQL authorization-state reads by **`pg_catalog`-qualifying built-in functions**, preventing `search_path`-based function shadowing.
- Added **temporary-object hardening for PostgreSQL**, permanently disabling caching when recognized temp-object creation makes name resolution unsafe to track.
- Added explicit distinction between **unknown/recoverable state** and **permanently untracked state**, with caching bypassed appropriately.
- Preserved the **legacy disabled-mode path**: no authorization-state queries/checks, original cache eligibility semantics, and legacy key behavior when multi-tenancy protection is off.
- Avoided putting authorization refreshes on the **normal cache-hit path**; state queries occur only on initialization, connection changes, and recognized session-state events.
- Expanded documentation covering **supported tenant-isolation mechanisms, limitations, performance cost, fail-closed cases, external GRANT/REVOKE/RLS limitations, and explicit opt-in configuration**.

wrapper/src/main/java/software/amazon/jdbc/plugin/DefaultConnectionPlugin.java
```diff
@@ -129,10 +135,28 @@ public <T, E extends Exception> T execute(
     final boolean doesSwitchAutoCommitFalseTrue = sqlMethodAnalyzer.doesSwitchAutoCommitFalseTrue(
         this.pluginService.getCurrentConnection(), methodName, jdbcMethodArgs);
 
+    final boolean isBatchExecution = methodName.endsWith(".executeBatch");
```
sergiyvamz: This code can be moved to Remote Plugin. It's not required for most of driver scenarios.

Frank-Gu-81: done in [30db1c3](https://github.com/aws/aws-advanced-jdbc-wrapper/pull/2140/commits/30db1c35c912bef0f12f391ce15f9bb03aa83cf5)

sergiyvamz (approved):

wrapper/src/main/java/software/amazon/jdbc/targetdriverdialect/PgTargetDriverDialect.java
```diff
@@ -52,6 +58,46 @@ public class PgTargetDriverDialect extends GenericTargetDriverDialect {
...
+      "(?:^|;)\\s*(?:"
+          + "SET\\s+(?:(?:SESSION|LOCAL)\\s+)?(?:"
+          + "ROLE\\b|\"ROLE\""
+          + "|SESSION\\s+AUTHORIZATION\\b"
+          + "|SESSION_AUTHORIZATION\\b|\"SESSION_AUTHORIZATION\""
+          + "|SEARCH_PATH\\b|\"SEARCH_PATH\""
+          + "|SCHEMA\\b)"
+          + "|RESET\\s+(?:"
+          + "ROLE\\b|\"ROLE\""
+          + "|SESSION\\s+AUTHORIZATION\\b"
```
QuChen88: This is a fairly complex regex which is pretty error-prone and hard to validate. Is it possible to not use this?

Frank-Gu-81: This was my original concern as well. But based on my discussion with Claude, I don't think the alternatives are better. They would require a SQL parser, additional database queries, or more conservative cache disabling. SInce this regex covers a small stable set of PostgreSQL commands with parameterized tests, I'd prefer keep it if you are okay with this.

QuChen88: OK, I am getting a bit worried about the maintainability of this SQL regex. If there are no better alternatives then it is OK to do this for now.

QuChen88 (approved):

